PR #7, opened 2019-11-05
ONL-3816: Added dropdown-search component.

mcibique:


src/components/ec-dropdown-search/ec-dropdown-search.spec.js
```diff
@@ -1,40 +1,256 @@
...
+  const MockedEcTooltipDirective = {
+    bind(el, { value }) {
+      if (value.content) {
+        el.setAttribute('mocked-tooltip-content', value.content);
+        el.setAttribute('mocked-tooltip-placement', value.placement || '');
+      }
+    },
+  };
+
+  function mountDropdownSearch(props, mountOpts) {
```
matallo: beautiful, I've followed this pattern before and definitely see it as the way to go

src/components/ec-dropdown-search/ec-dropdown-search.spec.js
```diff
@@ -1,40 +1,256 @@
...
+      const wrapper = mountDropdownSearch({ items: itemsToFilter.map(item => ({ text: item.text.replace('C', 'Č') })) });
+      expect(wrapper.findAll('.ec-dropdown-search__item').length).toBe(itemsToFilter.length);
+
+      wrapper.find('.ec-dropdown-search__search-input').setValue('č');
+      expect(getItemTexts(wrapper)).toEqual(['Item ABČ', 'Item BČD', 'Item cdf']);
+    });
+  });
+
+  describe('events', () => {
+    // .find('stub').trigger('event') does nothing, so we need to emit the event from inside of the stubbed
```
matallo: can we remove this commented code?

src/utils/diacritics.js
```diff
@@ -0,0 +1,10 @@
+export function removeDiacritics(str) {
+  if (typeof str === 'string' && str.normalize) {
```
matallo: love feature matching :+1:

src/components/ec-dropdown-search/ec-dropdown-search.spec.js
mcibique: I also like it, lots of repetition is removed and tests are more cleaner.

src/components/ec-dropdown-search/ec-dropdown-search.spec.js
mcibique: It's an explanation why I use `.vm` in the tests, because we should not use it in tests only as a last possible option. I'd like to give more context to the reader why is it here.

src/components/ec-dropdown-search/ec-dropdown-search.vue
```diff
@@ -1,435 +1,261 @@
...
+            <li
+              v-for="(item, index) of filteredItems"
+              :key="item.id || index"
+              v-ec-tooltip="{
+                placement: 'right',
+                content: item.disabled ? item.disabledReason : '',
+                ...tooltipOptions,
+                ...item.tooltip,
+              }"
+              :title="item.text"
```
jaimeperezpedra: I think we must discuss this with the design team, because I think for now we only put the tool-tip for the things that are disabled and the once that the text is bigger than the container and then we put 3 dots, but I think it needs to be done with a tool-tip not only with a tittle.

matallo:
Can we make sure you can navigate and select the options with the keyboard?

mcibique:
The current dropdown search doesn't have functionality like that, but I can add it if it's necessary.

src/components/ec-dropdown-search/ec-dropdown-search.vue
mcibique: We are not going to do the detection for ellipsis (...) in JS, so the title must be there all the time.

matallo:
Yep, for basic accessibility issues I wouldn't mind increasing the spec.

mcibique:
I have created a ticket for adding keyboard navigation: ONL-3957.

src/components/ec-dropdown-search/ec-dropdown-search.story.js
```diff
@@ -1,84 +1,99 @@
...
     },
-    itemFormat: {
-      default: object('itemFormat', itemFormat),
+    isSearchEnabled: {
+      default: boolean('isSearchEnabled', true),
     },
-    items: {
-      default: object('items', items),
+    paragraphText: {
+      default: boolean('paragraphText', 'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Lectus vestibulum mattis ullamcorper velit sed ullamcorper morbi tincidunt ornare. Consequat interdum varius sit amet mattis vulputate enim nulla. Eget mi proin sed libero enim sed faucibus turpis.'),
```
stavros-tomas: I am assuming you haven't pushed yet. If not, don't forget to swap to text here :)

mcibique:
I have rebased the branch because of conflicts and pushed latest fixes we found when integrating dropdown-search with user-info in EBO.

PR #17, opened 2019-11-11
Remove outline from search input in dropdown search and fix the width of the navigation

mcibique:
To verify dropdown search:
1. Open any dropdown search story
2. Click the dropdown search element
3. The search input should not have any outline when gaining the focus.

To verify navigation width:
1. Open container with navigation story
2. Check the width of the navigation is 280px when isCollapsable knob is false
3. Check the width of the navigation is 256px when isCollapsable knob is true
4. Verify the navigation can be collapsed.


matallo:
There's a small jump when collapsed/expanded, are we aware of this? Code looks good.

Screenshot from 2019-11-11 11-27-14

PR #18, opened 2019-11-12
ONL-3960 Alert component

jaimeperezpedra:
To test this component you need to go to the alert component in the storybook and check if the behavior is the same as in the Frontify page of that [link](https://company-160717.frontify.com/document/271391#/components/notifications/alerts).

This component needs to fix when the [story of reverse buttons](https://fxsolutions.atlassian.net/browse/ONL-3966) is done because of that, I don't gonna merge this PR before that.

### DIFF
--- a/src/components/ec-alert/ec-alert.spec.js
+++ b/src/components/ec-alert/ec-alert.spec.js
@@ -63,10 +63,18 @@ describe('EcAlert', () => {
   });
 
   it('should dismiss the alert when user clicks on the dismiss icon ', () => {
-    const wrapper = mountAlert({ dismissable: true });
+    const wrapper = mount(Vue.extend({
+      components: { EcAlert },
+      data() {
+        return { isOpen: true };
+      },
+      template: `
+        <ec-alert v-model="isOpen" type="info" title="Custom random" dismissable />
+      `,
+    }));
     expect(wrapper.find('.ec-alert__dismiss-icon').exists()).toBe(true);
     wrapper.find('.ec-alert__dismiss-icon').trigger('click');
-    expect(wrapper.isVisible()).toBe(true);
+    expect(wrapper.isVisible()).toBe(false);
   });
 
   it('should emit the event when user clicks on the button', () => {
--- a/src/components/ec-alert/ec-alert.spec.js
+++ b/src/components/ec-alert/ec-alert.spec.js
@@ -26,7 +26,6 @@ describe('EcAlert', () => {
 
   it('should display only with a title given and the type', () => {
     const wrapper = mountAlert({ title: 'Random Title', type: 'info' });
-
     expect(wrapper.element).toMatchSnapshot();
   });
 
@@ -37,13 +36,11 @@ describe('EcAlert', () => {
 
   it('should display with a button with the text given in the buttonText', () => {
     const wrapper = mountAlert({ buttonText: 'Button text' });
-
     expect(wrapper.element).toMatchSnapshot();
   });
 
   it('should use the given type for the alert type and for the button too', () => {
     const wrapper = mountAlert({ type: 'error', buttonText: 'Warning button' });
-
     expect(wrapper.element).toMatchSnapshot();
   });
 
@@ -62,20 +59,17 @@ describe('EcAlert', () => {
 
   it('should render the dismiss icon when is given the prop dismissable', () => {
     const wrapper = mountAlert({ dismissable: true });
-
     expect(wrapper.element).toMatchSnapshot();
   });
 
   it('should dismiss the alert when user clicks on the dismiss icon ', () => {
     const wrapper = mountAlert({ dismissable: true });
-
     expect(wrapper.find('.ec-alert__dismiss-icon').exists()).toBe(true);
     wrapper.find('.ec-alert__dismiss-icon').trigger('click');
   });
 
   it('should emit the event when user clicks on the button', () => {
     const wrapper = mountAlert({ buttonText: 'Click here' });
-
     wrapper.find('.ec-alert__button').trigger('click');
     expect(wrapper.emitted('action').length).toBe(1);
   });
--- a/src/components/ec-alert/ec-alert.spec.js
+++ b/src/components/ec-alert/ec-alert.spec.js
@@ -1,0 +1,114 @@
+import Vue from 'vue';
+import { mount } from '@vue/test-utils';
+import EcAlert from './ec-alert.vue';
+import { withMockedConsole } from '../../../tests/utils/console';
+
+describe('EcAlert', () => {
+  function mountAlert(props, mountOpts) {
+    return mount(EcAlert, {
+      propsData: {
+        title: 'Title example',
+        type: 'info',
+        ...props,
+      },
+      ...mountOpts,
+    });
+  }
+
+  it('should throw if no props were given', () => {
+    withMockedConsole((errorSpy) => {
+      mount(EcAlert);
+      expect(errorSpy).toHaveBeenCalledTimes(2);
+      expect(errorSpy.mock.calls[0][0]).toContain('Missing required prop: "type"');
+      expect(errorSpy.mock.calls[1][0]).toContain('Missing required prop: "title"');
+    });
+  });
+
+  it('should display only with a title given and the type', () => {
+    const wrapper = mountAlert({ title: 'Random Title', type: 'info' });
+
+    expect(wrapper.element).toMatchSnapshot();
+  });
+
+  it('should display with a subtitle given', () => {
+    const wrapper = mountAlert({ subtitle: 'Subtitle example' });
+    expect(wrapper.element).toMatchSnapshot();
+  });
+
+  it('should display with a button with the text given in the buttonText', () => {
+    const wrapper = mountAlert({ buttonText: 'Button text' });
+
+    expect(wrapper.element).toMatchSnapshot();
+  });
+
+  it('should use the given type for the alert type and for the button too', () => {
+    const wrapper = mountAlert({ type: 'error', buttonText: 'Warning button' });
+
+    expect(wrapper.element).toMatchSnapshot();
+  });
+
+  it.each(['error', 'info', 'success', 'warning'])('should use the type "%s"', (type) => {
+    const wrapper = mountAlert({ type });
+    expect(wrapper.element).toMatchSnapshot();
+  });
+
+  it('should throw an error if type is not valid', () => {
+    withMockedConsole((errorSpy) => {
+      mountAlert({ type: 'invalid-value' });
+      expect(errorSpy).toHaveBeenCalledTimes(1);
+      expect(errorSpy.mock.calls[0][0]).toContain('Invalid prop: custom validator check failed for prop "type"');
+    });
+  });
+
+  it('should render the dismiss icon when is given the prop dismissable', () => {
+    const wrapper = mountAlert({ dismissable: true });
+
+    expect(wrapper.element).toMatchSnapshot();
+  });
+
+  it('should dismiss the alert when user clicks on the dismiss icon ', () => {
+    const wrapper = mountAlert({ dismissable: true });
+
+    expect(wrapper.find('.ec-alert__dismiss-icon').exists()).toBe(true);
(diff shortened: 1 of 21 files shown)


src/components/ec-alert/ec-alert.spec.js
```diff
@@ -0,0 +1,114 @@
...
+    withMockedConsole((errorSpy) => {
+      mount(EcAlert);
+      expect(errorSpy).toHaveBeenCalledTimes(2);
+      expect(errorSpy.mock.calls[0][0]).toContain('Missing required prop: "type"');
+      expect(errorSpy.mock.calls[1][0]).toContain('Missing required prop: "title"');
+    });
+  });
+
+  it('should display only with a title given and the type', () => {
+    const wrapper = mountAlert({ title: 'Random Title', type: 'info' });
```
matallo: can we keep consistency with line breaks in specs? sometimes we use blank lines, sometimes we don't

src/components/ec-alert/ec-alert.spec.js
```diff
@@ -0,0 +1,114 @@
...
+    expect(wrapper.element).toMatchSnapshot();
+  });
+
+  it('should use the given type for the alert type and for the button too', () => {
+    const wrapper = mountAlert({ type: 'error', buttonText: 'Warning button' });
+
+    expect(wrapper.element).toMatchSnapshot();
+  });
+
+  it.each(['error', 'info', 'success', 'warning'])('should use the type "%s"', (type) => {
```
matallo: lovely use of array :heart_eyes: 

we can probably get rid of parenthesis in `type`

are we passing formatter/linter to the specs?

src/components/ec-alert/ec-alert.spec.js
```diff
@@ -0,0 +1,114 @@
...
+    const wrapper = mountAlert({ type });
+    expect(wrapper.element).toMatchSnapshot();
+  });
+
+  it('should throw an error if type is not valid', () => {
+    withMockedConsole((errorSpy) => {
```
matallo: we can probably get rid of parenthesis in `errorSpy`

src/components/ec-alert/ec-alert.story.js
```diff
@@ -0,0 +1,104 @@
...
+      showError: true,
+      showInfo1: true,
+      showSuccess1: true,
+      showWarning1: true,
+      showError1: true,
+    };
+  },
+  template: `
+  <div>
+      <ec-alert title="Info Alert" type="info" />
```
matallo: honest question, could we use a Set with all the configs, and iterate? Or do we prefer to be verbose and explicit writing every combination?

src/components/ec-alert/ec-alert.spec.js
jaimeperezpedra: image
This is without parenthesis, I think we cannot skip that.
The error is _Expected parentheses around arrow function argument having a body with curly braces.eslint(arrow-parens)_

matallo: oh, I did't catch that, my :eyes: went directly to the argument, thanks!

src/components/ec-alert/ec-alert.spec.js
jaimeperezpedra: My bad I just start to use the lint extension for VC and it confused me :S

src/components/ec-alert/ec-alert.story.js
jaimeperezpedra: I don't have strong opinion about that, but maybe we can set the config for that and don't write one by one combination for all the components.

But honestly I was following the current version for that in the button component, but as I said, I'm happy to change it! :smile:

jaimeperezpedra: What do you think guys @mcibique @stavros-tomas ?

src/components/ec-alert/ec-alert.spec.js
jaimeperezpedra: Done!

src/components/ec-alert/ec-alert.story.js
stavros-tomas: Buttons we should rename the "all story" as it become much more complicated as initially intended and we don't cover every possible combination in the "all story". 

It should be a per component case scenario. If the developer feels will be faster to iterate via a config/loop lets do like that, don't copy paste from other stories.

src/components/ec-alert/ec-alert.spec.js
```diff
@@ -0,0 +1,108 @@
...
+
+  it('should render the dismiss icon when is given the prop dismissable', () => {
+    const wrapper = mountAlert({ dismissable: true });
+    expect(wrapper.element).toMatchSnapshot();
+  });
+
+  it('should dismiss the alert when user clicks on the dismiss icon ', () => {
+    const wrapper = mountAlert({ dismissable: true });
+    expect(wrapper.find('.ec-alert__dismiss-icon').exists()).toBe(true);
+    wrapper.find('.ec-alert__dismiss-icon').trigger('click');
```
mcibique: there is no expectation after the icon is clicked.

src/components/ec-alert/ec-alert.spec.js
```diff
@@ -0,0 +1,108 @@
...
+  });
+
+  it('should dismiss or show the alert when we change the v-model', () => {
+    const wrapper = mount(Vue.extend({
+      components: { EcAlert },
+      data() {
+        return { isOpen: true };
+      },
+      template: `
+        <ec-alert v-model="isOpen" type="info" title="Custom random" :dismissable="true" />
```
mcibique: CAN BE MERGED: 
```
<ec-alert v-model="isOpen" type="info" title="Custom random" dismissable />
```
should be enough if you want to save some characters.

src/components/ec-alert/ec-alert.spec.js
```diff
@@ -0,0 +1,108 @@
...
+    }));
+
+    expect(wrapper.vm.isOpen).toBe(true);
+    expect(wrapper.isVisible()).toBe(true);
+    wrapper.find('.ec-alert__dismiss-icon').trigger('click');
+    expect(wrapper.isVisible()).toBe(false);
+    expect(wrapper.vm.isOpen).toBe(false);
+  });
+  it('should render with the slot given', () => {
+    const wrapper = mount(EcAlert, {
```
mcibique: You can use `mountAlert` here.

src/components/ec-alert/ec-alert.story.js
mcibique: We should rather generate it, imagine adding new type into this.

src/components/ec-alert/ec-alert.story.js
```diff
@@ -0,0 +1,104 @@
...
+  data() {
+    return {
+      showInfo: true,
+      showSuccess: true,
+      showWarning: true,
+      showError: true,
+      showInfo1: true,
+      showSuccess1: true,
+      showWarning1: true,
+      showError1: true,
```
mcibique: If you have to use numbers then rename `showInfo` > `showInfo1` and `showInfo1` > `showInfo2` (bad naming conventions should be consistent too :))

src/components/ec-alert/ec-alert.vue
```diff
@@ -0,0 +1,181 @@
...
+    <ec-icon
+      v-if="dismissable"
+      class="ec-alert__dismiss-icon"
+      :class="`ec-alert__dismiss-icon--${type}`"
+      name="simple-close"
+      :size="16"
+      @click="$emit('change', !open)"
+    />
+    <ec-icon
+      class="ec-mr--16 ec-mt--4 ec-mb--4"
```
mcibique: No utility classes in components. Add these margins to `<style>`. Utility classes are using `!important` and it won't be overridable anywhere then. Utility classes should be used only in overriding page specific designs.

src/components/ec-alert/ec-alert.vue
```diff
@@ -0,0 +1,181 @@
...
+  color: $white;
+  fill: $white;
+  position: relative;
+  display: flex;
+  align-items: center;
+
+  &__button {
+    vertical-align: middle;
+    align-items: center;
+    height: fit-content;
```
mcibique: This is not supported by MS Edge, IE11 and even Firefox.

src/components/ec-alert/ec-alert.vue
```diff
@@ -0,0 +1,181 @@
...
+    display: inline-flex;
+    margin: auto 24px auto auto;
```
mcibique: Why not `margin-right: 24px;`?

mcibique:
Don't forget to add the component to `main.js`

src/components/ec-alert/ec-alert.vue
jaimeperezpedra: I didn't know it thanks!

src/components/ec-alert/ec-alert.story.js
jaimeperezpedra: Done! If you guys can take a look :P

src/components/ec-alert/ec-alert.spec.js
```diff
@@ -0,0 +1,107 @@
...
+  it('should render the dismiss icon when is given the prop dismissable', () => {
+    const wrapper = mountAlert({ dismissable: true });
+    expect(wrapper.element).toMatchSnapshot();
+  });
+
+  it('should dismiss the alert when user clicks on the dismiss icon ', () => {
+    const wrapper = mountAlert({ dismissable: true });
+    expect(wrapper.find('.ec-alert__dismiss-icon').exists()).toBe(true);
+    wrapper.find('.ec-alert__dismiss-icon').trigger('click');
+    expect(wrapper.isVisible()).toBe(true);
```
mcibique: It should dismiss the alert the user clicks on dismiss icon but the expectation says

`expect(wrapper.isVisible()).toBe(true);`. I think it should be false.

jaimeperezpedra: Oh... Yes, you are right! I need to send the v-model too! :S

stavros-tomas:
Hi guys, shall we try to make the button drop below the text at mobile. Currently is a bit squashed.

src/scss/tools/_typography.scss
```diff
@@ -53,6 +53,13 @@
   color: $level-3-body-and-headings;
 }
 
+@mixin body-strong {
```
stavros-tomas: Can't we use just font-weight: 700? Instead of creating a mixin?

jaimeperezpedra: In Frontify we have this typography, I add because was missing, but I can remove and add the font-weight if you want, but I think we will use a lot of this mixin.

Here is the documentation of that: 
[link](https://company-160717.frontify.com/document/271391#/foundations/typography)

src/components/ec-alert/ec-alert.vue
```diff
@@ -0,0 +1,188 @@
...
+  position: relative;
+  display: flex;
+  align-items: center;
+
+  &__content {
+    flex-grow: 1;
+  }
+
+  &__button {
+    vertical-align: middle;
```
stavros-tomas: I think the only you need the margin. You can delete vertical-align, align-items, inline-flex, unless we add icons inside the buttons.

jaimeperezpedra: Yes, I did it, not push yet

src/components/ec-alert/ec-alert.vue
```diff
@@ -0,0 +1,188 @@
+<template>
+  <div
+    v-show="open"
```
stavros-tomas: v-show is costlier than v-if. Are you sure you want to use v-show?

src/scss/tools/_typography.scss
mcibique: We should not have random `font-*` anywhere in the code. All possible combinations should be defined by typography section for consistency.

matallo:
I wasn't totally convinced about the `src/components/ec-alert/ec-alert.story.scss` as mentioned because I missed some insight or explanation in the form of a comment. The rest look brilliant.

PR #19, opened 2019-11-12
ONL-3966 Reverse status and warning status

jaimeperezpedra:
To test this PR you need to go to the button component and test a new status here is the list of that:

- Primary-reverse with Outline.
- Secondary-reverse
- Secondary-reverse with Outline
- Success-reverse
- Success-reverse with Outline
- Error-reverse
- Error-reverse with Outline
- Warning
- Warning-reverse
- Warning-reverse with Outline

The Frontify needs to be updating to compare with that, I just ask for that change.

Here is the image with all the buttons, you can check it with the [here](https://chameleon-git-onl-3966-reverse-buttons.ebury.now.sh/?path=/story/button--all-buttons-dark)
image

### DIFF
--- a/src/scss/components/ec-btn/ec-btn.story.js
+++ b/src/scss/components/ec-btn/ec-btn.story.js
@@ -125,6 +125,7 @@ stories
         <button class="ec-btn ec-btn--md ec-btn--secondary ec-m--8">Secondary</button>
         <button class="ec-btn ec-btn--sm ec-btn--success ec-m--8">Success</button>
         <button class="ec-btn ec-btn--sm ec-btn--error ec-m--8">Error</button>
+        <button class="ec-btn ec-btn--sm ec-btn--warning ec-m--8">Warning</button>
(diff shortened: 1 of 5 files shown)


mcibique:
The dark background should be enforced by the story and not by the background addon, we need to create two stories, one for light, one for dark background.

src/scss/components/ec-btn/ec-btn.story.js
```diff
@@ -105,92 +119,127 @@ stories
...
-      iconSize: {
-        default: number('Icon size', 20),
-      },
-      buttonSize: {
-        default: select('Size', ['ec-btn--sm', 'ec-btn--md'], 'ec-btn--sm'),
-      },
+    data() {
+      return {
+        element,
+        buttons: [
```
mcibique: I don't see this data being used anywhere.

jaimeperezpedra: Solved

PR #21, opened 2019-11-13
ONL-3959 modal

stavros-tomas:
This PR is for Modal component

- Modal should have 2 states (normal and large)

- It emits 3 events (positive, negative, close)

- Help Link and Close Button are optional

- Buttons are centered on mobile, with positive action first

- Buttons are right aligned on desktop with positive action button always first from right

- Help link button is a tooltip which activates on click event


src/components/ec-modal/ec-modal.vue
```diff
@@ -0,0 +1,236 @@
...
+            class="ec-mr--8"
+            name="simple-help"
+            :size="18"
+          />
+          Need help?
+        </div>
+
+        <button
+          v-if="this.$slots.negative"
+          class="ec-modal__negative-btn ec-btn ec-btn--md ec-btn--secondary ec-btn--rounded ec-push-right"
```
mcibique: No utility classes in components.

src/components/ec-modal/ec-modal.vue
```diff
@@ -0,0 +1,236 @@
...
+      </main>
+
+      <footer class="ec-modal__footer">
+        <div
+          v-if="showHelpLink"
+          v-ec-tooltip="tooltipConfig"
+          class="ec-modal__help-link"
+        >
+          <ec-icon
+            class="ec-mr--8"
```
mcibique: No utility classes in components.

src/components/ec-modal/ec-modal.story.js
```diff
@@ -0,0 +1,113 @@
...
+      negativeMethod() {
+        this.showModal = false;
+        console.log('Submit a form or similar');
+      },
+      positiveMethod() {
+        this.showModal = false;
+        console.log('Submit a form or similar');
+      },
+      closeMethod() {
+        console.log('Register the event or similar');
```
mcibique: we can use `actions` addon storybook for showing messages like this without opening the console.

src/components/ec-modal/ec-modal.story.js
```diff
@@ -0,0 +1,113 @@
+/* eslint-disable no-alert, no-console */
```
mcibique: There is no alert in the code (luckily).

src/scss/utilities/_positioning.scss
```diff
@@ -0,0 +1,26 @@
...
+// Will push a flex item to the left
+.ec-push-l {
+  margin-right: auto;
+}
+
+// Will push a flex item to the right
+.ec-push-r {
+  margin-left: auto;
+
+  &-from-1024 {
```
mcibique: I would use `@` in the name for responsive selectors, e.g. `.ec-push-r@xl`

src/components/ec-modal/ec-modal.spec.js
```diff
@@ -0,0 +1,163 @@
...
+      propsData: {
+        showModal: true,
+      },
+      slots: {
+        positive: 'Positive Button',
+      },
+    });
+
+    wrapper.find('.ec-modal__positive-btn').trigger('click');
+    expect(wrapper.emitted().positiveAction).toBeTruthy();
```
mcibique: Can be merged: You can also do `wrapper.emitted('positiveAction')`.

src/components/ec-modal/ec-modal.spec.js
```diff
@@ -0,0 +1,163 @@
+/* eslint-disable no-use-before-define */
+import { mount } from '@vue/test-utils';
+import EcModal from './ec-modal.vue';
+
+describe('EcModal', () => {
+  it('should not render the modal', () => {
```
mcibique: I'd write `should not render the modal if <when>`. Because this sounds like EcModal should never render any modal.

src/components/ec-modal/ec-modal.spec.js
```diff
@@ -0,0 +1,163 @@
...
+        showModal: true,
+      },
+      slots: {
+        main: '<p>Before we can process your application we need you to upload your management accounts. You can do this now or leave it for later.</p>',
+      },
+    });
+    expect(wrapper.element).toMatchSnapshot();
+  });
+  it('should render close button', () => {
+    const wrapper = mount(EcModal, {
```
mcibique: To make the tests repeating less code we introduced a function for mounting components:
```
function mountModal(props, mountOpts) {
  return mount(EcModal, {
    propsData: { ...props },
    ...mountOpts,
  });
}
```
```
it('should do something', () => {
  const wrapper = mountModal({ showModal: true }); // shorter and simpler
});
```

You can add more defaults for the components here, mocks, stubs, store, router, default props, etc.

src/components/ec-modal/ec-modal.spec.js
```diff
@@ -0,0 +1,163 @@
...
+      propsData: {
+        showModal: true,
+      },
+      slots: {
+        main: '<p>Before we can process your application we need you to upload your management accounts. You can do this now or leave it for later.</p>',
+      },
+    });
+    expect(wrapper.element).toMatchSnapshot();
+  });
```
mcibique: Missing newline between tests.

src/components/ec-modal/ec-modal.spec.js
```diff
@@ -0,0 +1,163 @@
...
+  it('should render close button', () => {
+    const wrapper = mount(EcModal, {
+      propsData: {
+        showModal: true,
+      },
+    });
+    expect(wrapper.find('.ec-modal__close').exists()).toBe(true);
+    expect(wrapper.element).toMatchSnapshot();
```
mcibique: BTW, to make the snapshots smaller, you can take a snapshots of partial html, e.g. in this case:
```
expect(wrapper.find('.ec-modal__header').element).toMatchSnapshot();
```
... if you know the close button is in the header. See also: https://kentcdodds.com/blog/effective-snapshot-testing/

There are many places in this file where it can be used, e.g. testing if the buttons are present in footer.

src/components/ec-modal/ec-modal.spec.js
```diff
@@ -0,0 +1,163 @@
...
+  });
+
+  it('should not render close button if set to false', () => {
```
mcibique: `should render help link when showCloseIcon is set to false`, just like it is in `should render help link when showHelpLink is set to true`

src/components/ec-modal/ec-modal.vue
```diff
@@ -0,0 +1,236 @@
...
+  },
+  methods: {
+    negativeAction() {
+      this.$emit('negativeAction');
+    },
+    positiveAction() {
+      this.$emit('positiveAction');
+    },
+    closeModal() {
+      this.$emit('closeModal', false);
```
mcibique: I'd remove the word `Modal` from it, because we are going to attach the event handler to modal, we know where we are.

I'd also consider to remove word `Action` from positive and negative events, we are going to type a lot of these. I'd perfectly fine with the code like this:
```
<ec-modal v-model="show" @close="onClose()" @positive="accepted()" @negative="rejected()" />
```

src/components/ec-modal/ec-modal.vue
```diff
@@ -0,0 +1,236 @@
...
+          v-if="showHelpLink"
+          v-ec-tooltip="tooltipConfig"
+          class="ec-modal__help-link"
+        >
+          <ec-icon
+            class="ec-mr--8"
+            name="simple-help"
+            :size="18"
+          />
+          Need help?
```
mcibique: This text must be coming from props or slot. We do not have translations in chameleon-components, so all texts must be configurable from EBO.

src/components/ec-modal/ec-modal.vue
```diff
@@ -0,0 +1,236 @@
...
+  color: $ec-modal-color;
+
+  &__content {
+    width: calc(100% - 24px);
+    max-width: 680px;
+    max-height: calc(100% - 24px);
+    position: fixed;
+    top: 50%;
+    left: 50%;
+    overflow-y: auto;
```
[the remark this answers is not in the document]
matallo: Agree, I noticed the same. Please confirm this with UX/Design.

src/scss/settings/_breakpoints.scss
```diff
@@ -0,0 +1,8 @@
+$grid-breakpoints: (
+  'xs': 320px,
+  'sm': 576px,
+  'md': 768px,
+  'lg': 992px,
+  'xl': 1024px,
+  '2xl': 1280px
+) !default;
```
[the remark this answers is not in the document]
stavros-tomas: The breakpoints are a combination of bootstrap breakpoints and what we use now. If you feel they should not be there will remove. They were added because I wanted to have a map of our breakpoints. This should not limit us to add any extra breakpoints per component/page basis as necessary.

src/scss/utilities/_positioning.scss
matallo: I've usually approached this differently, and maybe we can move this discussion offline as it can turn philosophical even,

I namespace with the responsive utility prefix, for use with utility classes. At the end it's just adding the modifier in the prefix rather than as a suffix, but it helps if you want to loop through the viewports, in sass pseudocode:

```
@each $viewport, $breakpoint in $viewports {
  @media (--#{$viewport}) {
    // component modifiers loop
      .r-#{$viewport}-{{component}}--{{modifier}} {
        // component mixin
      }
    }
  }
}
```

in any case first is the discussion if we want viewport modifiers for components

src/scss/utilities/_positioning.scss
```diff
@@ -0,0 +1,26 @@
...
+    @include media__from-1024 {
+      margin-left: auto;
```
matallo: `margin-left: auto;` is the same as the original

src/components/ec-modal/ec-modal.spec.js
stavros-tomas: Awesome cheers :)

src/components/ec-modal/ec-modal.story.js
```diff
@@ -0,0 +1,113 @@
...
+          @negativeAction = "negativeMethod()"
+          @positiveAction = "positiveMethod()"
+          @closeModal = "closeMethod()"
+          v-model="showModal">
+          <template #header>
+            <h2>Update your management accounts</h2>
+          </template>
+
+          <template #main>
+            <div style="display:flex;flex-basis:1;">
```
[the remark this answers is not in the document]
stavros-tomas: I wouldn't adopt it now because it requires some thinking otherwise I might create some classes that tomorrow we delete. I feel it's part of the grid / grid helpers talk.

src/components/ec-modal/ec-modal.vue
```diff
@@ -0,0 +1,236 @@
...
+    large: {
+      type: Boolean,
+      default: false,
+    },
+  },
+  data() {
+    return {
+      tooltipConfig: {
+        content: "<p>If you are experiencing issues, please send an email to: <a href='mailto:operationsteam@ebury.com'>operationsteam@ebury.com</a></p>",
+        classes: ['ec-tooltip--bg-white'],
```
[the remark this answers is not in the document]
mcibique: Yep, warning, error, info, success are "common" modifiers, I'm little afraid of adding a random color modifier, used only on one place. I don't want to pollute the tooltip css.

src/scss/components/_v-tooltip.scss
```diff
@@ -20,6 +20,12 @@ $ec-tooltip-width-value: 5px;
     max-width: 260px;
 
     @include shape-border-radius;
+
+    .ec-tooltip--bg-white & {
```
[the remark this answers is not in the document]
stavros-tomas: you mean expand the tooltip story?

src/scss/utilities/_positioning.scss
stavros-tomas: Should be fine for modal as it always has the viewport as a parent so the behavior is always be the same.

matallo:
Do we need a minimum height?

Screenshot from 2019-11-15 13-47-29

Also, I think help link and icon colors are off

Screenshot from 2019-11-15 13-49-06

stavros-tomas:
@matallo is not min-height is the normal margin of the element. What we can do is check if content is passed into the slots (like a do with the buttons) and if it does not, not render at all. Then again, if we start doing this then we should start doing everywhere we have slots.

@the help link now I pass it in the story that's why is off. It's up to the person that wants the help style button to style it.

src/scss/settings/_breakpoints.scss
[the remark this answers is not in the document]
stavros-tomas: @mcibique @matallo any more thoughts regarding this?

src/scss/utilities/_positioning.scss
```diff
@@ -0,0 +1,31 @@
...
+  margin-left: auto;
+}
+
+.ec-push-r-from-1024 {
+  @include media__from-1024 {
+    margin-left: auto;
+  }
+}
+
+.ec-m--auto {
```
mcibique: Why this class is not grouped together with other `ec-m--XX` classes?

src/components/ec-modal/ec-modal.vue
```diff
@@ -0,0 +1,251 @@
...
+  },
+  methods: {
+    negativeAction() {
+      this.$emit('negative');
+    },
+    positiveAction() {
+      this.$emit('positive');
+    },
+    closeModal() {
+      if (this.isClosable === true) {
```
mcibique: You don't need to compare to `true` in JS.

src/components/ec-modal/ec-modal.vue
```diff
@@ -0,0 +1,251 @@
...
+      type: Boolean,
+      default: true,
+    },
+    large: {
+      type: Boolean,
+      default: false,
+    },
+  },
+  created() {
+    const thisM = this;
```
mcibique: I would remove this variable.  What is the point if the listener is getting an arrow function?

src/components/ec-modal/ec-modal.vue
```diff
@@ -0,0 +1,251 @@
...
+
+    document.addEventListener('keyup', (e) => {
```
mcibique: We must remove the listener when modal is being destroyed otherwise we create a memory leak.

src/components/ec-modal/ec-modal.vue
```diff
@@ -0,0 +1,251 @@
...
+      if (e.keyCode === 27) {
```
mcibique: I'd explicitly extract the code `27` into a variable, because it is not clear what key are you listening to, e.g.
```
const ESC = 27;
if (e.keyCode === ESC) {}
```

src/components/ec-modal/ec-modal.vue
```diff
@@ -0,0 +1,251 @@
...
+  model: {
+    prop: 'showModal',
+    event: 'close',
+  },
+  props: {
+    showModal: {
+      type: Boolean,
+      default: false,
+    },
+    showFooterLeftContent: {
```
mcibique: I'm not sure if this prop is necessary. When I want to display a left content, I have to pass a slot AND set this prop to true. You can just check whether `$scopedSlots.leftContent`  is truthy, just like for negative and positive buttons.

src/components/ec-modal/ec-modal.vue
```diff
@@ -0,0 +1,251 @@
...
+          <button
+            v-if="this.$scopedSlots.negative"
+            class="ec-modal__negative-btn ec-btn ec-btn--md ec-btn--secondary ec-btn--rounded ec-modal__negative-btn--right"
+            @click="negativeAction()"
+          >
+            <slot name="negative" />
+          </button>
+
+          <button
+            v-if="this.$scopedSlots.positive"
```
mcibique: You don't need `this` in the template, you can remove it everywhere.

src/components/ec-modal/ec-modal.vue
```diff
@@ -0,0 +1,251 @@
...
+          <div
+            v-if="showFooterLeftContent"
+            class="ec-modal__footer-left-content"
+          >
+            <slot name="footer-left-content" />
+          </div>
+
+          <button
+            v-if="this.$scopedSlots.negative"
+            class="ec-modal__negative-btn ec-btn ec-btn--md ec-btn--secondary ec-btn--rounded ec-modal__negative-btn--right"
```
mcibique: Pedantic: maybe group classes with the same block together? `ec-btn` classes first, `ec-modal__negative` second, or vice versa.

src/components/ec-modal/ec-modal.vue
```diff
@@ -0,0 +1,251 @@
+<template>
+  <transition name="ec-modal--fade">
```
mcibique: We declared previous fade transitions as elements, so this one should be `ec-modal__fade`

src/components/ec-modal/ec-modal.spec.js
```diff
@@ -0,0 +1,163 @@
...
+  });
+
+  it('should render positive button if slot is passed', () => {
+    const wrapper = mountModal({
+      showModal: true,
+    },
+    {
+      slots: {
+        positive: 'Update management accounts',
+
```
mcibique: Empty line

src/scss/settings/_breakpoints.scss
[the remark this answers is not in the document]
matallo: Same position, confirm with them the breakpoints (given that this is greenfield and we might not need them all), and keep those as a reference to be imported.

src/components/ec-modal/ec-modal.story.js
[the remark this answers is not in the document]
matallo: Let's apply those changes then.

src/components/ec-modal/ec-modal.vue
matallo: I like this approach, too.

src/scss/settings/_breakpoints.scss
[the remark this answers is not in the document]
stavros-tomas: I confirmed with UX team will keep the ones I mention above and if we make any amends we should inform them.

src/scss/utilities/_positioning.scss
stavros-tomas: Because even though it adjusts the margins in reality is for positioning/center the item.

src/components/ec-modal/ec-modal.vue
```diff
@@ -88,11 +92,11 @@ export default {
...
-    document.addEventListener('keyup', (e) => {
-      if (e.keyCode === 27) {
-        this.closeModal();
-      }
-    });
+    document.addEventListener('keyup', this.escapeIsPressed);
+  },
+  destroyed() {
+    document.removeEventListener('keyup', this.escapeIsPressed);
+    console.log('event removed');
```
matallo: do we want to print the `console` logs?

src/components/ec-modal/ec-modal.vue
```diff
@@ -106,6 +110,13 @@ export default {
         this.$emit('close');
       }
     },
+    escapeIsPressed(e) {
+      const ESC = 27;
```
matallo: probably can go to the Enums directory

src/components/ec-modal/ec-modal.vue
```diff
@@ -111,12 +114,29 @@ export default {
       }
     },
     escapeIsPressed(e) {
-      const ESC = 27;
-
-      if (e.keyCode === ESC) {
+      if (e.keyCode === KeyCode.ESC) {
```
matallo: yass :heart_eyes:

PR #23, opened 2019-11-15
Onl 3958 privacy policy component

jaimeperezpedra:
To test this component you need to check if it looks like this [web](https://online.ebury.com/login/?next=/) if you cannot see the privacy policy component then you must refresh the cache on your browser.

Also, you need to check if this component can be displayed on small screens like phones.


src/components/ec-privacy-policy/ec-privacy-policy.spec.js
```diff
@@ -0,0 +1,48 @@
+
+// import Vue from 'vue';
```
mcibique: Empty line and commented code.

jaimeperezpedra: Done!

PR #36, opened 2019-11-26
ONL 4037 donut component

jaimeperezpedra:
In order to test this component, you must check if it has the same behaviour as in the design.
[Invision](https://projects.invisionapp.com/share/HPQ2TPOQUM2#/screens)

image

### DIFF
--- a/src/components/ec-donut/ec-donut.spec.js
+++ b/src/components/ec-donut/ec-donut.spec.js
@@ -2,7 +2,7 @@ import { mount } from '@vue/test-utils';
 import EcDonut from './ec-donut.vue';
 import { withMockedConsole } from '../../../tests/utils/console';
 
-describe('EcAlert', () => {
+describe('EcDonut', () => {
   function mountDonut(props, mountOpts) {
     return mount(EcDonut, {
       propsData: {
@@ -23,7 +23,7 @@ describe('EcAlert', () => {
     });
   });
 
-  it('should display only with a the used amount and currency are given', () => {
+  it('should display only with a the used and amount are given', () => {
     const wrapper = mountDonut();
     expect(wrapper.element).toMatchSnapshot();
   });
@@ -43,23 +43,23 @@ describe('EcAlert', () => {
     expect(wrapper.element).toMatchSnapshot();
   });
 
+  it('should display with the full used when is more used than amount', () => {
+    const wrapper = mountDonut({ used: 120 });
+    expect(wrapper.element).toMatchSnapshot();
+  });
+
   it('should get the correct computed properties', () => {
     const wrapper = mountDonut();
     const circumference = 2 * Math.PI * wrapper.vm.radius;
     expect(wrapper.vm.percentageUsed).toBe(20);
     const offset = circumference * (1 - wrapper.vm.percentageUsed / 100);
     expect(wrapper.vm.dashArray).toBe(circumference);
-    expect(wrapper.vm.offset).toBe(offset);
+    expect(wrapper.vm.dashOffset).toBe(offset);
   });
 
   it('should render slots as expected', () => {
-    const wrapper = mount(
-      EcDonut,
-      {
-        propsData: {
-          used: 20,
-          amount: 100,
-        },
+    const wrapper = mountDonut(
+      {}, {
         scopedSlots: {
           'reminder-legend': '<p>Reminder legend</p>',
           'used-legend': '<p>Used legend</p>',
--- a/src/components/ec-donut/ec-donut.spec.js
+++ b/src/components/ec-donut/ec-donut.spec.js
@@ -1,0 +1,56 @@
+import { mount } from '@vue/test-utils';
+import EcDonut from './ec-donut.vue';
+import { withMockedConsole } from '../../../tests/utils/console';
+
+describe('EcAlert', () => {
+  function mountDonut(props, mountOpts) {
+    return mount(EcDonut, {
+      propsData: {
+        used: 20,
+        amount: 100,
+        currency: 'GBP',
+        ...props,
+      },
+      ...mountOpts,
+    });
+  }
+
+  it('should throw if no props were given', () => {
+    withMockedConsole((errorSpy) => {
+      mount(EcDonut);
+      expect(errorSpy).toHaveBeenCalledTimes(3);
+      expect(errorSpy.mock.calls[0][0]).toContain('Missing required prop: "currency"');
+      expect(errorSpy.mock.calls[1][0]).toContain('Missing required prop: "amount"');
+      expect(errorSpy.mock.calls[2][0]).toContain('Missing required prop: "used"');
+    });
+  });
+
+  it('should display only with a the used amount and currency are given', () => {
+    const wrapper = mountDonut();
+    expect(wrapper.element).toMatchSnapshot();
+  });
+
+  it('should display with the half used', () => {
+    const wrapper = mountDonut({ used: 50 });
+    expect(wrapper.element).toMatchSnapshot();
+  });
+
+  it('should display with the empty used', () => {
+    const wrapper = mountDonut({ used: 0 });
+    expect(wrapper.element).toMatchSnapshot();
+  });
+
+  it('should display with the full used', () => {
+    const wrapper = mountDonut({ used: 100 });
+    expect(wrapper.element).toMatchSnapshot();
+  });
+
+  it('should get the correct computed properties', () => {
+    const wrapper = mountDonut();
+    const circumference = 2 * Math.PI * wrapper.vm.radius;
+    expect(wrapper.vm.percentageUsed).toBe(20);
+    const offset = circumference * (1 - wrapper.vm.percentageUsed / 100);
+    expect(wrapper.vm.dashArray).toBe(circumference);
+    expect(wrapper.vm.offset).toBe(offset);
+  });
+});
--- a/src/components/ec-donut/ec-donut.spec.js
+++ b/src/components/ec-donut/ec-donut.spec.js
@@ -28,7 +28,7 @@ describe('EcDonut', () => {
     expect(wrapper.element).toMatchSnapshot();
   });
 
-  it('should display with the half a', () => {
+  it('should display empty with minus number in the used prop', () => {
     const wrapper = mountDonut({ used: -10, amount: 100 });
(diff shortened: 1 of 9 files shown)


src/components/ec-donut/ec-donut.spec.js
```diff
@@ -0,0 +1,56 @@
...
+
+  it('should display with the full used', () => {
+    const wrapper = mountDonut({ used: 100 });
+    expect(wrapper.element).toMatchSnapshot();
+  });
+
+  it('should get the correct computed properties', () => {
+    const wrapper = mountDonut();
+    const circumference = 2 * Math.PI * wrapper.vm.radius;
+    expect(wrapper.vm.percentageUsed).toBe(20);
```
jaimeperezpedra: I couldn't find a better way to do this, if anyone knows, happy to change it

src/components/ec-donut/ec-donut.vue
```diff
@@ -0,0 +1,169 @@
...
+    },
+  },
+  data() {
+    return {
+      radius: 48,
+    };
+  },
+  computed: {
+    percentageUsed() {
+      return (this.used / this.amount) * 100;
```
mcibique: We should handle the case that used is greater than amount, so it never goes above 100%.

src/components/ec-donut/ec-donut.vue
```diff
@@ -0,0 +1,169 @@
...
+    stroke: $level-7-backgrounds;
+  }
+
+  &__remaining {
+    stroke: $color-info;
+  }
+
+  &__used {
+    // This color needs to be discussed with UX team,
+    // The color is this one:rgba($color-info, 0.5) but we cannot do opacity in this scenario
```
mcibique: Can you use https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-opacity then?

src/components/ec-donut/ec-donut.vue
```diff
@@ -0,0 +1,169 @@
...
+        <slot name="reminder-legend" />
+      </div>
+    </div>
+  </div>
+</template>
+
+<script>
+import EcIcon from '../ec-icon';
+
+export default {
```
mcibique: Do not forget to export the component from `main.js`.

src/components/ec-donut/ec-donut.vue
```diff
@@ -0,0 +1,169 @@
...
+  computed: {
+    percentageUsed() {
+      return (this.used / this.amount) * 100;
+    },
+    // Calculate the circumference
+    dashArray() {
+      return 2 * Math.PI * this.radius;
+    },
+    // Calculate the how much needs to be offset of the used circle
+    offset() {
```
mcibique: We have prop `dashArray` so I would call this one `dashOffset`.

src/components/ec-donut/ec-donut.story.js
```diff
@@ -0,0 +1,41 @@
...
+  },
+  props: {
+    used: {
+      default: number('Used', 2500),
+    },
+    amount: {
+      default: number('Amount', 10000),
+    },
+    currency: {
+      default: text('Currency', 'GBP'),
```
mcibique: Because the currencyFormat filter expects valid currency codes, I'd changed this one to `select(['GBP', 'EUR', 'USD', 'CAD'], 'GBP')`

src/components/ec-donut/ec-donut.spec.js
```diff
@@ -0,0 +1,72 @@
...
+  it('should throw if no props were given', () => {
+    withMockedConsole((errorSpy) => {
+      mount(EcDonut);
+      expect(errorSpy).toHaveBeenCalledTimes(2);
+      expect(errorSpy.mock.calls[0][0]).toContain('Missing required prop: "amount"');
+      expect(errorSpy.mock.calls[1][0]).toContain('Missing required prop: "used"');
+    });
+  });
+
+  it('should display only with a the used amount and currency are given', () => {
```
mcibique: The currency prop was removed. There is also typo `only with a the used`.

src/components/ec-donut/ec-donut.spec.js
mcibique: I would remove this test, because we should not be testing component internals. Pass the props and evaluate the rendered output. Can you verify this in the generated html?

src/components/ec-donut/ec-donut.spec.js
```diff
@@ -0,0 +1,72 @@
...
+    expect(wrapper.element).toMatchSnapshot();
+  });
+
+  it('should display with the empty used', () => {
+    const wrapper = mountDonut({ used: 0 });
+    expect(wrapper.element).toMatchSnapshot();
+  });
+
+  it('should display with the full used', () => {
+    const wrapper = mountDonut({ used: 100 });
```
mcibique: We need a test when used is greater than amount.

src/components/ec-donut/ec-donut.spec.js
```diff
@@ -0,0 +1,72 @@
...
+    const wrapper = mountDonut();
+    const circumference = 2 * Math.PI * wrapper.vm.radius;
+    expect(wrapper.vm.percentageUsed).toBe(20);
+    const offset = circumference * (1 - wrapper.vm.percentageUsed / 100);
+    expect(wrapper.vm.dashArray).toBe(circumference);
+    expect(wrapper.vm.offset).toBe(offset);
+  });
+
+  it('should render slots as expected', () => {
+    const wrapper = mount(
```
mcibique: Use
```
mountDonut({
  scopedSlots: {...}
});
```

src/components/ec-donut/ec-donut.spec.js
```diff
@@ -0,0 +1,72 @@
+import { mount } from '@vue/test-utils';
+import EcDonut from './ec-donut.vue';
+import { withMockedConsole } from '../../../tests/utils/console';
+
+describe('EcAlert', () => {
```
mcibique: I'd find a better name :)

mcibique:
We said in grooming with Uve that the component will display the legend bellow the chart on small devices.

src/components/ec-donut/ec-donut.vue
jaimeperezpedra: I cannot it will change the colour, because it has a background and the other line before

src/components/ec-donut/ec-donut.spec.js
matallo: Shall we test if we pass a negative?

src/components/ec-donut/ec-donut.story.js
```diff
@@ -0,0 +1,49 @@
...
+import { number, select } from '@storybook/addon-knobs';
+import EcDonut from './ec-donut.vue';
+
+const stories = storiesOf('Donut chart', module);
+
+stories.add('basic', () => ({
+  components: { EcDonut },
+  filters: {
+    currencyFormat(value, currency) {
+      return new Intl.NumberFormat('gb-GB', { style: 'currency', currency, currencyDisplay: 'code' }).format(value);
```
matallo: nice one :+1:

src/components/ec-donut/ec-donut.story.js
```diff
@@ -0,0 +1,49 @@
...
+      return this.amount - this.used;
+    },
+  },
+  template: `
+  <div style="display: flex; height: 100vh">
+    <div style="margin: auto" class="ec-card">
+      <div style="text-align: center;" class="ec-mb--24">Credit line: <strong>{{ amount | currencyFormat(currency) }}</strong></div>
+      <ec-donut class="ec-p--8" :used="used" :amount="amount">
+        <template #reminder-legend>
+          <span><strong>Reminder: </strong>{{ reminder | currencyFormat(currency) }}</span>
```
matallo: Reminder? :joy:

src/components/ec-donut/ec-donut.vue
```diff
@@ -0,0 +1,192 @@
...
+</script>
+
+<style lang="scss">
+@import '../../scss/settings/colors/index';
+@import '../../scss/tools/';
+
+.ec-donut {
+  display: flex;
+  flex-wrap: wrap;
+  // justify-content: center;
```
matallo: can we :fire: this?

src/components/ec-donut/ec-donut.vue
```diff
@@ -0,0 +1,192 @@
...
+
+  &__legend-icon {
+    flex-shrink: 0;
+    margin-right: 8px;
+
+    &--used {
+      fill: rgba($color-info, 0.5);
+    }
+
+    &--reminder {
```
matallo: this is a typo

src/components/ec-donut/ec-donut.vue
```diff
@@ -0,0 +1,192 @@
...
+      fill: $color-info;
+    }
+  }
+
+  &__svg {
+    transform: rotate(-90deg);
+  }
+
+  &__background {
+    // This color need to be discussed with UX team
```
matallo: let's add these as `// TODO: ` for discoverability

src/components/ec-donut/ec-donut.spec.js
```diff
@@ -0,0 +1,72 @@
...
+    expect(wrapper.element).toMatchSnapshot();
+  });
+
+  it('should display with the empty used', () => {
+    const wrapper = mountDonut({ used: 0 });
+    expect(wrapper.element).toMatchSnapshot();
+  });
+
+  it('should display with the full used', () => {
+    const wrapper = mountDonut({ used: 100, amount: 100 });
```
matallo: wtf, marked as resolved still not added the negative spec?

jaimeperezpedra: Done now!

src/components/ec-donut/ec-donut.spec.js
```diff
@@ -0,0 +1,77 @@
...
+      expect(errorSpy.mock.calls[1][0]).toContain('Missing required prop: "used"');
+    });
+  });
+
+  it('should display only with a the used and amount are given', () => {
+    const wrapper = mountDonut();
+    expect(wrapper.element).toMatchSnapshot();
+  });
+
+  it('should display with the half a', () => {
```
matallo: "should display with the half a" what?

PR #38, opened 2019-11-26
ONL- 4035 grid layout component

stavros-tomas:
Initial Version of a very simple grid.
We can amend as we progress and in case we need to add extra features

* 12 columns
* 24px gutters
* Flexbox bases
* Can be used for screen resolutions 320px and above.
* Tested the current layouts on IE11


src/scss/components/ec-grid/_ec-grid.scss
```diff
@@ -0,0 +1,69 @@
+.ec-grid {
+  display: flex;
+  width: 100%;
+  justify-content: space-between;
+
+  // If you don't want the outer columns to have margins then use .ec-grid--full-width
+  &--full-width {
```
matallo: I prefer to fix this via an inner wrapper, and negative margins. Also, stylistic, use only one word, either `--full` or `--wide`

src/scss/components/ec-grid/ec-grid.story.js
```diff
@@ -0,0 +1,151 @@
+/* eslint-disable no-use-before-define */
+/* eslint-disable no-plusplus */
+import { storiesOf } from '@storybook/vue';
+import './_ec-grid.story.scss';
```
mcibique: Remove the `_` from the name. It is not a partial scss file.

src/scss/components/ec-grid/ec-grid.story.js
```diff
@@ -0,0 +1,151 @@
...
+      </div>
+    </div>
+  `,
+  }))
+  .add('TFO example', () => ({
+    template: `
+      <div class="ec-ml--24 ec-mr--24">
+        <div class="ec-grid ec-grid--full-width ec-align-center-md ec-grid--col-xs ec-grid--row-md">
+          <div class="ec-col-9">
+            <h1>Lending</h1>
```
mcibique: We stopped using the word Lending. Rename it to Trade finance.

src/scss/utilities/_spacing.scss
```diff
@@ -61,3 +61,12 @@ $sides: (top, bottom, left, right); // Don't Amend
     margin-left: auto;
   }
 }
+
+// for each breakpoint
```
mcibique: I would remove comments like this.

src/scss/utilities/_spacing.scss
```diff
@@ -61,3 +61,12 @@ $sides: (top, bottom, left, right); // Don't Amend
...
+@each $breakpoint, $size in $grid-breakpoints {
```
mcibique: Rename `$grid-breakpoints` to `$breakpoints` or `$global-breakpoints`.

src/scss/components/ec-grid/ec-grid.story.js
```diff
@@ -0,0 +1,151 @@
...
+        <div
+          v-for="i in 2"
+          class="ec-col-6 grid-story__column">
+          <p>.ec-col-6</p>
+        </div>
+      </div>
+    </div>
+    `,
+  }))
+  .add('Full Width vs Normal', () => ({
```
mcibique: We name stories in lower case. We need to fix 4 of those.

src/scss/components/ec-grid/_ec-grid.scss
```diff
@@ -0,0 +1,69 @@
...
+    flex-basis: calc(100% * ( #{$i} / 12));
+  }
+}
+
+// For each breakpoint
+@each $breakpoint, $size in $grid-breakpoints {
+  @media screen and (min-width: #{$size}) {
+    .ec-grid--row {
+      &-#{$breakpoint} {
+        flex-direction: row;
```
mcibique: we need to add `flex-wrap: wrap` for row.

src/scss/components/ec-grid/_ec-grid.scss
```diff
@@ -0,0 +1,69 @@
...
+.ec-col {
+  flex: 1;
+}
+
+// Gutters and size of columns for a 12 column grid
+@for $i from 1 through 12 {
+  .ec-col-#{$i} {
+    flex-basis: calc(100% * ( #{$i} / 12));
+  }
+}
```
mcibique: We should add `ec-col-offset-{1-12}` with margin-left too.

src/scss/components/ec-grid/ec-grid.story.js
```diff
@@ -0,0 +1,151 @@
+/* eslint-disable no-use-before-define */
+/* eslint-disable no-plusplus */
```
jaimeperezpedra: Remove the eslint-disable, it  is not needed

src/scss/components/ec-grid/_ec-grid.scss
```diff
@@ -0,0 +1,69 @@
...
+// For each breakpoint
+@each $breakpoint, $size in $grid-breakpoints {
+  @media screen and (min-width: #{$size}) {
+    .ec-grid--row {
+      &-#{$breakpoint} {
+        flex-direction: row;
+
+        @for $i from 1 through 12 {
+          .ec-col-#{$i} {
+            margin: 12px;
```
jaimeperezpedra: This 12px is used more than one time, shall we create a variable?

stavros-tomas:
Refactored grid. I am not strong opinionated about it, we can go to another solution if you want.

src/scss/components/ec-grid/ec-grid.scss
```diff
@@ -0,0 +1,92 @@
+$gutters: 24px;
```
mcibique: These two variables are in global scope so we should give them more unique name, e.g.
`$ec-grid-gutters` and `$ec-grid-number-of-columns`

src/scss/components/ec-grid/ec-grid.scss
```diff
@@ -0,0 +1,92 @@
...
+$number-of-columns: 12;
+
+@mixin common {
```
mcibique: Same naming problem with this one. We already had several mixins and placeholders named like this. It would override them if haven't rename it.

src/scss/components/ec-grid/ec-grid.scss
```diff
@@ -0,0 +1,92 @@
...
+  display: flex;
+  justify-content: flex-start;
+  width: calc(100% + #{$gutters});
+  margin-left: calc(#{$gutters} / -2);
+}
+
+.ec-grid {
+  overflow: hidden;
```
mcibique: Curious to know why we want this behaviour?

src/scss/components/ec-grid/ec-grid.scss
```diff
@@ -0,0 +1,92 @@
...
+
+@for $i from 1 through $number-of-columns {
+  .ec-col-#{$i} {
+    margin: calc(#{$gutters} / 2);
+    flex-grow: 0;
+    flex-shrink: 0;
+    flex-basis: calc((100% * ( #{$i} / #{$number-of-columns})) - #{$gutters});
+
+    // For IE11 remove extra 4 px in case we have borders
+    @media all and (-ms-high-contrast: none) {
```
mcibique: Do you have any link where this problem is shown? I'm curious to know why `-ms-high-contrast` was used as a detection?

src/scss/components/ec-grid/ec-grid.scss
```diff
@@ -0,0 +1,92 @@
...
+  flex-direction: row;
+
+  &--reverse {
+    @include common;
+
+    flex-direction: row-reverse;
+  }
+
+  &--wrap {
+    flex-wrap: wrap;
```
mcibique: I would make this default and rather created `&--no-wrap` version.

src/scss/components/ec-grid/ec-grid.scss
```diff
@@ -0,0 +1,92 @@
...
+      flex-basis: calc((100% * ( #{$i} / #{$number-of-columns})) - #{$gutters} - 4px);
+    }
+  }
+
+  .ec-offset-#{$i} {
+    margin-left: calc(100% * ( #{$i} / $number-of-columns) + #{$gutters});
+  }
+}
+
+.ec-grid__row {
```
mcibique: Why have you change them from `ec-grid--row` to `ec-grid__row`? Does this mean we have to write 3 divs instead of 2?
```
<div class="ec-grid">
  <div class="ec-grid__row">
    <div class="ec-grid__col-1">
```

src/scss/components/ec-grid/ec-grid.scss
```diff
@@ -0,0 +1,92 @@
...
+    flex-basis: calc((100% * ( #{$i} / #{$number-of-columns})) - #{$gutters});
+
+    // For IE11 remove extra 4 px in case we have borders
+    @media all and (-ms-high-contrast: none) {
+      flex-basis: calc((100% * ( #{$i} / #{$number-of-columns})) - #{$gutters} - 4px);
+    }
+  }
+
+  .ec-offset-#{$i} {
+    margin-left: calc(100% * ( #{$i} / $number-of-columns) + #{$gutters});
```
mcibique: You need to use the sass interpolation inside of `calc()`, otherwise it will produce this:

```
margin-left: calc(100% * ( 4 / $number-of-columns) + 24px);
```
^^ the variable name is not resolved.

src/scss/components/ec-grid/ec-grid.story.js
```diff
@@ -0,0 +1,175 @@
...
+            </div>
+            <div class="ec-col-3" style="align-self: center;text-align:right;">
+              <button class="ec-column ec-btn ec-btn--md ec-btn--rounded ec-btn--primary">New Request</button>
+            </div>
+          </div>
+        </div>
+
+        <div class="ec-grid">
+          <div class="ec-grid__col ec-grid__row-md">
+            <div class="ec-col-8" style="height; border: 1px solid black;"> Credit line component</div>
```
mcibique: height doesn't have any value.

src/scss/components/ec-grid/ec-grid.story.js
```diff
@@ -0,0 +1,175 @@
...
+  }))
+  .add('responsive', () => ({
+    template: `
+    <div class="grid-story">
+      <h1 class="grid-story__title">Resize to swap between column and row @xs @sm @md @lg @xl breakpoints</h1>
+      <h2 class="grid-story__title">Grandparent: ec-grid</h2>
+      <h2 class="grid-story__title"> Parent: ec-grid__col ec-grid__row-xs ec-grid__col-sm ec-grid__row-md ec-grid__col-lg ec-grid__row-xl</h2>
+
+      <div class="ec-grid grid-story__container">
+        <div class="ec-grid__col ec-grid__row-xs ec-grid__col-sm ec-grid__row-md ec-grid__col-lg ec-grid__row-xl">
```
mcibique: Second opinion wanted: Are we sure we don't want to use `@` for specifying the breakpoint? e.g.
```
<div class="ec-grid__col ec-grid__row@xs ec-grid__col@sm ec-grid__row@md ec-grid__col@lg ec-grid__row@xl">
  <div class="ec-grid__col-6@lg ec-grid__col-9@xs">
```
so it's more visible what's happening.

src/scss/components/ec-grid/ec-grid.story.scss
```diff
@@ -0,0 +1,31 @@
+.grid-story {
```
mcibique: It's not mandatory here but I would follow the conventions we do have for our production scss even here and name the class '.ec-grid-story' and maybe reuse variables, paddings and typography we normally use for our components.

src/scss/components/ec-grid/ec-grid.scss
matallo: I don't see this a necessarily bad thing.

src/scss/components/ec-grid/ec-grid.scss
```diff
@@ -0,0 +1,92 @@
...
+    flex-wrap: wrap;
+  }
+}
+
+.ec-grid__col {
+  @include common;
+
+  flex-direction: column;
+
+  &-reverse {
```
matallo: is this a modifier? i.e. double--dash

src/scss/components/ec-grid/ec-grid.scss
```diff
@@ -0,0 +1,92 @@
...
+@each $breakpoint, $size in $breakpoints {
+  @media screen and (min-width: #{$size}) {
+    .ec-grid__row {
+      &-#{$breakpoint} {
+        @include common;
+
+        flex-direction: row;
+      }
+
+      &-reverse-#{$breakpoint} {
```
matallo: is this a modifier, too?

src/scss/components/ec-grid/ec-grid.scss
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    .ec-grid__col {
+      &-#{$breakpoint} {
+        @include common;
+
+        flex-direction: column;
+      }
+
+      &-reverse-#{$breakpoint} {
```
matallo: modifier spotted :eyes:

src/scss/components/ec-grid/ec-grid.scss
mcibique: I don't like typing that much, especially I'm used to grids that are using only two layers of divs. The initial version of grid was also doing it that way and I liked it.

src/scss/components/ec-grid/ec-grid.scss
stavros-tomas: In order to have offsets and always full width I need to have negative margins which makes the ec-row moving outside the parent by "gutter px/ 2". I am hiding this with overflow. It's ain't elegant but for now it seem to work.

src/scss/components/ec-grid/ec-grid.scss
stavros-tomas: 1. -ms-high-contrast is so i can target only IE11 with css. Rest of the browsers will ignore.
2. IE11 seems to ignore border-box when flex. I remove 4px which seems to be the max px we have for tfo (each border 1px)

src/scss/components/ec-grid/ec-grid.scss
stavros-tomas: Yes now we have to use 3 layers. Couldn't make it have offset with 2.

src/scss/components/ec-grid/ec-grid.scss
mcibique: What about adding padding-left: 12px and padding-right: 12px instead? I think we want to have a gap on the sides too.

src/scss/components/ec-grid/ec-grid.scss
mcibique: I'm not sure I follow, what borders are problem? There are no borders in the grid, only in the story. In case that there is no border on the container, these 4 pixels will be missing.

src/scss/components/ec-grid/ec-grid.scss
```diff
@@ -0,0 +1,92 @@
+$gutters: 24px;
+$number-of-columns: 12;
+
+@mixin common {
+  display: flex;
+  justify-content: flex-start;
+  width: calc(100% + #{$gutters});
+  margin-left: calc(#{$gutters} / -2);
```
mcibique: All these call should be resolved in SASS, we don't need unnecessary calc() to do `calc(24px / 2)` when we could do
```
margin-left: $gutters / -2;
```

src/scss/components/ec-grid/ec-grid.scss
mcibique: This hasn't been resolved

src/scss/components/ec-grid/ec-grid.scss
mcibique: This hasn't been resolved

src/scss/utilities/_spacing.scss
matallo: does this remain @stavros-tomas ?

src/scss/components/ec-grid/ec-grid.scss
matallo: > I think we want to have a gap on the sides too.
I do agree with this.

stavros-tomas: > > I think we want to have a gap on the sides too.
> 
> I do agree with this.

That's how I had it initially and with a modifier class I was making it full width but we said to have it permanently full width.

stavros-tomas:
@mcibique shall we merge this?

PR #39, opened 2019-11-27
Added normalize.css and cherry picked rules we need.

mcibique:
Added normalize.css and cherry picked rules we need. This will fix a problem with `main` element not having paddings/margins on IE11.

I have cherry-picked only things we need. The only potential problem would be with this:
```
button,
input,
optgroup,
select,
textarea {
  font-family: inherit; /* 1 */
  font-size: 100%; /* 1 */
  line-height: 1.15; /* 1 */
  margin: 0; /* 2 */
}
```
... because we already have the button implemented. I have verified that the `ec-btn` overrides all four properties so it will not be a problem.

### DIFF
--- a/src/scss/elements/_elements.scss
+++ b/src/scss/elements/_elements.scss
@@ -50,10 +50,6 @@ h6 {
   @include h6;
 }
 
-strong {
-  font-weight: bolder;
-}
-
 a {
   @include body-link;
 }
@@ -67,4 +63,9 @@ p {
 
 ::placeholder {
   color: $level-5-placeholders;
+}
+
+textarea {
+  display: block;
+  resize: none;
 }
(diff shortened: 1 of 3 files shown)


matallo:
Fun exercise, what would prevent us from using the whole file? i.e. could we activate it and make a run with our visual regression tests (once we have them)

mcibique:
Some of these rules are only for IE10, some of them are really edge cases, like `details`, `small`, `sub`, `sup`. I have never used these so I don't think it's a good thing to carry rules for them in our CSS.

matallo:
Fair enough :+1:

mcibique:
I also disagree with some rules, e.g. this:
```
pre {
   font-family: monospace, monospace; /* 1 */
   font-size: 1em; /* 2 */
}
```
^^ this should be only `code` and `pre` should care only about proper white-space wrapping. If you want to output the JSON to the website, you should then do this:
```
<pre><code>{ ...JSON  }</code></pre>
```

PR #42, opened 2019-11-27
ONL 3987 fix table component

jaimeperezpedra:
Changes:

- Remove the table-cell component(not used)
- Fix the styles on the table and remove classes not used
- Add the title to the table
- Change the icon for the current 'ec-icon' component
- Change the custom tooltip for the current 'ec-tooltip' directive

Check it with the current [designs](https://company-160717.frontify.com/document/271391#/containers/tables)

Chameleon:
image


matallo:
Tooltips seem a bit different than the designs

Screenshot from 2019-11-27 16-42-29

jaimeperezpedra:
That's because you must pass the HTML in the content if you need some style or behaviour

src/components/ec-table/ec-table.story.js
```diff
@@ -37,12 +37,21 @@ const data = [
   ],
 ];
 
+const tooltipConfig = {
+  content: 'This is the tooltip info',
+  classes: ['ec-tooltip--bg-bright', 'ec-tooltip--shadow'],
```
mcibique: I'm not very happy to see these `ec-tooltip` classes here because the ec-tooltip can change to something else and this one will not match it anymore. I'd rather see those as props, e.g.
```
background: 'bright',
hasShadow: true,
```

I'm not sure if it's going to possible to refactor the ec-tooltip to support it because it uses third party lib.

jaimeperezpedra: I don't know how I can do this, if anyone has time to help me, just ping me

mcibique: I'm also sceptical that this is easily doable without rewriting the entire ec-tooltip. I would merge it without that and create a spike for it.

PR #43, opened 2019-12-03
change the line height to 24px

jaimeperezpedra:
I double-check it with Uve


jaimeperezpedra:
Checked, the only one thing is that we must add 4px of margin-bottom in the user info (in the user name text).
I double-checked with Uve

PR #44, opened 2019-12-03
ONL-4039 - Submenu

stavros-tomas:
This is pr is for the submenu component. 
 
*We should be able to pass as many tabs as we want
*We should be able to choose the type between tabs (bordered) and a simple nav (submenu)
*In the future we need to review it and a scrollable mobile version for the tabs header


src/scss/tools/_scrollbars.scss
```diff
@@ -14,3 +14,19 @@
     background-color: $level-4-interactive-elements;
   }
 }
+
+@mixin scrollbar-thin {
```
jaimeperezpedra: Are we using that? and this is reviewed by the UX team?

src/components/ec-submenu/ec-submenu.spec.js
```diff
@@ -0,0 +1,76 @@
...
+  {
+    headerTitle: 'Another Request',
+    route: '/submitted',
+    additionalText: '(30)',
+    slotName: 'menu-item-3-content',
+  },
+];
+
+const slots = {
+  'menu-item-1-content': '<p>Lorem ipsum dolor sit, amet consectetur adipisicing elit. Temporibus, tempora esse? Fugit minus consequatur harum culpa sed laborum nulla expedita molestias, quia cupiditate commodi repudiandae labore dolore dolor. Eum sapiente repellendus tempora tenetur accusantium, molestias hic adipisci aspernatur, sed possimus explicabo culpa quas laudantium? Quidem sunt totam maxime magni minus dignissimos eos libero nobis accusantium aspernatur qui inventore odio adipisci at dolorum consectetur velit ipsa quia assumenda porro, consequatur nisi mollitia deserunt. Iusto possimus debitis assumenda harum atque libero voluptates fugit quibusdam ipsa voluptatibus eligendi ex doloremque ducimus molestias natus adipisci, earum facere repudiandae sint nobis error maiores iste. Quam.</p>',
```
matallo: the 3 slots have the same content, can we extract the lorem ipsum to a constant? Actually, you could easily refactor this further and make this a function

src/components/ec-submenu/ec-submenu.vue
```diff
@@ -0,0 +1,197 @@
+<template>
+  <div
+    ref="submenu"
+    v-if="submenu"
+    class="ec-submenu"
+    :class="{'ec-submenu--tabs': type === 'tabs'}"
```
matallo: nice one :heart_eyes_cat:

src/components/ec-submenu/ec-submenu.vue
```diff
@@ -0,0 +1,197 @@
...
+      </ul>
+    </div>
+
+    <main class="ec-submenu__main">
+      <transition-group
+        name="ec-submenu__fade"
+        tag="div"
+      >
+        <div
+          v-for="(menuItem, index ) in submenu"
```
matallo: extra blank space?

stavros-tomas:
@matallo the small jump was intentional. It was on frontify. I removed tabs and resolved rest of comments.

src/components/ec-submenu/ec-submenu.story.js
```diff
@@ -0,0 +1,77 @@
...
+    },
+    data() {
+      return {
+        activeIndex,
+      };
+    },
+    template: `
+    <div style="margin: 20px;">
+      <ec-submenu :submenu="submenu" v-model="activeIndex" >
+          <template #menu-item-1>
```
mcibique: These header templates can be removed.

src/components/ec-submenu/ec-submenu.spec.js
```diff
@@ -0,0 +1,64 @@
...
+    propsData: { ...props },
+    stubs: ['router-link'],
+    ...mountOpts,
+  });
+}
+
+const submenu = [
+  {
+    headerTitle: 'Submitted Requests',
+    href: String,
```
mcibique: `href: String`?

src/components/ec-submenu/ec-submenu.vue
```diff
@@ -0,0 +1,171 @@
...
+    v-if="submenu"
+    class="ec-submenu"
+  >
+    <div
+      ref="header"
+      class="ec-submenu__header-container"
+    >
+      <ul
+        class="ec-submenu__header"
+        size="24"
```
mcibique: `ul` cannot have a size.

src/components/ec-submenu/ec-submenu.vue
```diff
@@ -0,0 +1,171 @@
+<template>
+  <div
+    ref="submenu"
```
mcibique: ref can be removed. it's not used.

src/components/ec-submenu/ec-submenu.vue
```diff
@@ -0,0 +1,171 @@
...
+    v-if="submenu"
+    class="ec-submenu"
+  >
+    <div
+      ref="header"
```
mcibique: ref can be removed. it's not used.

src/components/ec-submenu/ec-submenu.vue
```diff
@@ -0,0 +1,171 @@
...
+      class="ec-submenu__header-container"
+    >
+      <ul
+        class="ec-submenu__header"
+        size="24"
+      >
+        <li
+          v-for="(menuItem, index) in submenu"
+          :key="menuItem.index"
```
mcibique: `:key="index"`. It wasn't throwing a warning that the key is always the same?

src/components/ec-submenu/ec-submenu.spec.js
```diff
@@ -0,0 +1,64 @@
+import { mount } from '@vue/test-utils';
```
mcibique: v-model is not tested, changing activeIndex should be tested too. Passing null to submenu should be tested. Passing empty array to the submenu prop should be tested.

src/components/ec-submenu/ec-submenu.vue
```diff
@@ -0,0 +1,171 @@
...
+    <main class="ec-submenu__main">
+      <transition-group
+        name="ec-submenu__fade"
+        tag="div"
+      >
+        <div
+          v-for="(menuItem, index) in submenu"
+          :key="index"
+          v-show=" index === activeIndex"
+          :class="{isActive: index === activeIndex}"
```
mcibique: isActive must be BEM-ified or removed because it's not used anywhere.

src/components/ec-submenu/ec-submenu.vue
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  &__header-title {
+    text-decoration: none;
+    outline: none;
+    color: $ec-submenu-body;
+    transition: color 0.3s ease-out;
+
+    &:focus {
+      outline-width: 0;
```
mcibique: I think we should change the color when gaining focus. Tabbing into submenu has no visual indication.

src/components/ec-submenu/ec-submenu.vue
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    &:hover {
+      text-decoration: none;
+      color: $ec-submenu-hover;
+    }
+  }
+
+  &__header-item--active {
```
mcibique: I'd change it to `--is-active`.

src/components/ec-submenu/ec-submenu.vue
```diff
@@ -0,0 +1,171 @@
...
+      <transition-group
+        name="ec-submenu__fade"
+        tag="div"
+      >
+        <div
+          v-for="(menuItem, index) in submenu"
+          :key="index"
+          v-show=" index === activeIndex"
+          :class="{isActive: index === activeIndex}"
+          class="fade-item"
```
mcibique: This need to get BEM-ified or removed, it's not used.

src/components/ec-submenu/ec-submenu.vue
```diff
@@ -0,0 +1,171 @@
...
+          :key="menuItem.index"
+          class="ec-submenu__header-item"
+          :class="{'ec-submenu__header-item--active': index === activeIndex}"
+          @click="$emit('change', index)"
+        >
+          <a
+            v-if="menuItem.href"
+            :href="menuItem.href"
+            class="ec-submenu__header-title"
+            @click.prevent
```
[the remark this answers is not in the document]
stavros-tomas: If I remove this instead to change tab it opens a new page.

src/components/ec-submenu/ec-submenu.vue
```diff
@@ -0,0 +1,171 @@
...
+          >
+            {{ menuItem.headerTitle }} + {{ menuItem.additionalText }}
```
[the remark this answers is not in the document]
stavros-tomas: Will amend to one without additional text

src/components/ec-submenu/ec-submenu.vue
stavros-tomas: Changed it to the same kind of blue for focus. The user can still understand what is active tab because with have the line underneath. I also confirmed this with Uve.

src/components/ec-submenu/ec-submenu.vue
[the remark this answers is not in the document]
mcibique: In that case we should not allow to pass any href to the menuItem and make it always `#`. Maybe the logic should be: If there is a route, render router-link (no prevent is needed there), otherwise render anchor with `#` and prevent.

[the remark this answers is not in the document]
stavros-tomas: i removed totally the href from a. It still loads another page. If you remove it doesn't

src/scss/tools/_transitions.scss
```diff
@@ -15,3 +15,9 @@
     opacity: 0;
   }
 }
+
+@mixin transition-ease-out(
```
mcibique: We should also update all existing occurrences, I see places in navigation-link, modal, user-info where we are transitioning color/fill, sometimes with 0.3s, sometimes with 0.5s. It would be nice to unify that.

src/components/ec-submenu/ec-submenu.vue
```diff
@@ -0,0 +1,165 @@
+<template>
+  <div
+    v-if="submenu && submenu.length>0"
```
mcibique: Add spaces around `>`.

src/components/ec-submenu/ec-submenu.spec.js
```diff
@@ -0,0 +1,112 @@
...
+          };
+        },
+      },
+    );
+
+    expect(wrapper.findByDataTest('ec-submenu__panel-0').isVisible()).toBe(true);
+    expect(wrapper.findByDataTest('ec-submenu__panel-1').isVisible()).toBe(false);
+
+    wrapper.findByDataTest('ec-submenu__header-item-1').trigger('click');
+    expect(wrapper.findByDataTest('ec-submenu__panel-1').isVisible()).toBe(true);
```
mcibique: You should also check that:
```
expect(wrapper.findByDataTest('ec-submenu__panel-0').isVisible()).toBe(false);
```

mcibique:
I have to start writing more explanatory requests. What I meant is to create one single reusable transition mixin for color and unify all occurrences that use it. If color in one component has transition 0.3s, then all components should use 0.3s. I wouldn't make it configurable and then set it everywhere to different value because that's not consistent at all.
What we ended with is this:
```
@include transition-ease-out(color, 0.5s);
``` 
to replace this:
```
transition: color 0.5s ease-out;
```
^^ there is absolutely no difference between them, they both does the same thing, there is no typing saved, what we really wanted was:
```
@mixin color-transition($property: 'color') {
  transition: $property 0.3 ease-out;
}
```
And then apply it everywhere where the color or fill does any transition. If component is using 0.5s, then scrap it and it will be using 0.3s from now. Or ask yourself why this new component is using 0.3s when the others are using 0.5s? In the future, once we have this mixin, we will not be asking this question anymore, because we just include the mixin in the component styles and the transition will be consistent as in every other component.

matallo:
These are good points and this definitely improves what we used to have regarding transitions.

But we should be really mindful how these kind of changes creep the scope of a task, and I'm a bit concerned we're spending way too much time discussing and improving things that are absolutely additional to the original scope more so since we don't have a clear directive by UX/Design.

Let's apply the suggested changes, as with the suggestion by Milos:

```@mixin color-transition($property: 'color') {
  transition: $property 0.3 ease-out;
}
```

But, even before asking ourselves what transition to use, let's ask ourselves if we should be discussing this at all at this point @mcibique @stavros-tomas

mcibique:
We must be discussing things like this because we take consistency seriously. We have three components that does this on hover:
`transition: color 0.3s ease-out`
`transition: color 0.5s`
`transition: color 0.5s ease-out`
If you see repetition like this, extract it into variable or mixin and fix all occurrences. We are doing nothing but this.

PR #45, opened 2019-12-05
fix donut component and add index to the table component

jaimeperezpedra:
image


mcibique:
Seems good to me, just to clarify: the second comment from Uve: "can the distance between the credit line text and the donut be smaller? (similar to the one between donut and used)" is going to be done in EBO, right?

jaimeperezpedra:
Yes I did already, but I need to push the version to have all the changes

PR #47, opened 2019-12-06
ONL-4040 loading component

jaimeperezpedra:
One thing regarding the component:
I was thinking to have this component in a directive, but I wasn't sure that I could do that in a short time.
I think this will be an improvement, but for now, we can leave like that, I'm not saying that this is wrong but in my point of view, the other way is more pretty.

**E.g. directive**

`<div class="ec-card" v-ec-loading="true">`

**E.g. component**

`<ec-loading show="true">`
`     <div class="ec-card" v-ec-loading="true">`
`</ec-loading>`

### DIFF
--- a/src/components/ec-loading/ec-loading.story.js
+++ b/src/components/ec-loading/ec-loading.story.js
@@ -1,0 +1,57 @@
+import { storiesOf } from '@storybook/vue';
+import { action } from '@storybook/addon-actions';
+import EcLoading from './ec-loading.vue';
+import { DARK_THEME } from '../../../.storybook/backgrounds';
+
+const stories = storiesOf('Loading', module);
+
+stories.add('with dark background', () => ({
+  components: { EcLoading },
+  template: `
+  <div style="height: 100vh">
+    <ec-loading :show="true">
+      <div>
+        Lorem ipsum dolor sit amet consectetur adipisicing elit. Fugiat ullam architecto obcaecati, facere corrupti,
+        repellat veniam quam odit esse eum soluta sequi ea minus itaque exercitationem dignissimos rerum dicta earum iste,
+        magni necessitatibus. Quisquam beatae alias fugiat cumque omnis accusamus asperiores vel doloremque repudiandae quas,
+        molestias consequatur ducimus modi reiciendis, voluptatum, veritatis deleniti commodi numquam dolores? Repudiandae,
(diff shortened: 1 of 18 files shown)


src/components/ec-loading/ec-loading.vue
```diff
@@ -0,0 +1,76 @@
+<template>
+  <div class="ec-loading__container">
```
mcibique: We should use elements inside the block, so the __container should be inside ec-loading. Give it better name or make this one `ec-loading` and `ec-loading__content` the nested one.

src/components/ec-loading/ec-loading.vue
```diff
@@ -0,0 +1,76 @@
...
+    stroke: $level-4-tech-blue;
+    stroke-linecap: round;
+    stroke-dasharray: 280;
+    stroke-dashoffset: 140;
+    stroke-width: 10px;
+    transform-origin: 50% 50%;
+  }
+}
+
+@keyframes circle-animation {
```
mcibique: Again, don't name classes/animations after elements, this should be `ec-loading__animation` or just `ec-loading-animation`.

src/components/ec-loading/ec-loading.story.js
```diff
@@ -0,0 +1,38 @@
...
+      return {
+        show: true,
+      };
+    },
+    template: `
+  <div>
+    <button class="ec-btn ec-btn--primary ec-btn--sm ec-btn--rounded" @click="show = !show">Show me</button>    
+    <ec-loading :show="show"> 
+      <div class="ec-card" style="background: tomato; color: white"> 
+        Lorem ipsum dolor sit amet consectetur adipisicing elit. Fugiat ullam architecto obcaecati, facere corrupti, repellat veniam quam odit esse eum soluta sequi ea minus itaque exercitationem dignissimos rerum dicta earum iste, magni necessitatibus. Quisquam beatae alias fugiat cumque omnis accusamus asperiores vel doloremque repudiandae quas, molestias consequatur ducimus modi reiciendis, voluptatum, veritatis deleniti commodi numquam dolores? Repudiandae, dicta laboriosam sed voluptatibus obcaecati vel laudantium et perspiciatis. Adipisci repellendus id mollitia autem? Animi odio, fuga quasi dolorem sed adipisci ipsam, ad ut fugiat officia quaerat placeat commodi ducimus! Aut beatae sequi a reiciendis harum inventore consectetur ullam rerum, adipisci mollitia nam?
```
mcibique: Could you please add anchor to the text so we can test that the link is not clickable when loading is being displayed?

PR #49, opened 2019-12-11
Onl 3996 fix tooltip

stavros-tomas:
Small PR

* amended z-index for tooltip and modal
* added box shadow and border to the existing white/bright tooltip according to design
* Updated story


src/directives/ec-tooltip/ec-tooltip.story.js
```diff
@@ -15,37 +15,68 @@ const placement = 'top';
...
 const classes = ['ec-tooltip--level-1'];
 
+const tooltipConfig = {
+  content: '<p>This is the normal tooltip</p>',
+  placement: 'bottom',
+  show: true,
+};
+
+const whiteTooltipConfig = {
+  content: '<p>This is a popover style tooltip</p>',
```
mcibique: 'popover like'

src/scss/components/_v-tooltip.scss
```diff
@@ -26,9 +26,11 @@ $ec-tooltip-width-value: 5px;
     }
 
     .ec-tooltip--bg-bright & {
+      @include box-shadow-level-1;
```
mcibique: I wouldn't add this line, we have another modifier for it: `.ec-tooltip--shadow`.

src/directives/ec-tooltip/ec-tooltip.story.js
```diff
@@ -15,37 +15,68 @@ const placement = 'top';
...
-stories.add('basic', () => ({
-  directives: { EcTooltip },
-  props: {
-    content: {
-      default: text('content', content),
-    },
-    placement: {
-      default: select('placement', ['top', 'bottom', 'left', 'right'], placement),
+stories
+  .add('basic', () => ({
```
mcibique: The ticket says: `Add all versions in a single story for regression tests`. Please modify this story  and add tooltips for all 4 positions.

src/scss/components/_v-tooltip.scss
stavros-tomas: Not sure if is correct to have a modifier as we can have only 1 type of tooltip and 1 type of popover on ebo. The only type of tooltip we can have comes with a shadow. Shall I remove the modifier?

mcibique: I wouldn't remove that, these components should follow specs in frontify and not what EBO needs. I disagree with removing functionality which we already have built because it's not in MVP for EBO.

BTW, I see 3 types of tooltips on EBO, the dark one in navigation (no shadows), the dark one in the submitted requests (with a shadow) and the bright one in the footer of the table (with the shadow).

victoriaportillo: Hi team! As we discussed in Slack, the version on black tooltip with a shadow is an error from InVision, I've already updated that screen. We only have the two scenarios that @stavros-tomas defined on the thread :)

PR #50, opened 2019-12-11
Grid quickfix

stavros-tomas:
Small PR to fix box-shadow being hidden on ebo.

It comes with caveat but should cover our current needs


jaimeperezpedra:
Remember to do the release :)

PR #54, opened 2019-12-17
WIP - Table centering by type

stavros-tomas:
Following design some content type (icons/dates/inline actions) should be automatically justified and vertically aligned. 

This is work in progress with only the icons being vertically aligned for now.


src/components/ec-table/ec-table.spec.js
```diff
@@ -82,6 +97,38 @@ describe('EcTable', () => {
...
+            data: [
+              ['foo', 'bar', { type: 'icon', text: 'lorem' }],
+              ['widgets', 'doodads'],
+            ],
+          };
+        },
+      },
+    );
+
+    expect(wrapper.findByDataTest('ec-table__cell-2').isVisible()).toBe(true);
```
matallo: is this really needed?

PR #57, opened 2019-12-18
fix table cell

jaimeperezpedra:
fix table cell, center vertically all the columns and the column who has type icon center the text also

### DIFF
--- a/src/components/ec-table-head/ec-table-head.spec.js
+++ b/src/components/ec-table-head/ec-table-head.spec.js
@@ -41,7 +41,6 @@ describe('EcTableHead', () => {
       propsData,
     });
 
-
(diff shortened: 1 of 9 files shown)


src/components/ec-table-head/ec-table-head.spec.js
```diff
@@ -25,6 +25,27 @@ describe('EcTableHead', () => {
...
+        {
+          name: 'Column B',
+          type: 'icon',
+        },
+      ],
+    };
+    const wrapper = mount(EcTableHead, {
+      propsData,
+    });
+
```
matallo: blank line break

stavros-tomas: Man...you have superpowers

stavros-tomas: 30s after push :P

PR #58, opened 2019-12-19
ec-panel component

stavros-tomas:
* Ec panel component for chameleon. 
* 2 slots for passing content (header and main)
* 1 close event
* tested on IE11

### DIFF
--- a/src/components/ec-container/ec-container.vue
+++ b/src/components/ec-container/ec-container.vue
@@ -30,6 +30,7 @@ export default {
   flex-direction: row;
   min-height: 100vh;
   align-items: stretch;
+  position: relative;
 
   &__navigation {
     flex-basis: 280px;
--- a/src/components/ec-container/ec-container.vue
+++ b/src/components/ec-container/ec-container.vue
@@ -30,7 +30,6 @@ export default {
   flex-direction: row;
   min-height: 100vh;
   align-items: stretch;
-  position: relative;
 
   &__navigation {
     flex-basis: 280px;
--- a/src/components/ec-panel/ec-panel.spec.js
+++ b/src/components/ec-panel/ec-panel.spec.js
@@ -1,0 +1,64 @@
+import { mount } from '@vue/test-utils';
+import EcPanel from './ec-panel.vue';
+
+function mountPanel(props, mountOpts) {
+  return mount(EcPanel, {
+    propsData: { ...props },
+    ...mountOpts,
+  });
+}
+
+describe('EcPanel', () => {
+  describe(':props', () => {
+    it(':showPanel - should render the panel when is true', () => {
+      const wrapper = mountPanel({ showPanel: true });
+
+      expect(wrapper.find('.ec-panel').exists()).toBe(true);
+      expect(wrapper.element).toMatchSnapshot();
+    });
+
+    it(':showPanel - should not render the panel when is false', () => {
+      const wrapper = mountPanel({ showPanel: false });
+
+      expect(wrapper.find('.ec-panel').exists()).toBe(false);
+      expect(wrapper.element).toMatchSnapshot();
+    });
+  });
+
+  describe('@events', () => {
+    it('@close - should emit a "close" event when the simple-close icon is clicked', () => {
+      const wrapper = mountPanel({ showPanel: true });
+      wrapper.find('.ec-panel__close').trigger('click');
+
+      expect(wrapper.emitted().close).toBeTruthy();
+    });
+  });
+
+  describe('#slots', () => {
+    it('#header - should render header slot section if slot is passed', () => {
+      const wrapper = mountPanel(
+        { showPanel: true },
+        {
+          slots: {
+            header: 'Submitted Requests',
+          },
+        },
+      );
+      expect(wrapper.html()).toContain('<div>Submitted Requests</div>');
+      expect(wrapper.element).toMatchSnapshot();
+    });
+
+    it('#main - should render main slot section if slot is passed', () => {
+      const wrapper = mountPanel(
+        { showPanel: true },
+        {
+          slots: {
+            main: '<p>Main content section</p>',
+          },
+        },
+      );
+      expect(wrapper.html()).toContain('<main><p>Main content section</p></main>');
+      expect(wrapper.element).toMatchSnapshot();
+    });
+  });
+});
--- a/src/components/ec-panel/ec-panel.spec.js
+++ b/src/components/ec-panel/ec-panel.spec.js
@@ -1,4 +1,4 @@
-import { mount } from '@vue/test-utils';
+import { mount, createLocalVue } from '@vue/test-utils';
 import EcPanel from './ec-panel.vue';
 
 function mountPanel(props, mountOpts) {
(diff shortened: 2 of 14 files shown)


src/components/ec-panel/ec-panel.vue
```diff
@@ -0,0 +1,96 @@
...
+@import '../../scss/settings/colors/index';
+@import '../../scss/tools/index';
+
+$ec-panel-close-btn-fill: $level-4-interactive-elements !default;
+$ec-panel-close-btn-fill-hover: $level-4-tech-blue !default;
+
+.ec-panel {
+  display: block;
+  width: 100%;
+  max-width: 352px;
```
mcibique: I'm curious where we get this magic number from. maybe extract it into variable and give it name others can better understand.

src/components/ec-panel/ec-panel.vue
```diff
@@ -0,0 +1,96 @@
...
+  height: 100vh;
+  min-height: 100%;
+  position: absolute;
+  background: white;
```
mcibique: hard-coded color name

src/components/ec-container/ec-container.vue
```diff
@@ -30,6 +30,7 @@ export default {
   flex-direction: row;
   min-height: 100vh;
   align-items: stretch;
+  position: relative;
```
mcibique: This is a "logic" required by ec-panel, it should not be in ec-container. You are tightly coupling both components together and you are saying ec-panel must be always a child of the ec-container. I'd rather avoid that.
I would remove this logic and move it to EBO. Add CSS class to the `ec-container` in EBO and add the `position: relative` rule.

src/components/ec-panel/ec-panel.spec.js
```diff
@@ -0,0 +1,64 @@
...
+      expect(wrapper.element).toMatchSnapshot();
+    });
+  });
+
+  describe('@events', () => {
+    it('@close - should emit a "close" event when the simple-close icon is clicked', () => {
+      const wrapper = mountPanel({ showPanel: true });
+      wrapper.find('.ec-panel__close').trigger('click');
+
+      expect(wrapper.emitted().close).toBeTruthy();
```
mcibique: ```expect(wrapper.emitted('close').length).toBe(1);```
would be more specific and prevents false positives.

src/components/ec-panel/ec-panel.spec.js
```diff
@@ -0,0 +1,64 @@
...
+    it('#header - should render header slot section if slot is passed', () => {
+      const wrapper = mountPanel(
+        { showPanel: true },
+        {
+          slots: {
+            header: 'Submitted Requests',
+          },
+        },
+      );
+      expect(wrapper.html()).toContain('<div>Submitted Requests</div>');
```
mcibique: I would remove this line because that's exactly what should be checked in `.toMatchSnapshot()`.

Same for the `main` slot.

src/components/ec-panel/ec-panel.vue
```diff
@@ -0,0 +1,96 @@
...
+    v-if="showPanel"
+    class="ec-panel"
+  >
+    <header class="ec-panel__header">
+      <div class="ec-panel__header-icons">
+        <ec-icon
+          class="ec-panel__close"
+          name="simple-close"
+          :size="24"
+          @click="$emit('close')"
```
mcibique: You need to emit the new value because the event is used for v-model. Otherwise this will never work:
```
<ec-panel v-model="myProp" />
```
and you always have to pass `<ec-panel v-model="myProp" @close="myProp = !myProp"  />`. That's why is it working in your story but if you drop `@close` handler, it would stop. 

Write a test for a `v-model`.

src/components/ec-panel/ec-panel.story.js
```diff
@@ -0,0 +1,53 @@
...
+        this.showPanel = false;
+      },
+    },
+    template: `
+      <div style="width: 100vw; height: 100vh; position: fixed; top: 0; left: 0;">
+        <div v-for="i in 5">Lorem, ipsum dolor sit amet consectetur adipisicing elit. Eum nobis obcaecati optio magnam, porro inventore? Suscipit sit atque a! Ullam provident quidem recusandae, itaque error labore porro inventore? Suscipit sit atque a! Ullam provident quidem recusandae, itaque error labore</div>
+
+        <ec-panel v-model="showPanel" @close="onClose()">
+          <template #header>
+            <h2 style="margin-bottom: 24px;">Submit New Request</h2>
```
mcibique: class="ec-mb--24"

src/components/ec-panel/ec-panel.vue
```diff
@@ -0,0 +1,96 @@
+<template>
+  <aside
+    v-if="showPanel"
+    class="ec-panel"
+  >
+    <header class="ec-panel__header">
+      <div class="ec-panel__header-icons">
```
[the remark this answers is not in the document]
stavros-tomas: We will add eventually more icons there (like the "go back" icon "<") that's why the plural.

src/components/ec-panel/ec-panel.vue
stavros-tomas: Got it from UX team. There is nowhere on documentation, just the max-width of the side panel. Shall I make it a variable even though we might not reuse it somewhere else?

src/components/ec-panel/ec-panel.spec.js
```diff
@@ -0,0 +1,64 @@
...
+
+function mountPanel(props, mountOpts) {
+  return mount(EcPanel, {
+    propsData: { ...props },
+    ...mountOpts,
+  });
+}
+
+describe('EcPanel', () => {
+  describe(':props', () => {
```
matallo: uhm, this is new, shall we adopt this structure in the rest of our test suite?

src/components/ec-panel/ec-panel.story.js
```diff
@@ -0,0 +1,53 @@
+import { storiesOf } from '@storybook/vue';
```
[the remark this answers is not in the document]
matallo: Are there `a11y` rules we can add to eslint for these?

src/components/ec-panel/ec-panel.story.js
```diff
@@ -0,0 +1,53 @@
...
+import { boolean } from '@storybook/addon-knobs';
+import EcPanel from './ec-panel.vue';
+
+
```
matallo: double space :eyes:

src/components/ec-panel/ec-panel.vue
matallo: +1, however personally I prefer not to extract white and black color names and always hardcode `#fff` and `#000` white will always be white, and black will always be black

src/components/ec-panel/ec-panel.spec.js
stavros-tomas: Saw it here https://youtu.be/OIpfWTThrK8?t=133 was instant love.

src/components/ec-panel/ec-panel.vue
stavros-tomas: I saw we already added white as a var so I went for "$ec-panel-background-color: $white !default"

src/components/ec-panel/ec-panel.story.js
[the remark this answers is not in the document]
stavros-tomas: Fixed the violations. 

You can pass "addParameters" but not sure how helpful this can be https://devstore.io/js/@storybook/addon-a11y

src/components/ec-panel/ec-panel.vue
mcibique: I think these colors should be in variables, but I don't like the names `$white` and `$black`. White might not be white everywhere, imagine we have light and dark theme. How the code would look like? 
```
// light.scss
$black: #000;
$white: #fff;

// dark.scss
$black: #fff;
$white: #000;
```

We have nice names for `error`, `disabled`, and `interactive-elements`. Why do we treat white as `$white`?

PR #60, opened 2019-12-23
Onl 4119 text field

jaimeperezpedra:
For this story, we only want the error state with the message and the normal behaviour on a text-field.

In the future, we will add icons to the right, tooltips, and all that is on the designs but for now, will be with the basics, here are the [designs](https://company-160717.frontify.com/document/271391#/components/input-text)

This is the actual behaviour, the first element is focused on the image.
image


src/components/ec-text-field/ec-text-field.vue
```diff
@@ -0,0 +1,106 @@
+<template>
+  <div class="ec-text-field">
+    <div
```
mcibique: 1. There is a `<label>` element.
2. We need to add basic accessibility: when I click on label, the input must gain a focus. This can be achieved by adding `<label for="input-id">Label</label><input id="input-id">` (the ID must be unique, thus generated) or just simply `<label>Label<input></label>` which doesn't require any ID.

src/components/ec-text-field/ec-text-field.vue
```diff
@@ -0,0 +1,106 @@
...
+      v-if="label"
+      class="ec-text-field__label"
+    >{{ label }}</div>
+    <input
+      class="ec-text-field__input"
+      :class="{
+        'ec-text-field__input--error': state === 'error',
```
mcibique: DRY. `state === 'error'` is used multiple times in the template, move it to computed property, don't do logic in the template.

src/components/ec-text-field/ec-text-field.vue
```diff
@@ -0,0 +1,106 @@
...
+  props: {
+    type: {
+      type: String,
+      default: 'text',
+      validator(value) {
+        return ['text', 'date', 'number'].includes(value);
+      },
+    },
+    value: {
+      default: '',
```
mcibique: Are you sure the empty string is okay for Number and Date?

src/components/ec-text-field/ec-text-field.vue
```diff
@@ -0,0 +1,106 @@
...
+    },
+    label: {
+      default: '',
+      type: String,
+    },
+    errorMessage: {
+      default: '',
+      type: String,
+    },
+    state: {
```
mcibique: I strongly disagree with the name of this prop. A `state` of the component is its internal state, a `vm`. It's very misleading to have a property with name like that, it indicates you can pass the entire state of the component from the parent, which is a big anti-pattern.

The prop can only have two values, so it would be better to make it Boolean and call it `has-error`.
Even better it would be to get rid of it and rather check if `errorMessage` has been passed. If errorMessage is empty, input is valid, otherwise make it invalid.

src/components/ec-text-field/ec-text-field.vue
```diff
@@ -0,0 +1,106 @@
...
+    },
+  },
+};
+</script>
+
+<style lang="scss">
+@import '../../scss/settings/index';
+@import '../../scss/tools/index';
+
+$ec-primay-color: $level-4-tech-blue !default;
```
mcibique: This is very generic name. `ec-primary-color` sounds like it's a primary color for every component in chameleon. You wanted `ec-text-field-primary-color`. Same for `ec-border`.

src/components/ec-text-field/ec-text-field.vue
```diff
@@ -0,0 +1,106 @@
...
+  model: {
+    prop: 'value',
+    event: 'change',
+  },
+  props: {
+    type: {
+      type: String,
+      default: 'text',
+      validator(value) {
+        return ['text', 'date', 'number'].includes(value);
```
mcibique: Why the component is called a "text field" if you can pass date and number? Either remove this prop or (more preferably) change  the name of the component to `EcInput` or `EcInputField`.

src/components/ec-text-field/ec-text-field.vue
```diff
@@ -0,0 +1,106 @@
...
+    &:focus {
+      border: 1px solid $ec-primay-color;
+      outline: none;
+    }
+  }
+
+  &__label {
+    @include caption-text;
+
+    font-weight: bold;
```
mcibique: If you find yourself setting manually `font-*` on your component to override typography mixin, you are doing it wrong. Find the proper typography mixin or ask design guys to specify new typography in specs and create a mixin for it.

src/components/ec-text-field/ec-text-field.vue
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  &__error-label {
+    @include flags;
+
+    color: $color-error-hover;
+    font-size: 12px;
```
mcibique: Same typography problem. BTW, you are overriding font-size of 12px with 12px.

src/components/ec-text-field/ec-text-field.vue
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  &__label {
+    @include caption-text;
+
+    font-weight: bold;
+  }
+
+  &__error-label {
+    @include flags;
```
mcibique: We need to rename this into `flags-text` or somehow indicate it is a typography mixin. Because I couldn't understand what kind of flags are you including here. Some boolean flags or country flags? Nope, it's a typography.

src/components/ec-text-field/ec-text-field.vue
```diff
@@ -0,0 +1,106 @@
...
+
+    padding: 8px 12px;
+    border: 1px solid $ec-border;
+    border-radius: 5px;
+    width: inherit;
+
+    &--error {
+      border: 1px solid $color-error-hover;
+
+      &:hover {
```
mcibique: It's missing the `:focus`.

src/components/ec-text-field/ec-text-field.vue
```diff
@@ -0,0 +1,106 @@
...
+
+.ec-text-field {
+  width: 100%;
+
+  &__input {
+    @include body-text;
+
+    padding: 8px 12px;
+    border: 1px solid $ec-border;
+    border-radius: 5px;
```
mcibique: `@include shape-border-radius;`

src/components/ec-text-field/ec-text-field.story.js
```diff
@@ -0,0 +1,66 @@
+import { storiesOf } from '@storybook/vue';
+import { text, select } from '@storybook/addon-knobs';
+import EcTextField from './ec-text-field.vue';
+
+const stories = storiesOf('TextField', module);
```
mcibique: We do have 
"Dropdown Seach"
"Privacy Policy"
"Inline Actions"

Could you please match the text to the pattern? It would be nice to fix the "Donut chart" too.

src/components/ec-text-field/ec-text-field.spec.js
```diff
@@ -0,0 +1,43 @@
...
+        errorMessage: 'error msg',
+        state: '',
+        label: 'label test',
+        ...props,
+      },
+      ...mountOpts,
+    });
+  }
+
+  it('should throw if no props were given', () => {
```
mcibique: The test is testing something else. You are actually passing all props. It's validations that are failing. 

In this test, call `mount(EcTextField)` and check the error, then create another test where you  `mountTextField({ type: 'random', state: 'random' });` and call it "should validate given props".

src/components/ec-text-field/ec-text-field.spec.js
```diff
@@ -0,0 +1,43 @@
...
+    withMockedConsole((errorSpy) => {
+      mountTextField({ type: 'random', state: 'random' });
+      expect(errorSpy).toHaveBeenCalledTimes(2);
+      expect(errorSpy.mock.calls[0][0]).toContain('Invalid prop: custom validator check failed for prop "type"');
+      expect(errorSpy.mock.calls[1][0]).toContain('Invalid prop: custom validator check failed for prop "state"');
+    });
+  });
+
+  it('should display properly with the given props', () => {
```
mcibique: No props are given. The description doesn't match the test.

src/components/ec-text-field/ec-text-field.spec.js
```diff
@@ -0,0 +1,43 @@
+import { mount } from '@vue/test-utils';
```
mcibique: `v-model` is not tested

src/components/ec-text-field/ec-text-field.vue
```diff
@@ -0,0 +1,106 @@
...
+    width: inherit;
+
+    &--error {
```
mcibique: Maybe `&--has-error`?

src/components/ec-text-field/ec-text-field.vue
```diff
@@ -0,0 +1,106 @@
+<template>
+  <div class="ec-text-field">
+    <div
+      v-if="label"
+      class="ec-text-field__label"
+    >{{ label }}</div>
+    <input
```
mcibique: We need to add `v-bind="$attrs" v-on="$listeners"` here because currently we cannot do this:
`<ec-text-field type="number" min="5" max="10" @focus="myHandler()" />`. We need to write a test for it too.

src/components/ec-text-field/ec-text-field.vue
jaimeperezpedra: I did in that way thinking about the future because we will need status success, and about the errorMessage, I thought it will better to set the message and play with the status better than set and unset the message of the error

src/components/ec-text-field/ec-text-field.spec.js
matallo: "renders properly"? Personally, I try to avoid boilerplate wording, otherwise we will be repeating "should" at the beginning of every spec.

src/components/ec-text-field/ec-text-field.spec.js
```diff
@@ -0,0 +1,43 @@
...
+    const wrapper = mountTextField();
+    expect(wrapper.element).toMatchSnapshot();
+  });
+
+  it('should display properly with the given state "error"', () => {
+    const wrapper = mountTextField({ state: 'error' });
+    expect(wrapper.element).toMatchSnapshot();
+  });
+
+  it('should display properly without label', () => {
```
matallo: Probably I'd approach the tests differently, empty label by default, "renders properly with label" on top of it

src/components/ec-text-field/ec-text-field.vue
matallo: +1, also add a hash map to the top with the different statuses, and use it such `STATUSES.ERROR`

maybe in the future we could have a HOC (Higher-order component) for components with a definite set of statuses

src/components/ec-text-field/ec-text-field.vue
matallo: let's name this status then, or downsize to just error as suggested by Milos

src/components/ec-text-field/ec-text-field.vue
matallo: `ec-input-field-primary-color`?

src/components/ec-text-field/ec-text-field.vue
```diff
@@ -0,0 +1,106 @@
...
+</script>
+
+<style lang="scss">
+@import '../../scss/settings/index';
+@import '../../scss/tools/index';
+
+$ec-primay-color: $level-4-tech-blue !default;
+$ec-border: $level-5-placeholders !default;
+
+.ec-text-field {
```
matallo: If we change component name, we might want to rename this class too

src/components/ec-text-field/ec-text-field.vue
```diff
@@ -0,0 +1,106 @@
...
+  width: 100%;
+
+  &__input {
```
matallo: and if the input is a child, maybe just `ec-form-field` would be more appropriate

src/components/ec-text-field/ec-text-field.vue
```diff
@@ -0,0 +1,106 @@
...
+    }
+  }
+
+  &__label {
+    @include caption-text;
+
+    font-weight: bold;
+  }
+
+  &__error-label {
```
matallo: error label? is this a label with an error status? is this an error text? I think this needs better naming or a change in the selector

matallo:
LGTM, no console errors anymore? update descriptionn accordingly :pray:

jaimeperezpedra:
Nice call @matallo

PR #64, opened 2020-01-08
Added an option to specify a template for dropdown-search item

mcibique:
Changed the way how maxHeight for the dropdown is calculated (moved from CSS to popper logic)

### DIFF
--- a/src/components/ec-dropdown-search/ec-dropdown-search.story.js
+++ b/src/components/ec-dropdown-search/ec-dropdown-search.story.js
@@ -1,6 +1,6 @@
 import { storiesOf } from '@storybook/vue';
 import {
-  boolean, object, select, text,
+  boolean, object, select, text, number,
 } from '@storybook/addon-knobs';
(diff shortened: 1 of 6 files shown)


src/components/ec-dropdown-search/ec-dropdown-search.story.js
```diff
@@ -102,4 +110,52 @@ stories.add('in container with a dynamic width', () => ({
...
+    };
+  },
+  props: {
+    items: {
+      default: object('Items', items),
+    },
+    isSearchEnabled: {
+      default: boolean('isSearchEnabled', true),
+    },
+    paragraphText: {
```
jaimeperezpedra: Can be merged: This is needed? I mean, there are no changes from the basic, just the custom slot right?

mcibique: The text is there only to demonstrate that dropdown floats over the content and not pushing it down.

PR #66, opened 2020-01-09
EcCheckbox

stavros-tomas:
Checkbox component for chameleon components

### DIFF
--- a/src/components/ec-checkbox/ec-checkbox.spec.js
+++ b/src/components/ec-checkbox/ec-checkbox.spec.js
@@ -1,8 +1,23 @@
-import { mount } from '@vue/test-utils';
+import { mount, createLocalVue } from '@vue/test-utils';
 import EcCheckbox from './ec-checkbox.vue';
 
 function mountCheckbox(props, mountOpts) {
   return mount(EcCheckbox, {
+    propsData: { ...props },
+    ...mountOpts,
+  });
+}
+
+function mountCheckboxAsTemplate(template, props, mountOpts) {
+  const localVue = createLocalVue();
+
+  const Component = localVue.extend({
+    components: { EcCheckbox },
+    template,
+  });
+
+  return mount(Component, {
+    localVue,
     propsData: { ...props },
     ...mountOpts,
   });
@@ -74,6 +89,28 @@ describe('EcCheckbox', () => {
 
       wrapper.find('input').trigger('click');
       expect(wrapper.emitted('checked-value-change').length).toBe(1);
+    });
+  });
+
+  describe('v-model', () => {
+    it('should render the checkbox and toggle v-model value when we click on input', () => {
+      const wrapper = mountCheckboxAsTemplate(
+        '<ec-checkbox v-model="checked"></ec-checkbox>',
+        {},
+        {
+          data() {
+            return {
+              checked: true,
+            };
+          },
+        },
+      );
+
+      expect(wrapper.find('.ec-checkbox').exists()).toBe(true);
+      expect(wrapper.element).toMatchSnapshot();
+
+      wrapper.find('.ec-checkbox__input').trigger('click');
+      expect(wrapper.vm.checked).toBe(false);
     });
   });
 });
--- a/src/components/ec-checkbox/ec-checkbox.spec.js
+++ b/src/components/ec-checkbox/ec-checkbox.spec.js
@@ -1,0 +1,79 @@
+import { mount } from '@vue/test-utils';
+import EcCheckbox from './ec-checkbox.vue';
+
+function mountCheckbox(props, mountOpts) {
+  return mount(EcCheckbox, {
+    propsData: { ...props },
+    ...mountOpts,
+  });
+}
+
+describe('EcCheckbox', () => {
+  describe(':props', () => {
+    it(':labelMessage - should render the checkbox with a label', () => {
+      const wrapper = mountCheckbox({
+        labelMessage: 'Test label message prop',
+      });
+
+      expect(wrapper.find('.ec-checkbox__label').exists()).toBe(true);
+      expect(wrapper.element).toMatchSnapshot();
+    });
+
+    it(':errorMessage - should render the checkbox with an errorMessage', () => {
+      const wrapper = mountCheckbox({
+        errorMessage: 'Test error message prop',
+      });
+
+      expect(wrapper.find('.ec-checkbox__error-text').exists()).toBe(true);
+      expect(wrapper.element).toMatchSnapshot();
+    });
+
+    it(':large - should render a large checkbox', () => {
+      const wrapper = mountCheckbox({
+        large: true,
+      });
+
+      expect(wrapper.find('.ec-checkbox__label--large').exists()).toBe(true);
+      expect(wrapper.element).toMatchSnapshot();
+    });
+  });
+
+  describe('#slots', () => {
+    it('#errorMessage - should render the error message slot if passed', () => {
+      const wrapper = mountCheckbox(
+        {},
+        {
+          slots: {
+            'error-message': 'Test error message slot',
+          },
+        },
+      );
+
+      expect(wrapper.find('.ec-checkbox__error-text').exists()).toBe(true);
+      expect(wrapper.element).toMatchSnapshot();
+    });
+
+    it('#labelMessage - should render the label message slot if passed', () => {
+      const wrapper = mountCheckbox(
+        {},
+        {
+          slots: {
+            'label-message': 'Test label message slot',
+          },
+        },
+      );
+
+      expect(wrapper.find('.ec-checkbox__label').exists()).toBe(true);
+      expect(wrapper.element).toMatchSnapshot();
+    });
+  });
+
+  describe('@events', () => {
+    it('@checked-value-change - should be emitted when we click on the input', () => {
+      const wrapper = mountCheckbox();
+
+      wrapper.find('input').trigger('click');
+      expect(wrapper.emitted('checked-value-change').length).toBe(1);
+    });
+  });
+});
--- a/src/components/ec-checkbox/ec-checkbox.story.js
+++ b/src/components/ec-checkbox/ec-checkbox.story.js
@@ -1,0 +1,241 @@
+import { storiesOf } from '@storybook/vue';
+import { boolean, text } from '@storybook/addon-knobs';
+import EcCheckbox from './ec-checkbox.vue';
+
+const stories = storiesOf('Checkbox', module);
+
+stories
+  .add('basic', () => ({
+    components: { EcCheckbox },
+    props: {
+      valueFromPropsChecked1: {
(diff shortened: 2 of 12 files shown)


src/components/ec-checkbox/ec-checkbox.vue
```diff
@@ -0,0 +1,212 @@
...
+      :aria-describedby="errorId"
+      type="checkbox"
+      class="ec-checkbox__input"
+      :disabled="disabled"
+    >
+    <label
+      :for="id"
+      class="ec-checkbox__label"
+      :class="{
+        'ec-checkbox__label--has-error': isInvalid || $slots['error-message'],
```
jaimeperezpedra: You are checking two times the **$slots['error-message']**

src/components/ec-checkbox/ec-checkbox.vue
```diff
@@ -0,0 +1,212 @@
...
+    >
+      <span v-if="labelMessage || $slots['label-message']">
+        {{ labelMessage }}
+        <slot name="label-message" />
+      </span>
+    </label>
+
+    <div
+      :id="errorId"
+      v-if="isInvalid || $slots['error-message']"
```
jaimeperezpedra: Same here

src/components/ec-checkbox/ec-checkbox.spec.js
```diff
@@ -0,0 +1,116 @@
...
+        {
+          data() {
+            return {
+              checked: true,
+            };
+          },
+        },
+      );
+
+      expect(wrapper.find('.ec-checkbox').exists()).toBe(true);
```
jaimeperezpedra: I'm not sure if this snapshot is needed because you are checking the model and you cannot see the model on the snapshot, on the other hand, you must check the checked property must be true before you trigger the click on the input

src/components/ec-checkbox/ec-checkbox.vue
```diff
@@ -0,0 +1,212 @@
...
+  name: 'EcCheckbox',
+  model: {
+    prop: 'checked',
+    event: 'checked-value-change',
+  },
+  props: {
+    checked: {
+      type: Boolean,
+    },
+    labelMessage: {
```
mcibique: sync this name with `ec-input-field` and name it just `label`.

src/components/ec-checkbox/ec-checkbox.vue
mcibique: you can just do this:
```
<slot name="error-message">{{errorMessage}}</slot>
```
^^ if slot is passed, it will be used, otherwise it till fallback to errorMessage prop.

Same trick can be use for label.

src/components/ec-checkbox/ec-checkbox.vue
```diff
@@ -0,0 +1,212 @@
...
+  },
+};
+</script>
+
+<style lang="scss">
+@import '../../scss/settings/index';
+@import '../../scss/tools/index';
+
+$ec-checkbox-white: $white !default;
+$ec-checkbox-tech-blue: $level-4-tech-blue !default;
```
mcibique: These variables make no sense, give them a meaning, is it background or hover or error or what color?. Remapping tech-blue to checkbox-tech-blue is waste of time, you can use the tech-blue in the rules directly.

src/components/ec-input-field/ec-input-field.vue
```diff
@@ -109,13 +109,13 @@ $ec-input-field-border: $level-5-placeholders !default;
   }
 
   &__label {
-    @include caption-text-bold;
+    @include input-text-bold;
```
mcibique: We do have typography named "Input Label". We should use that name and not using "bold" in the name.

src/components/ec-checkbox/ec-checkbox.vue
stavros-tomas: On the "ec-checkbox" unlike "ec-input-field" we decided that should be able to pass the error and label messages from both from props and slots.

Because the error can come from 2 different places I am checking if one those places is true so I can add the class.

src/components/ec-checkbox/ec-checkbox.vue
stavros-tomas: same like above

src/components/ec-checkbox/ec-checkbox.vue
```diff
@@ -0,0 +1,212 @@
...
+  }
+
+  &__label {
+    @include input-text-bold;
+
+    position: relative;
+    padding-left: 24px;
+    cursor: pointer;
+
+    .ec-checkbox__input:disabled:hover ~ & {
```
mcibique: We should be using modifiers and not `~`. tilde relies on the structure of the HTML which is worse than BEM modifiers. If I wrap the label inside a div, the CSS will stop working.

src/components/ec-checkbox/ec-checkbox.vue
```diff
@@ -0,0 +1,212 @@
...
+      cursor: default;
+    }
+
+    &:before {
+      background: $ec-checkbox-white;
```
mcibique: we should be using ec-icon (simple-check) instead of div with border.

src/components/ec-checkbox/ec-checkbox.vue
```diff
@@ -0,0 +1,212 @@
...
+      type="checkbox"
+      class="ec-checkbox__input"
+      :disabled="disabled"
+    >
+    <label
+      :for="id"
+      class="ec-checkbox__label"
+      :class="{
+        'ec-checkbox__label--has-error': isInvalid || $slots['error-message'],
+        'ec-checkbox__label--large': large
```
matallo: I don't see the need of using a `large` prop, all checkboxes should behave the same, and actually this can be detrimental as the length may vary between languages.

src/components/ec-checkbox/ec-checkbox.vue
```diff
@@ -0,0 +1,212 @@
+<template>
```
mcibique: unchecked :focus is not working. If I tab into unchecked checkbox I don't see any visual change.

src/components/ec-checkbox/ec-checkbox.vue
```diff
@@ -0,0 +1,212 @@
...
+      content: '';
+    }
+
+    &:hover {
+      &:before {
+        border-color: $ec-checkbox-level-3-hover;
+      }
+    }
+
+    &--large {
```
mcibique: `--is-large` maybe?

src/components/ec-checkbox/ec-checkbox.story.js
```diff
@@ -0,0 +1,241 @@
...
+          </template>
+        </ec-checkbox>
+      </div>
+    `,
+  }))
+  .add('all', () => ({
+    components: { EcCheckbox },
+    data() {
+      return {
+        errorMessage: 'An error has occured',
```
mcibique: 'occurred'

src/components/ec-checkbox/ec-checkbox.story.js
```diff
@@ -0,0 +1,241 @@
...
+
+        <ec-checkbox
+          v-else="valueFromPropsHasError"
+          :disabled="valueFromPropsDisabled"
+          :large="valueFromPropsLarge">
+          <template #label-message>
+            I accept the <a href="#" @click.stop.prevent="$emit('OpenTerms')"> terms and conditions </a>
+          </template>
+          <template #error-message>
+            An error has occured
```
mcibique: 'occurred'

src/components/ec-checkbox/ec-checkbox.story.js
```diff
@@ -0,0 +1,241 @@
...
+        default: boolean('Checkbox 1'),
+      },
+      valueFromPropsChecked2: {
+        default: boolean('Checkbox 2'),
+      },
+      valueFromPropsLabelMessage: {
+        default: text('Label Message', 'I accept the terms and conditions'),
+      },
+      valueFromPropsErrorMessage: {
+        default: text('Error Message', 'An error has occured'),
```
mcibique: 'occurred'

src/components/ec-checkbox/ec-checkbox.spec.js
```diff
@@ -0,0 +1,116 @@
...
+
+  return mount(Component, {
+    localVue,
+    propsData: { ...props },
+    ...mountOpts,
+  });
+}
+
+describe('EcCheckbox', () => {
+  describe(':props', () => {
```
mcibique: We should test `:disabled` prop too.

src/components/ec-checkbox/ec-checkbox.spec.js
```diff
@@ -0,0 +1,116 @@
...
+      const wrapper = mountCheckbox({
+        large: true,
+      });
+
+      expect(wrapper.find('.ec-checkbox__label--large').exists()).toBe(true);
+      expect(wrapper.element).toMatchSnapshot();
+    });
+  });
+
+  describe('#slots', () => {
```
mcibique: Maybe we should verify race conditions when I pass error-message slot and error-message prop together, which one I should see? the slot.

src/components/ec-checkbox/ec-checkbox.spec.js
mcibique: Yep, do not pollute the snapshot files, otherwise it is going to be harder to review all changes when refactoring.

src/components/ec-checkbox/ec-checkbox.vue
matallo: :-1:

src/components/ec-checkbox/ec-checkbox.spec.js
```diff
@@ -0,0 +1,116 @@
...
+    it('#errorMessage - should render the error message slot if passed', () => {
```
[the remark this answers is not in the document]
stavros-tomas: I am testing if I pass the error message as a prop and as I slot. Why should be tested for both isn't it a scenario that should never happen? Also for the last point I don't have functionality that will render one of the 2 if both are passed.

src/components/ec-checkbox/ec-checkbox.vue
stavros-tomas: I need the div

src/components/ec-checkbox/ec-checkbox.spec.js
stavros-tomas: Any ideas how to test this?

src/components/ec-checkbox/ec-checkbox.spec.js
stavros-tomas: Is similar to what I answered Carlos above. Should we hide one of the 2 and make it "dummy-proof". IMHO I don't like this. The person that writing the "HTML" should get a visual clue that he is doing something wrong.

matallo:
Duplicated https://github.com/Ebury/chameleon/pull/68

PR #67, opened 2020-01-10
Add custom background to the disabled state

jaimeperezpedra:
Add the background for the disabled state that I need on the form of the deposit section


matallo:
Bump version

PR #74, opened 2020-01-23
ONL-4158 feature: create the button group component

jaimeperezpedra:
**Button group component**:

-You can add more than 2 options on the component.
-You can add the state disabled one by one on all the buttons.
-The button must have 'text' and 'value'.
-To change the option selected you must have 'v-model'.
-The 'v-model' can be a 'Number/String/Boolean'.

[Designs](https://projects.invisionapp.com/d/main#/console/15539217/322821573/preview).

(The frontier page to this component is missing)


src/components/ec-button-group/ec-button-group.vue
```diff
@@ -0,0 +1,62 @@
...
+import { arrayOfObjectsContainsKey } from '../../utils/validators';
+
+export default {
+  name: 'EcButtonGroup',
+  model: {
+    prop: 'value',
+    event: 'change',
+  },
+  props: {
+    value: [Number, String, Boolean],
```
mcibique: I wouldn't limit the type of the value, what if I want to pass a object as a value?

src/components/ec-button-group/ec-button-group.story.js
```diff
@@ -0,0 +1,96 @@
...
+    props: {
+      items: {
+        default: object('Items', [
+          { text: 'Yes', value: 'si', disabled: true },
+          { text: 'No', value: 'dsa' },
+        ]),
+      },
+    },
+    template: `
+      <ec-button-group v-model="value" :items="items" />
```
mcibique: Could you please add at least `ec-m--16` class or grid so it does look ugly

src/components/ec-button-group/ec-button-group.vue
```diff
@@ -0,0 +1,62 @@
+<template>
+  <div class="ec-btn-group">
+    <button
+      v-for="(button, index) in items"
+      :key="index"
```
mcibique: I'd do `:key="button.id || index"` and make the id optional.

src/components/ec-button-group/ec-button-group.vue
```diff
@@ -0,0 +1,62 @@
...
+</script>
+
+<style lang="scss">
+@import '../../scss/settings/index';
+
+$ec-btn-primary-color: $level-4-tech-blue !default;
+
+.ec-btn-group {
+  &__btn:first-child {
+    border-radius: 32px 0 0 32px;
```
mcibique: I'd extract the 32px into a variable

src/utils/validators.js
```diff
@@ -0,0 +1,12 @@
...
+  if (!Array.isArray(array)) {
+    return false;
+  }
+  let isValid = true;
+  array.forEach((obj) => {
+    if (!Object.keys(obj).includes(key)) {
+      isValid = false;
+    }
+  });
+  return isValid;
```
mcibique: `return array.every(obj => (key in obj));`

src/components/ec-button-group/ec-button-group.vue
```diff
@@ -0,0 +1,62 @@
+<template>
+  <div class="ec-btn-group">
+    <button
+      v-for="(button, index) in items"
```
mcibique: I'd name the button as `item` or I'd rename items to `buttons` then (I strongly prefer the first approach)

src/components/ec-button-group/ec-button-group.vue
```diff
@@ -0,0 +1,62 @@
...
+  model: {
+    prop: 'value',
+    event: 'change',
+  },
+  props: {
+    value: [Number, String, Boolean],
+    items: {
+      type: Array,
+      required: true,
+      validator: array => arrayOfObjectsContainsKey(array, 'value') && arrayOfObjectsContainsKey(array, 'text'),
```
mcibique: CAN BE MERGED: Ideally, it would be nice to iterate the array only once, something like this: `arr => arrayOfObjectsContainsKeys(arr, 'value', 'text');`

src/components/ec-button-group/ec-button-group.vue
```diff
@@ -0,0 +1,62 @@
...
+.ec-btn-group {
+  &__btn:first-child {
+    border-radius: 32px 0 0 32px;
+  }
+
+  &__btn:not(:last-child) {
+    border-right: 0;
+  }
+
+  &__btn:disabled:not(:first-child) {
```
mcibique: What if there are two disabled buttons next to each other? They will have a blue border between them.

src/components/ec-button-group/ec-button-group.story.js
```diff
@@ -0,0 +1,96 @@
...
+import EcButtonGroup from './ec-button-group.vue';
+
+const stories = storiesOf('Button Group', module);
+
+stories
+  .add('basic', () => ({
+    components: { EcButtonGroup },
+    data() {
+      return {
+        value: 'si',
```
mcibique: Nothing against Spanish but we should have code base in English.

src/components/ec-button-group/ec-button-group.spec.js
```diff
@@ -0,0 +1,218 @@
...
+            { text: 'Maybe' },
+          ],
+        });
+        expect(errorSpy).toHaveBeenCalled();
+        expect(errorSpy.mock.calls[0][0]).toContain('Invalid prop: custom validator check failed for prop "items"');
+      });
+    });
+  });
+
+  describe(':modal', () => {
```
mcibique: v-model

src/components/ec-button-group/ec-button-group.spec.js
```diff
@@ -0,0 +1,218 @@
...
+        items: [
+          { text: 'Yes', value: 'yes' },
+          { text: 'No', value: 'no', disabled: true },
+        ],
+      });
+
+      expect(wrapper.element).toMatchSnapshot();
+    });
+
+    it(':items - should render the button group with three items with one disabled', () => {
```
matallo: honest question: what value does this add to the previous test? We could go on an on "four items and one disabled, five items and one disabled, ...)

src/components/ec-button-group/ec-button-group.spec.js
```diff
@@ -0,0 +1,218 @@
...
+              items: [
+                { text: 'Yes', value: 'yes' },
+                { text: 'No', value: 'no' },
+              ],
+            };
+          },
+        },
+      );
+      expect(wrapper.vm.value).toBe('yes');
+      expect(wrapper.find('.ec-btn--outline').text()).toBe('No');
```
matallo: I'm not the biggest fan of targeting an item (i.e. the one with the outline) with the style class, specially a modifier! that might change at any moment. and check the value has changed depending on the interaction

actually I'd approach this differently:

```
click action => that changes the selected model => that changes the .is-active class
```

thought?

src/components/ec-button-group/ec-button-group.spec.js
```diff
@@ -0,0 +1,218 @@
...
+              value: '',
+              items: [
+                { text: 'Yes', value: 'yes' },
+                { text: 'No', value: 'no' },
+              ],
+            };
+          },
+        },
+      );
+      expect(wrapper.vm.value).toBe('');
```
matallo: similar comment as above, not totally sold on this approach

src/components/ec-button-group/ec-button-group.spec.js
```diff
@@ -0,0 +1,218 @@
...
+              value: 'yes',
+              items: [
+                { text: 'Yes', value: 'yes' },
+                { text: 'No', value: 'no', disabled: true },
+              ],
+            };
+          },
+        },
+      );
+      expect(wrapper.vm.value).toBe('yes');
```
matallo: I'd keep this test scenario, but target the disabled button and checking the value of the model doesn't change after clicking the disabled button

src/components/ec-button-group/ec-button-group.spec.js
```diff
@@ -0,0 +1,218 @@
...
+      expect(wrapper.find('.ec-btn--outline').text()).toBe('No');
+      wrapper.find('.ec-btn--outline').trigger('click');
+      expect(wrapper.find('.ec-btn--outline').text()).toBe('No');
+      expect(wrapper.vm.value).toBe('yes');
+    });
+
+    it('should render the button group and change the model when click on the button with a number values', () => {
```
matallo: I find this somewhat redundant, values should be anything really , as suggested by @mcibique

src/components/ec-button-group/ec-button-group.spec.js
```diff
@@ -0,0 +1,218 @@
...
+        },
+      );
+      expect(wrapper.vm.value).toBe(1);
+      expect(wrapper.find('.ec-btn--outline').text()).toBe('No');
+      wrapper.find('.ec-btn--outline').trigger('click');
+      expect(wrapper.find('.ec-btn--outline').text()).toBe('Yes');
+      expect(wrapper.vm.value).toBe(2);
+    });
+
+    it('should render the button group and change the model when click on the button with a boolean values', () => {
```
matallo: same comment as above, we're just repeating specs here

src/components/ec-button-group/ec-button-group.vue
```diff
@@ -0,0 +1,62 @@
...
+  },
+  props: {
+    value: [Number, String, Boolean],
+    items: {
+      type: Array,
+      required: true,
+      validator: array => arrayOfObjectsContainsKey(array, 'value') && arrayOfObjectsContainsKey(array, 'text'),
+    },
+  },
+  methods: {
```
matallo: I'm a bit ignorant about Vue api, can we get rid of the empty methods object?

jaimeperezpedra: Yes sure, my bad! Thx

src/components/ec-button-group/ec-button-group.spec.js
mcibique: We will have the same problems with targeting elements in integration tests. We should add data-test attributes, e.g. data-test="ec-button-group__btn-1" and select the elements using .findByDataTest().

src/components/ec-button-group/ec-button-group.spec.js
mcibique: I suggested to remove the types checking and these tests are redundant then. The button group always do $emit('change', button.value).

src/components/ec-button-group/ec-button-group.spec.js
jaimeperezpedra: I cannot see anything you are right, I will remove it

src/components/ec-button-group/ec-button-group.spec.js
jaimeperezpedra: Done!

src/components/ec-button-group/ec-button-group.spec.js
jaimeperezpedra: Done!

src/components/ec-button-group/ec-button-group.vue
jaimeperezpedra: Good point, working on it!

jaimeperezpedra: Solved

src/components/ec-button-group/ec-button-group.spec.js
```diff
@@ -0,0 +1,168 @@
...
+            { text: 'Yes', value: 'yes' },
+            { text: 'Maybe' },
+          ],
+        });
+        expect(errorSpy).toHaveBeenCalled();
+        expect(errorSpy.mock.calls[0][0]).toContain('Invalid prop: custom validator check failed for prop "items"');
+      });
+    });
+
+    it(':items - should throw an error when one or more items doesn\'t have value and text property', () => {
```
matallo: we've already tested this separately above, we can get rid of it

jaimeperezpedra: Done!

matallo:
Go go go

PR #76, opened 2020-01-27
quickfix:update border color for ec-input

stavros-tomas:
Update the border color for the text input fields. 

Dropdown search and dropdown are already correct.


jaimeperezpedra:
Don't forget to bump the version! Thx!

PR #77, opened 2020-01-28
Passing z-index to from ec-dropdown to ec-popover

stavros-tomas:
Pairing with @jaimeperezpedra on this.

While working on add supplier modal discovered that ec-dropdown not showing on modal.
We need to pass the level as a prop to the ec-dropdown--> ec-dropdowns-search--> ec-popover

Thoughts about this solution?


matallo:
Instinctively I wouldn't say I'm totally sold but it is true that popovers already work like that, LGTM then.

PR #84, opened 2020-02-11
ONL 4220 create suppliers loading state

jaimeperezpedra:
In order to add the loading state for the create suppliers, we need to add a loading state on the modal, and for that, we create a new status of the loading component.

### DIFF
--- a/src/components/ec-loading/ec-loading.spec.js
+++ b/src/components/ec-loading/ec-loading.spec.js
@@ -31,6 +31,18 @@ describe('EcLoading', () => {
     expect(wrapper.element).toMatchSnapshot();
   });
 
+  it('should render the loading with a background white if the transparent prop is set to false', () => {
+    const wrapper = mountLoading({ transparent: false });
+    expect(wrapper.find('.ec-loading__content--is-transparent').exists()).toBe(false);
+    expect(wrapper.element).toMatchSnapshot();
+  });
+
+  it('should not render the loading with a background white if the transparent prop is set to false and the show is set to false', () => {
+    const wrapper = mountLoading({ transparent: false, show: false });
+    expect(wrapper.find('.ec-loading__content--is-transparent').exists()).toBe(true);
+    expect(wrapper.element).toMatchSnapshot();
+  });
+
   it('should render the loading with a given icon size when the prop size is set to not 48', () => {
     const wrapper = mountLoading({ size: 30 });
     expect(wrapper.element).toMatchSnapshot();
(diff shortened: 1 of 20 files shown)


src/components/ec-loading/ec-loading.spec.js
```diff
@@ -31,6 +31,18 @@ describe('EcLoading', () => {
     expect(wrapper.element).toMatchSnapshot();
   });
 
+  it('should render the loading with a background white if the transparent prop is set to false', () => {
```
mcibique: How the test verifies that the background is white?

src/components/ec-loading/ec-loading.spec.js
```diff
@@ -31,6 +31,18 @@ describe('EcLoading', () => {
...
+    const wrapper = mountLoading({ transparent: false });
+    expect(wrapper.find('.ec-loading__content--is-transparent').exists()).toBe(false);
```
mcibique: We should check 2 classes are not there.

src/components/ec-loading/ec-loading.vue
```diff
@@ -31,6 +37,15 @@ export default {
...
       default: 48,
     },
+    transparent: {
+      type: Boolean,
+      default: true,
+    },
+  },
+  computed: {
+    isTransparent() {
+      return this.transparent || (!this.show && !this.transparent);
```
mcibique: This could be simplified to `this.transparent || !this.show`.  Try it yourself: https://www.dcode.fr/boolean-expressions-calculator.

src/components/ec-modal/ec-modal.story.js
```diff
@@ -37,6 +37,9 @@ stories
       positiveHasText: {
         default: boolean('Positive Has Text', true),
       },
+      isLoading: {
+        default: object('Loading Positive Button', { positive: true, negative: false }),
+      },
```
mcibique: I'd split this into two booleans. It's easier to check/uncheck a checkbox than editing JSON object in textarea.

you can then change the template to:
`:isLoading="{ positive: isPositiveLoading, negative: isNegativeLoading }"`

src/components/ec-loading/ec-loading.spec.js
jaimeperezpedra: Also, this is wrong I was calling it white but is without any background

src/components/ec-modal/ec-modal.story.js
jaimeperezpedra: Good call!

stavros-tomas:
@jaimeperezpedra  LGTM, the only objection for me is that when we apply the class "transparent" we make the items/content visible. Should have been the opposite as transparent means see-though almost like hidden. Can get bit confusing. Up to you

jaimeperezpedra:
@stavros-tomas The thing is the transparent property is for the loading, then when it is transparent means that you will show the things that are behind.

PR #86, opened 2020-02-13
ONL-4217: feat: Add interactive cards

jmesao:
Add interactive cards

### DIFF
--- a/src/scss/components/ec-card/_ec-card.scss
+++ b/src/scss/components/ec-card/_ec-card.scss
@@ -6,6 +6,17 @@
   background: $white;
   overflow: hidden;
 
+  &--is-interactive {
+    cursor: pointer;
+    transition: transform ease 0.4s, box-shadow ease 0.4s;
+  }
+
+  &--is-interactive:hover {
+    @include box-shadow-level-2;
+
+    transform: translateY(-8px);
+  }
+
   @include media__from-sm {
     padding: 24px;
   }
--- a/src/scss/components/ec-card/_ec-card.scss
+++ b/src/scss/components/ec-card/_ec-card.scss
@@ -1,3 +1,6 @@
+$ec-card-interactive-transition-duration: 0.4s;
+$ec-card-interactive-transition-timing-function: ease;
+
 .ec-card {
   @include box-shadow-level-1;
   @include shape-border-radius;
@@ -9,7 +12,7 @@
   &--is-interactive {
     position: relative;
     cursor: pointer;
-    transition: all 0.4s ease;
+    transition: transform $ec-card-interactive-transition-duration $ec-card-interactive-transition-timing-function;
 
     &:after {
       @include box-shadow-level-2;
@@ -22,7 +25,7 @@
       width: 100%;
       height: 100%;
       opacity: 0;
-      transition: all 0.4s ease;
+      transition: opacity $ec-card-interactive-transition-duration $ec-card-interactive-transition-timing-function;
     }
 
     &:hover {
--- a/src/scss/components/ec-card/_ec-card.scss
+++ b/src/scss/components/ec-card/_ec-card.scss
@@ -4,16 +4,33 @@
 
   padding: 8px;
   background: $white;
-  overflow: hidden;
+  word-break: break-word;
 
   &--is-interactive {
+    position: relative;
     cursor: pointer;
-    transition: transform ease 0.4s, box-shadow ease 0.4s;
+    transition: all 0.4s ease;
 
-    &:hover {
+    &:after {
       @include box-shadow-level-2;
 
+      content: '';
+      position: absolute;
+      z-index: -1;
+      top: 0;
+      left: 0;
+      width: 100%;
+      height: 100%;
+      opacity: 0;
+      transition: all 0.4s ease;
+    }
+
+    &:hover {
       transform: translateY(-8px);
+
+      &:after {
+        opacity: 1;
+      }
     }
   }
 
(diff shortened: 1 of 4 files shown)


src/scss/components/ec-card/_ec-card.scss
```diff
@@ -6,6 +6,17 @@
   background: $white;
   overflow: hidden;
 
+  &--is-interactive {
+    cursor: pointer;
+    transition: transform ease 0.4s, box-shadow ease 0.4s;
```
jaimeperezpedra: Maybe this can be a mixin, @mcibique ?

src/scss/components/ec-card/_ec-card.scss
```diff
@@ -6,6 +6,17 @@
...
+  }
+
+  &--is-interactive:hover {
```
mcibique: I'd join these with the rules above

```
&--is-interactive {
  // ...
  &:hover {
     // ...
  }
}

File: src/scss/components/ec-card/_ec-card.scss (hunk 1 above)
Comment (mcibique) [reply src/scss/components/ec-card/_ec-card.scss]: I'm not sure. I haven't seen any other component which could reuse it, but it could be extracted into something like this:
```
@mixin interactive-transition {
  @include box-shadow-level-1;

  cursor: pointer;
  transition: transform ease 0.4s, box-shadow ease 0.4s;

  &:hover {
    @include box-shadow-level-2;

    transform: translateY(-8px);
  }
}
```

I'd rather focus on performance of this thing, because transitioning box-shadow is always super slow everywhere, check this out: https://tobiasahlin.com/blog/how-to-animate-box-shadow/.

File: src/scss/components/ec-card/_ec-card.scss (hunk 2 above)
Comment (cukorlyon87) [reply src/scss/components/ec-card/_ec-card.scss]: Totally agree, thanks.

File: src/scss/components/ec-card/_ec-card.scss (hunk 1 above)
Comment (cukorlyon87) [reply src/scss/components/ec-card/_ec-card.scss]: @jaimeperezpedra @mcibique I don't think that we can reuse it, it's too specific, so this time I'd prefer not to do a mixin.

Comment (cukorlyon87) [reply src/scss/components/ec-card/_ec-card.scss]: I applied the solution you shared, many thanks.

File: src/scss/components/ec-card/_ec-card.scss [hunk 3]
```diff
@@ -4,7 +4,35 @@
...
+    position: relative;
+    cursor: pointer;
+    transition: all 0.4s ease;
+
+    &:after {
+      @include box-shadow-level-2;
+
+      content: '';
+      position: absolute;
+      z-index: -1;
```
jaimeperezpedra: Why this is needed, and we have a file where we define all the z-index, if this one is needed you can create one and used it here

src/scss/components/ec-card/_ec-card.scss
matallo: this is one of the few cases where I'd still encourage nesting

src/scss/components/ec-card/ec-card.story.js
```diff
@@ -15,6 +15,17 @@ stories
     </div>
       `,
   }))
+  .add('interactive card', () => ({
+    template: `
+    <div style="display: flex; height: 100vh; margin: 10px">
```
matallo: wouldn't we want to have an interactive card with an `a` tag?

cukorlyon87: @matallo The UX part does not have _text-decoration_ when we do hover or some other properties, so we would also need to cancel those properties that tag _a_ has predefined, anyway this could be to the tastes of consumers because it is a sass component and it is not a Vue component forcing a markup defined.

src/scss/components/ec-card/_ec-card.scss
cukorlyon87: @jaimeperezpedra This is following the link shared by Milos, putting a pseudo-element behind and playing with the opacity having a better performance.

src/scss/components/ec-card/_ec-card.scss
```diff
@@ -4,7 +4,35 @@
 
   padding: 8px;
   background: $white;
-  overflow: hidden;
+  word-break: break-word;
+
+  &--is-interactive {
+    position: relative;
+    cursor: pointer;
+    transition: all 0.4s ease;
```
mcibique: We should not use `transition: all` because that's not very future proof. It will cause colors, paddings, margins to transition too, which is not a case right now but it might change in the future.

Just specify the property you have intention to do the transition for, just like you had it in the initial commit.

src/scss/components/ec-card/_ec-card.scss
```diff
@@ -4,7 +4,35 @@
...
+
+      content: '';
+      position: absolute;
+      z-index: -1;
+      top: 0;
+      left: 0;
+      width: 100%;
+      height: 100%;
+      opacity: 0;
+      transition: all 0.4s ease;
```
mcibique: Same here. Plus you repeated `0.4s ease` 2 times, I'd extract it into a variable (e.g. `ec-card-interactive-transition`) because the transitions should be in sync, right? Do we expect having different transition for opacity and different for transform?

src/scss/components/ec-card/_ec-card.scss
jmesao: Good point @mcibique I'll do it, thanks!

src/scss/components/ec-card/_ec-card.scss
jmesao: @mcibique At first, we expect to have the same, done, thanks!

PR #87, opened 2020-02-24
Onl 4266 button component

stavros-tomas:
A button component that will work hand in hand with the chameleon styles component.

### DIFF
--- a/src/components/ec-btn/ec-btn.spec.js
+++ b/src/components/ec-btn/ec-btn.spec.js
@@ -1,0 +1,200 @@
+import { mount, createLocalVue } from '@vue/test-utils';
+import EcBtn from './ec-btn.vue';
+
+describe('EcBtn', () => {
+  function mountBtn(props, mountOpts) {
+    const localVue = createLocalVue();
+
+    return mount(EcBtn, {
+      localVue,
+      propsData: { ...props },
+      ...mountOpts,
+    });
+  }
+
+  describe(':props', () => {
+    it('should render a <button> element by default', () => {
+      const wrapper = mountBtn();
+
+      expect(wrapper.find('.ec-btn').exists()).toBe(true);
+      expect(wrapper.element).toMatchSnapshot();
+    });
+
+    it('should render a <button> element with class ".ec-btn--sm" when size is set to "sm"', () => {
+      const wrapper = mountBtn({
+        size: 'sm',
+      });
+
+      expect(wrapper.find('.ec-btn--md').exists()).toBe(false);
+      expect(wrapper.find('.ec-btn--sm').exists()).toBe(true);
+      expect(wrapper.element).toMatchSnapshot();
+    });
+
+    it('should render a <button> element with class ".ec-btn--md" when size is set to "md"', () => {
+      const wrapper = mountBtn({
+        size: 'md',
+      });
+
+      expect(wrapper.find('.ec-btn--sm').exists()).toBe(false);
+      expect(wrapper.find('.ec-btn--md').exists()).toBe(true);
+      expect(wrapper.element).toMatchSnapshot();
+    });
+
+    it('should render an <a> element when we define the "href" prop', () => {
+      const wrapper = mountBtn({
+        href: 'https://ebury.com/',
+      });
+
+      expect(wrapper.attributes().href).toBe('https://ebury.com/');
+      expect(wrapper.element).toMatchSnapshot();
+    });
+
+    it('should render a router link when we define the "to" prop', () => {
+      const wrapper = mountBtn({
+        to: 'trade-finance',
+      },
+      {
+        stubs: ['router-link'],
+      });
+
+      expect(wrapper.attributes().to).toBe('trade-finance');
+      expect(wrapper.element).toMatchSnapshot();
+    });
+
+    it('should render a router link with a custom tag when we define the "tag option"', () => {
+      const wrapper = mountBtn({
+        to: 'trade-finance',
+        tag: 'ebury',
+      },
+      {
+        stubs: ['ebury'],
+      });
+
+      expect(wrapper.element).toMatchSnapshot();
+    });
+
+    it('should render a disabled button when disabled is set to true', () => {
+      const wrapper = mountBtn({
+        disabled: true,
+      });
+
+      expect(wrapper.attributes().disabled).toBe('disabled');
+      expect(wrapper.element).toMatchSnapshot();
+    });
+
+    it('should render a button with an icon when the icon prop is defined', () => {
+      const wrapper = mountBtn({
+        icon: 'simple-check',
+      });
+
+      expect(wrapper.find('.ec-btn__icon').exists()).toBe(true);
+      expect(wrapper.element).toMatchSnapshot();
+    });
+
+    it('should render a button with an outline when "outline" prop is set to true', () => {
+      const wrapper = mountBtn({
+        outline: true,
+      });
+
+      expect(wrapper.find('.ec-btn--outline').exists()).toBe(true);
+      expect(wrapper.element).toMatchSnapshot();
+    });
+
+    it('should render a full width button when "fullWidth" prop is set to true', () => {
+      const wrapper = mountBtn({
+        fullWidth: true,
+      });
+
+      expect(wrapper.find('.ec-btn--full-width').exists()).toBe(true);
+      expect(wrapper.element).toMatchSnapshot();
+    });
+
+    it.each(['primary', 'secondary', 'success', 'error', 'warning'])('should render a button with type "%s" when type prop is set', (type) => {
+      const wrapper = mountBtn({
+        type,
+      });
+
+      expect(wrapper.find(`.ec-btn--${type}`).exists()).toBe(true);
+      expect(wrapper.element).toMatchSnapshot();
+    });
+
+    it('should render a reverse button when reverse is set to true', () => {
+      const wrapper = mountBtn({
+        reverse: true,
+      });
+
+      expect(wrapper.find('.ec-btn--reverse').exists()).toBe(true);
+      expect(wrapper.element).toMatchSnapshot();
+    });
+
+    it('should render a button with a spinner loader if loading is set to true', () => {
+      const wrapper = mountBtn({
+        loading: true,
+      });
+
+      expect(wrapper.attributes().disabled).toBe('disabled');
+      expect(wrapper.find('.ec-btn__spinner').exists()).toBe(true);
+      expect(wrapper.element).toMatchSnapshot();
+    });
+
+    it('should render a button with type submit if submit is set to true', () => {
+      const wrapper = mountBtn({
+        submit: true,
+      });
+
+      expect(wrapper.attributes().type).toBe('submit');
+      expect(wrapper.element).toMatchSnapshot();
+    });
+  });
+
+  describe(':slots', () => {
+    it('should render a button with "Click me!" text', () => {
+      const wrapper = mountBtn(
+        {},
+        {
+          slots: {
+            default: 'Click Me!',
+          },
+        },
+      );
+      expect(wrapper.find('.ec-btn__spinner').exists()).toBe(false);
+      expect(wrapper.findByDataTest('ec-btn__text-loader-slot').exists()).toBe(false);
+      expect(wrapper.findByDataTest('ec-btn__default-slot').exists()).toBe(true);
+      expect(wrapper.element).toMatchSnapshot();
+    });
+
+    it('should render a loading button with a spinner loader and a "Click me!" text set to invisble', () => {
+      const wrapper = mountBtn(
+        {
+          loading: true,
+        },
+        {
+          slots: {
+            default: 'Click Me!',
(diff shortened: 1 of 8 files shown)


src/components/ec-btn/ec-btn.vue
```diff
@@ -0,0 +1,169 @@
...
+        disabled: !this.isNormalLink && (this.disabled || this.loading),
+      };
+    },
+  },
+};
+</script>
+
+<style lang="scss">
+  .ec-btn {
+    &__content-invisible {
```
stavros-tomas: @mcibique not sure if "invisible" looks ok, found it the most appropriate. Let me know.

src/components/ec-btn/ec-btn.vue
```diff
@@ -0,0 +1,169 @@
...
+    },
+    outline: {
+      type: Boolean,
+      default: false,
+    },
+    fullWidth: {
+      type: Boolean,
+      default: false,
+    },
+    type: {
```
stavros-tomas: @matallo @jaimeperezpedra. We were discussing this with @mcibique. Shall we rename from "type" to "category"? Type seems bit vague :/

src/scss/components/ec-btn/ec-btn.story.js
```diff
@@ -4,7 +4,7 @@ import StoryRouter from 'storybook-vue-router';
 import EcIcon from '@/components/ec-icon';
 import { LIGHT_THEME, DARK_THEME } from '../../../../.storybook/backgrounds';
 
-const stories = storiesOf('Button', module);
+const stories = storiesOf('Button Styles', module);
```
stavros-tomas: @mcibique @matallo @jaimeperezpedra As we have 3 different button stories I renamed the old one "Button styles" since is not a component, and the new one I named it "button component" so be a clear distinction of what is what. Another approach can be to completely delete the other story. Let me know your thoughts.

src/components/ec-btn/ec-btn.story.js
```diff
@@ -0,0 +1,163 @@
...
+import StoryRouter from 'storybook-vue-router';
+import EcBtn from './ec-btn.vue';
+
+const stories = storiesOf('Button component', module).addDecorator(StoryRouter());
+
+stories
+  .add('basic', () => ({
+    components: { EcBtn },
+    props: {
+      roundedFromProps: {
```
mcibique: You need `FromProps` suffix only if the value you are passing to the component is v-model. Component cannot change value of any of these props. You could name them just `rounded`, `outline`, 'fullWidth' and then use it in the template:
```
<ec-btn v-bind="$props" ... />
```

src/components/ec-btn/ec-btn.story.js
```diff
@@ -0,0 +1,163 @@
...
+          :text="textFromProps"
+          :type="typeFromProps"
+          class="ec-mt--20"
+          to="tradeFinance"
+          >
+            {{textFromProps}}
+        </ec-btn>
+
+        <ec-btn
+          v-if="iconFromProps.length > 0"
```
mcibique: `v-if="iconFromProps"`

src/components/ec-btn/ec-btn.story.js
```diff
@@ -0,0 +1,163 @@
...
+          :icon="iconFromProps"
+          :outline="outlineFromProps"
+          :reverse="reverseFromProps"
+          :size="sizeFromProps"
+          :type="typeFromProps"
+          class="ec-ml--20 ec-mt--20"
+          to="tradeFinance"
+        />
+
+        <h3 class="ec-mt--20">normal link - a tag</h3>
```
mcibique: I'd rather name it "anchor link" and avoid the word "normal".

src/components/ec-btn/ec-btn.spec.js
```diff
@@ -0,0 +1,200 @@
...
+    it('should render a full width button when "fullWidth" prop is set to true', () => {
+      const wrapper = mountBtn({
+        fullWidth: true,
+      });
+
+      expect(wrapper.find('.ec-btn--full-width').exists()).toBe(true);
+      expect(wrapper.element).toMatchSnapshot();
+    });
+
+    it.each(['primary', 'secondary', 'success', 'error', 'warning'])('should render a button with type "%s" when type prop is set', (type) => {
```
matallo: actually, to the previous comment, we could test the sizes following this same approach

src/components/ec-btn/ec-btn.story.js
```diff
@@ -0,0 +1,163 @@
...
+          :text="textFromProps"
+          :type="typeFromProps"
+          class="ec-mt--20"
+          href="http://www.ebury.com"
+          >
+            {{textFromProps}}
+        </ec-btn>
+
+        <ec-btn
+          v-if="iconFromProps.length > 0"
```
mcibique: `v-if="iconFromProps"`

src/components/ec-btn/ec-btn.story.js
```diff
@@ -0,0 +1,163 @@
...
+          :disabled="disabledFromProps"
+          :full-width="fullWidthFromProps"
+          :icon="iconFromProps"
+          :outline="outlineFromProps"
+          :reverse="reverseFromProps"
+          :size="sizeFromProps"
+          :type="typeFromProps"
+          class="ec-ml--20 ec-mt--20"
+            href="http://www.ebury.com"
```
mcibique: check the indentation.

src/components/ec-btn/ec-btn.spec.js
```diff
@@ -0,0 +1,200 @@
...
+
+    return mount(EcBtn, {
+      localVue,
+      propsData: { ...props },
+      ...mountOpts,
+    });
+  }
+
+  describe(':props', () => {
+    it('should render a <button> element by default', () => {
```
mcibique: This doesn't test any prop, we can move it outside of the `:props`.

src/components/ec-btn/ec-btn.spec.js
```diff
@@ -0,0 +1,200 @@
...
+      const wrapper = mountBtn();
+
+      expect(wrapper.find('.ec-btn').exists()).toBe(true);
```
mcibique: I think the wrapper is `'ec-btn`, so you don't need to do the `.find()`.

src/components/ec-btn/ec-btn.spec.js
```diff
@@ -0,0 +1,200 @@
...
+      expect(wrapper.element).toMatchSnapshot();
+    });
+
+    it('should render a <button> element with class ".ec-btn--sm" when size is set to "sm"', () => {
+      const wrapper = mountBtn({
+        size: 'sm',
+      });
+
+      expect(wrapper.find('.ec-btn--md').exists()).toBe(false);
+      expect(wrapper.find('.ec-btn--sm').exists()).toBe(true);
```
mcibique:
```
      expect(wrapper.classes('ec-btn--md')).toBe(false);
      expect(wrapper.classes('ec-btn--sm')).toBe(true);
```

src/components/ec-btn/ec-btn.spec.js
```diff
@@ -0,0 +1,200 @@
...
+      expect(wrapper.element).toMatchSnapshot();
+    });
+
+    it('should render a <button> element with class ".ec-btn--md" when size is set to "md"', () => {
+      const wrapper = mountBtn({
+        size: 'md',
+      });
+
+      expect(wrapper.find('.ec-btn--sm').exists()).toBe(false);
+      expect(wrapper.find('.ec-btn--md').exists()).toBe(true);
```
mcibique: expect(wrapper.classes('ec-btn--md')).toBe(true);
      expect(wrapper.classes('ec-btn--sm')).toBe(false);

src/components/ec-btn/ec-btn.spec.js
```diff
@@ -0,0 +1,200 @@
...
+      expect(wrapper.element).toMatchSnapshot();
+    });
+
+    it('should render an <a> element when we define the "href" prop', () => {
+      const wrapper = mountBtn({
+        href: 'https://ebury.com/',
+      });
+
+      expect(wrapper.attributes().href).toBe('https://ebury.com/');
```
mcibique: you can also test:
```
expect(wrapper.is('a')).toBe(true);
```
if you want to match the description of the test.

src/components/ec-btn/ec-btn.spec.js
```diff
@@ -0,0 +1,200 @@
...
+      expect(wrapper.find('.ec-btn__icon').exists()).toBe(true);
+      expect(wrapper.element).toMatchSnapshot();
+    });
+
+    it('should render a button with an outline when "outline" prop is set to true', () => {
+      const wrapper = mountBtn({
+        outline: true,
+      });
+
+      expect(wrapper.find('.ec-btn--outline').exists()).toBe(true);
```
mcibique: `expect(wrapper.classes('.ec-btn--outline')).toBe(true);`

please update it everywhere in this spec file.

src/components/ec-btn/ec-btn.spec.js
```diff
@@ -0,0 +1,200 @@
...
+        reverse: true,
+      });
+
+      expect(wrapper.find('.ec-btn--reverse').exists()).toBe(true);
+      expect(wrapper.element).toMatchSnapshot();
+    });
+
+    it('should render a button with a spinner loader if loading is set to true', () => {
+      const wrapper = mountBtn({
+        loading: true,
```
matallo: I find boolean state props better named `isSomething` in this case `isLoading` ?

src/components/ec-btn/ec-btn.vue
matallo: I'd approach this differently, instead of modifiers I'd use states with `.is-hidden` or `.has-text` for these two variations below

src/components/ec-btn/ec-btn.spec.js
```diff
@@ -0,0 +1,200 @@
...
+      const wrapper = mountBtn(
+        {},
+        {
+          slots: {
+            default: 'Click Me!',
+          },
+        },
+      );
+      expect(wrapper.find('.ec-btn__spinner').exists()).toBe(false);
+      expect(wrapper.findByDataTest('ec-btn__text-loader-slot').exists()).toBe(false);
```
mcibique: You can user `findByDataTest` for other tests too. Many `.find()` calls can be replaced with `.classes()`,  rest of them can be done via `findByDataTest()`. It's more preferable so we do not tie-up our classes to the tests.

src/scss/components/ec-btn/ec-btn.story.js
matallo: LGTM, in the future we'll have Chameleon styles, and chameleon components which will be grouped separately in Storybook

src/components/ec-btn/ec-btn.spec.js
```diff
@@ -0,0 +1,200 @@
...
+      expect(wrapper.findByDataTest('ec-btn__default-slot').exists()).toBe(true);
+      expect(wrapper.element).toMatchSnapshot();
+    });
+
+    it('should render a loading button with a spinner loader and a "Click me!" text set to invisble', () => {
```
mcibique: We cannot check something is invisible in unit tests. We don't have the CSS.

src/components/ec-btn/ec-btn.vue
stavros-tomas: In this case I want it as a class because I don't want to remove it from the DOM with a v-if I want it to remain in the DOM and make it invisible (so the content is still there to force the button to remain the same size)

src/components/ec-btn/ec-btn.vue
```diff
@@ -0,0 +1,169 @@
...
+      v-else-if="this.$slots.default"
+      data-test="ec-btn__default-slot"
+      :class="{'ec-btn__content-invisible' : loading }"
+    >
+      <slot />
+    </span>
+
+    <ec-loading
+      v-if="loading && !this.$slots['text-loader']"
+      class="ec-btn__spinner"
```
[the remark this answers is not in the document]
stavros-tomas: I mostly agree, with a small amend to call it `ec-btn_loading-spinner` if you don't mind like we will do the other one "loading-text". Loading alone sounds like a state and not an element.

src/components/ec-btn/ec-btn.vue
```diff
@@ -0,0 +1,169 @@
...
+export default {
+  name: 'EcBtn',
+  components: {
+    EcIcon,
+    EcLoading,
+  },
+  props: {
+    size: {
+      type: String,
+      default: 'md',
```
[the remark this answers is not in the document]
stavros-tomas: We didn't but comes from UX team. Default is 'md'. I think we use only in one place "sm".

src/components/ec-btn/ec-btn.spec.js
stavros-tomas: I kept the test but rephrased the description. I would like to test if the default slot has the loading class (so should be invisible unless someone amends the class)

PR #89, opened 2020-02-25
ONL 4271 Amount Input Component

jaimeperezpedra:
This is the amount-input component, this component will be used when we need to get an amount in any form.

The group separator and the decimal separator will be calculated with the locale prop if not will be with the default options.

We can also want the string of this input, for that we can set the masked prop to true and we will get the string of that input, with the group separator and the decimal separator.


src/components/ec-amount-input/ec-amount-input.vue
```diff
@@ -0,0 +1,119 @@
...
+    event: 'change',
+  },
+  props: {
+    ...EcInputField.props,
+    value: {
+      validator: prop => typeof prop === 'number' || typeof prop === 'string' || prop === null,
+      required: true,
+      default: null,
+    },
+    masked: {
```
matallo: `isMasked`?

matallo:
Also, bump `package.json`

src/directives/ec-money/utils.spec.js
```diff
@@ -0,0 +1,70 @@
+import { format, unFormat, sanitizeInput } from './utils';
```
mcibique: We should check misplaced group separators e.g. `1,1,1,1,` or `111,,111`
and also check multiple decimal separators e.g. `1..11` or `1.1.11`

src/directives/ec-money/utils.spec.js
```diff
@@ -0,0 +1,70 @@
...
+      ['1,111.11', '111111'],
+      ['1.11', '111'],
+      ['1.1111', '11111'],
+      ['1,111,111.00', '111111100'],
+      ['1`111`111.00', '111111100'],
+      ['1,111,111`00', '111111100'],
+      ['£1,111,111.00', '111111100'],
+      ['1,111,111.00£', '111111100'],
+      ['1,111,111.001', '1111111001'],
+    ])('should sanitize the u %s into a %s', (value, valueExpected) => {
```
mcibique: `the u`?

src/main.js
```diff
@@ -1,4 +1,5 @@
 export { default as EcAlert } from './components/ec-alert';
+export { default as EcAmountInput } from './components/ec-amount-input';
```
mcibique: We should export the directive too.

PR #92, opened 2020-03-03
ONL-4310 input note

jaimeperezpedra:


src/components/ec-input-field/ec-input-field.spec.js
```diff
@@ -53,6 +54,11 @@ describe('EcInputField', () => {
     expect(wrapper.element).toMatchSnapshot();
   });
 
+  it('renders properly without note', () => {
+    const wrapper = mountInputField({ note: '' });
```
matallo: is this the best approach, rather than not passing the prop?

src/components/ec-input-field/ec-input-field.vue
```diff
@@ -153,7 +167,19 @@ $ec-input-field-invalid-color: $color-error !default;
...
   &__label {
+    display: flex;
+    flex-wrap: wrap;
+  }
+
+  &__label-text {
     @include input-label;
+
+    flex-grow: 1;
+    margin-right: 6px;
```
stavros-tomas: trivial, multiple of 4.

src/components/ec-input-field/ec-input-field.story.js
```diff
@@ -20,6 +20,9 @@ stories
       labelNumber: {
         default: text('label', 'Number input', GROUPS.NUMBER),
       },
+      noteNumber: {
+        default: text('note', 'Max 80 char', GROUPS.NUMBER),
```
mcibique: Max 80 chars

src/components/ec-input-field/ec-input-field.story.js
```diff
@@ -32,6 +35,9 @@ stories
       labelText: {
         default: text('label', 'Text input', GROUPS.TEXT),
       },
+      noteText: {
+        default: text('note', 'Max 80 char', GROUPS.TEXT),
```
mcibique: Max 80 chars

src/components/ec-input-field/ec-input-field.story.js
```diff
@@ -44,6 +50,9 @@ stories
       labelDate: {
         default: text('label', 'Date input', GROUPS.DATE),
       },
+      noteDate: {
+        default: text('note', 'Max 80 char', GROUPS.DATE),
```
mcibique: Max 80 chars

src/components/ec-input-field/ec-input-field.spec.js
matallo: thoughts?

jaimeperezpedra: If we don't pass the prop it will be filled, we want to put it with an empty string because we declared with a string on the function.

I think it can be done in both ways, don't fill on the function and fill in every test or fill on the function and put an empty string on the tests, but I just follow the current implementation.

src/components/ec-input-field/ec-input-field.vue
```diff
@@ -1,14 +1,30 @@
...
+        v-if="note"
+        class="ec-input-field__note"
+        data-test="ec-input-field__note"
+      >{{ note }}</span>
+    </label>
     <input
       :id="id"
       v-model="inputModel"
       class="ec-input-field__input"
+      :data-test="$attrs['data-test'] ? $attrs['data-test'] : 'ec-input-field__input'"
```
jaimeperezpedra: I don't know if this is okay or not, but I couldn't find another solution to the problem.

The problem is, for example in the dropdown we set a data-test on the ec-input-field and we want to have this data-test on the input of the ec-input-field, if we set the data-test to this input without any validation we never gonna get the data test from the parent.

mcibique: First of all, we need to preserve both, so it should be `${$attrs['data-test']} ec-input-field__input`, then we need to create something more reusable, because we are going to have the same problem with every component, e.g.

```
<ec-alert data-test="my-alert" />
```
^^ will override `data-test="ec-alert"` set inside of `ec-alert.vue`. In previous project I used a directive v-merge-data-test="'ec-alert'", which can merge everything the same way. It's not hard to implement, we just need to catch all cases, e.g. multiple nested components:
```
<ebo-suppliers-dropdown data-test="one" />
contains in template:
<ec-dropdown data-test="two" />
contains in template:
<ec-input-field data-test="three" />
```
should output `<input data-test="one two three" >`

src/components/ec-input-field/ec-input-field.vue
```diff
@@ -1,14 +1,30 @@
...
+        v-if="note"
+        class="ec-input-field__note"
+        data-test="ec-input-field__note"
+      >{{ note }}</span>
+    </label>
     <input
       :id="id"
       v-model="inputModel"
       class="ec-input-field__input"
+      :data-test="$attrs['data-test'] ? `${$attrs['data-test']} ec-input-field__input` : 'ec-input-field__input'"
```
matallo:
```suggestion
      data-test="ec-input-field__input"
      :data-test="$attrs['data-test']"
```

would achieve the same purpose?

jaimeperezpedra: This is not working, data-test doesn't work in the same way that the classes and the styles

Good suggestion anyway I didn't now either! :)

matallo: oops, good to know! not a Vue expert :blush:

PR #93, opened 2020-03-05
ONL-4273 Table-sort

mcibique:
example of using vue-hoc

3 **BREAKING CHANGES** introduced in this PR:
1. data-test for table cells renamed from `ec-table__cell-X` to `ec-table__cell ec-table__cell--X` so we can select all table cells or one specific.
2. data-test for table row renamed from `ec-table__row-X` to `ec-table__row ec-table__row--X` so we can select all table rows or one specific.
3. columns definition now contains name and title props, previously we only had name. Name contained the header title of the column, but that has changed. We will use name as a unique identifier of the column which will be used for sorting and filtering. The header title should be passed, from now on, in the title property. The name will be passed to the backend in the API request.


matallo:
first reaction, before any actual review is while you were describing the _smart_ table I was thinking in a HOC, happy to see this implemented :+1:

src/components/ec-table-sort/ec-table-sort.vue
```diff
@@ -0,0 +1,120 @@
...
+    },
+    isAsc() {
+      return this.direction === SortDirection.ASC;
+    },
+    isDesc() {
+      return this.direction === SortDirection.DESC;
+    },
+    directionTitle() {
+      if (this.isAsc) {
+        return 'Sorted ascending';
```
matallo: Would we want to add i18n in the future?

src/components/ec-table/ec-table.vue
```diff
@@ -5,16 +5,21 @@
...
+      <ec-table-head
+        :columns="columns"
+        :sorts="sorts"
+        @sort="onSort"
+      />
       <tbody>
         <tr
           v-for="(row, rowIndex) in data"
           :key="rowIndex"
+          :data-test="'ec-table__row ec-table__row--' + rowIndex"
```
matallo: candidate for template literals?

matallo:
I really like the idea of using name attributes for the events, when I mentioned in other PRs (tab panes) about assuming indexes and the possibility of handle that differently I was thinking of something like this, kudos

mcibique:
> The implementation seems flawless but the `ec-table-sort` naming was a bit misleading, is "sort" a child element? `ec-table-thead-sortable` would be more eloquent (just throwing ideas here)

yes, ec-table-sort is a child of ec-table-head. I didn't want to name it sort-trigger or sort-indicator, because that's what it is. It contains text and the icon which is changed based on the sorting direction.

src/components/ec-table-sort/ec-table-sort.vue
mcibique: Yeah, it's kind of complicated passing these 3 texts through table and table-head components. Now I regret adding this. I wanted to do it only for accessibility.

mcibique:
WIP is over, you can review this one.

src/components/ec-table/ec-table.vue
mcibique: done.

matallo:
I'm not entirely convinced about the multi-sort (I might be wrong) and I couldn't find anything regarding this in the product description https://fxsolutions.atlassian.net/browse/ONL-4273

Really appreciate laying the ground for this possibility, but I'm not sure it will eventually be used. Can you confirm with Product and UX?

src/components/ec-table-sort/ec-table-sort.spec.js
```diff
@@ -0,0 +1,68 @@
...
+      const wrapper = mountEcTableSort({ direction: '' });
+      expect(wrapper.element).toMatchSnapshot();
+    });
+
+    it('should render with direction set to ASC', () => {
+      const wrapper = mountEcTableSort({ direction: SortDirection.ASC });
+      expect(wrapper.element).toMatchSnapshot();
+    });
+
+    it('should render with direction set to DESC', () => {
```
matallo: good example where duplication is better than abstraction (having both specs in an array) :+1:

src/components/ec-table-sort/ec-table-sort.vue
matallo: let's keep it but I don't like these leftovers (shall we at least open a bug to keep track?)

mcibique:
To me, the single sort is just a special case of multi sort, so I coded it like you can always pass multiple sorts. Also maintaining the logic is not hard at all because it is only 2 lines of code:
```
if (this.multiSort) {
        sorts.push({ column: columnName, direction: SortDirection.ASC });
      } else {
        sorts = [{ column: columnName, direction: SortDirection.ASC }];
      }
```

matallo:
I'll trust your judgement, then

I realizaed outside the smart table case the carets still appear, is this expected? It seems confusing to me (even if the columns are indeed sorted, I'd only show the icons for the sortable table)

I reckon this task is for the sort component, are we creating the sorting implementation in a different PR?

mcibique:
I enabled sorting via `sorts` property in all stories, so you can play with that in knobs, but the table is dumb, so if you click on the caret, nothing happens. I can remove that and keep only one example in basic and smart.

PR #98, opened 2020-03-13
feat(ONL-4295): Adding back button to EcPanel

jafcalvente:
- This story is for adding to the EcPanel the possibility of displaying (is optional) a new button in the top-left corner, with a left arrow icon.

- This button just will emit an event to point out that user wants to go back to a previous panel.

- To display it in Storybook, go to Panel basic component, and mark "With the back button visible" in the Knobs tab. 

image

### DIFF
--- a/src/components/ec-panel/ec-panel.spec.js
+++ b/src/components/ec-panel/ec-panel.spec.js
@@ -39,15 +39,15 @@ describe('EcPanel', () => {
       expect(wrapper.element).toMatchSnapshot();
     });
 
-    it(':goBackEnabled - should render the back button within the panel when is true', () => {
-      const wrapper = mountPanel({ show: true, goBackEnabled: true });
+    it(':isBackEnabled - should render the back button within the panel when is true', () => {
+      const wrapper = mountPanel({ show: true, isBackEnabled: true });
 
       expect(wrapper.find('.ec-panel__header-action--back').exists()).toBe(true);
       expect(wrapper.element).toMatchSnapshot();
     });
 
-    it(':goBackEnabled - should not render the back button within the panel when is false', () => {
-      const wrapper = mountPanel({ show: true, goBackEnabled: false });
+    it(':isBackEnabled - should not render the back button within the panel when is false', () => {
+      const wrapper = mountPanel({ show: true, isBackEnabled: false });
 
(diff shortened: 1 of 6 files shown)


src/components/ec-panel/ec-panel.spec.js
```diff
@@ -38,15 +38,38 @@ describe('EcPanel', () => {
       expect(wrapper.find('.ec-panel').exists()).toBe(false);
       expect(wrapper.element).toMatchSnapshot();
     });
+
+    it(':goBackEnabled - should render the back button within the panel when is true', () => {
+      const wrapper = mountPanel({ show: true, goBackEnabled: true });
```
matallo: these props shoudl be named `isSomething`, ask @jaimeperezpedra for more context

jafcalvente: Amended.

src/components/ec-panel/ec-panel.vue
```diff
@@ -36,24 +49,37 @@ export default {
...
     prop: 'show',
-    event: 'close',
+    event: 'show-panel',
   },
   props: {
     show: {
       type: Boolean,
       default: false,
     },
+    isBackEnabled: {
```
mcibique: I wouldn't make this as a prop. I would convert it to a computed prop and check if $listeners['back'] exists inside. Then everything you have to do to enable back functionality is to `<ec-panel @back="doSomething()">`

src/components/ec-panel/ec-panel.vue
```diff
@@ -5,14 +5,27 @@
...
+          <ec-icon
+            class="ec-panel__header-icon"
+            name="simple-chevron-left"
+            :size="24"
+          />
+        </a>
         <a
           aria-label="Close panel"
-          class="ec-panel__close"
+          class="ec-panel__header-action ec-panel__header-action--close"
```
mcibique: OT: While you are here, could you please add data-test on everything clickable in this component? We have lots of references to these classes in integration tests and unit tests, we should replace them with data-test.

jafcalvente: Well spotted. Done.

src/components/ec-panel/ec-panel.vue
jafcalvente: Smarter approach. Done.

PR #99, opened 2020-03-20
ONL-4270: currencies without decimals

mcibique:
Two things have been fixed:

1. Amount input now properly handles currencies without decimals and doesn't allow typing decimal separator.
2. Negative sign in amount input was causing weird behaviours, e.g. you could type `-` in the middle of a number; or you have to press `-` twice in empty input to get at least one `-` in the input.


mcibique:
Merged without bump, amount input is not used yet.

PR #100, opened 2020-03-20
ONL-4069 add z-index to the modal component

jaimeperezpedra:
In this PR I add the prop z-index to the modal component to solve the ONL-4069 issue.

Check the comments on the code to see if all agree on have the z-index prop fixed between 200-250.

### DIFF
--- a/src/components/ec-modal/ec-modal.spec.js
+++ b/src/components/ec-modal/ec-modal.spec.js
@@ -1,4 +1,5 @@
 import { mount, createLocalVue } from '@vue/test-utils';
+import { withMockedConsole } from '../../../tests/utils/console';
 import EcModal from './ec-modal.vue';
 
 jest.mock('../../directives/ec-focus-trap');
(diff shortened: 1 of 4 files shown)


src/components/ec-modal/ec-modal.vue
```diff
@@ -205,6 +216,13 @@ $ec-modal-close-btn-fill-hover: $level-4-tech-blue !default;
 
   @include z-index-modal;
 
+  // The z-index will be on between 200 and 250 because the 250 is the z-index for the loading
+  @for $i from 200 through 250 {
```
mcibique: We are not going to generate 50 CSS classes and then rarely use them. 

We can use a `style` tag for this. Nobody is going to override this property once is set.

Or we can use the same level mechanism as with tooltips, using predefined levels.

src/components/ec-modal/ec-modal.spec.js
```diff
@@ -72,6 +73,30 @@ describe('EcModal', () => {
     expect(wrapper.find('.ec-modal__content').element).toMatchSnapshot();
   });
 
+  it('should have the ec-modal--z-index-X class when the z-index props is given', () => {
```
matallo: does this statement make sense after last changes?

jaimeperezpedra: true

src/components/ec-modal/ec-modal.spec.js
```diff
@@ -72,6 +73,30 @@ describe('EcModal', () => {
...
+  it('should have the style attribute z-index when the z-index props is given', () => {
+    const wrapper = mountModal({
+      showModal: true,
+      zIndex: 210,
+    });
+
+    expect(wrapper.findByDataTest('ec-modal').attributes().style).toBe('z-index: 210;');
+    wrapper.setProps({ zIndex: 235 });
+    expect(wrapper.findByDataTest('ec-modal').attributes().style).toBe('z-index: 235;');
+    expect(wrapper.find('ec-modal').element).toMatchSnapshot();
```
mcibique: This yields no element and snapshot is undefined. You should always check what `toMatchSnapshot()` outputs to the snap file.

src/components/ec-modal/ec-modal.vue
```diff
@@ -123,6 +131,9 @@ export default {
     isLoadingNegativeButton() {
       return !!this.isLoading.negative;
     },
+    zIndexStyle() {
+      return this.zIndex ? `z-index: ${this.zIndex}` : '';
```
mcibique: We should pass `null` if zIndex is not defined, so Vue will not render the style attribute at all. Returning empty string will render empty style attribute.

mcibique: Also use objects for style attributes instead of string concatenation:

```
return this.zIndex ? { zIndex: this.zIndex } : null;
```

mcibique:
Maybe we should add a story with examples how stackable modals work. Story with three modals, two with custom zIndex, 3 boolean knobs to show all of them. Inside of each modal, add clickable link with action attached, so we see that we can interact with modal on top. WDYT?

PR #101, opened 2020-03-25
Onl 4270 currency component

stavros-tomas:
This currency component is a combination of 2 other components <ec-amount-input> and <ec-dropdown> in order to combine them we added a new prop `is-in-group` which for now accepts 'left' or 'right' value.

The currency component accepts an array of currencies and also the model is an object that looks like this { currency: 'GBP', amount: 12123.34 }.
<ec-currency-input v-model="value" :currencies="currencies">

In the future with Vue 3 we should be able to do this
<ec-currency-input v-model:currency="currency" v-model:amount="amount" />

### DIFF
--- a/src/components/ec-currency-input/ec-currency-input.spec.js
+++ b/src/components/ec-currency-input/ec-currency-input.spec.js
@@ -10,7 +10,7 @@ describe('EcCurrencyInput', () => {
     template: '<div data-popover-stub><slot /><slot name="popover" /></div>',
   });
 
-  const currencies = [{ text: 'GBP' }, { text: 'EUR' }, { text: 'USD' }];
+  const currencies = ['GBP', 'EUR', 'USD'];
 
   function mountCurrencyInput(props, mountOpts) {
     const localVue = createLocalVue();
@@ -52,12 +52,12 @@ describe('EcCurrencyInput', () => {
   });
 
   describe(':props', () => {
-    it('should render the label when the label  is given', () => {
+    it('should render the label when the label is given', () => {
       const wrapper = mountCurrencyInput({ label: 'Currency Input' });
       expect(wrapper.findByDataTest('ec-currency-input__label-text').exists()).toBe(true);
     });
 
-    it('should render the note when the note  is given', () => {
+    it('should render the note when the note is given', () => {
       const wrapper = mountCurrencyInput({ note: 'max 80 chars' });
       expect(wrapper.findByDataTest('ec-currency-input__note').exists()).toBe(true);
     });
@@ -82,15 +82,19 @@ describe('EcCurrencyInput', () => {
       const wrapper = mountCurrencyInput();
 
       selectItem(wrapper, 1);
-      // expect(wrapper.emitted('change').length).toEqual(1);
+      expect(wrapper.emitted('change').length).toEqual(1);
+      selectItem(wrapper, 2);
+      expect(wrapper.emitted('change').length).toEqual(2);
     });
 
-    it('should emit change event when amount is set', () => {
-      // const wrapper = mountCurrencyInput();
+    // it('should emit change event when amount is set', () => {
+    //   const wrapper = mountCurrencyInput();
 
-      // type data to the input
-      // expect(wrapper.emitted('change').length).toEqual(1);
-    });
+    //   // wrapper.find('.ec-currency-input__amount input').setValue('11');
+    //   selectItem(wrapper, 1);
+    //   // expect(wrapper.vm.value.amount).toEqual(11);
+    //   expect(wrapper.emitted('change').length).toEqual(1);
+    // });
   });
 
   describe('v-model', () => {
--- a/src/components/ec-currency-input/ec-currency-input.spec.js
+++ b/src/components/ec-currency-input/ec-currency-input.spec.js
@@ -66,15 +66,22 @@ describe('EcCurrencyInput', () => {
       expect(wrapper.findByDataTest('ec-dropdown').element).toMatchSnapshot();
     });
 
-    it.each('should render the amount correctly when we swap locale', async () => {
+    it('should render the amount correctly when we swap locale', async () => {
       const wrapper = mountCurrencyInput({ locale: 'es' });
       wrapper.findByDataTest('ec-currency-input__amount').setValue('1111,11');
 
       await wrapper.vm.$nextTick();
       expect(wrapper.findByDataTest('ec-currency-input__amount').element.value).toEqual('1.111,11');
 
-      wrapper.setData({ locale: 'en' });
+      wrapper.setProps({ locale: 'en' });
+      await wrapper.vm.$nextTick();
       expect(wrapper.findByDataTest('ec-currency-input__amount').element.value).toEqual('1,111.11');
+    });
+
+    it('should render a loading state when currenciesLoading is set true', () => {
+      const wrapper = mountCurrencyInput({ currenciesAreLoading: true });
+
+      expect(wrapper.findByDataTest('ec-popover-dropdown-search').element).toMatchSnapshot();
     });
   });
 
@@ -146,15 +153,13 @@ describe('EcCurrencyInput', () => {
       );
 
       expect(wrapper.findByDataTest('ec-currency-input__amount').element.value).toEqual('11,111.11');
-      selectItem(wrapper, 3);
+      selectItem(wrapper, currencies.indexOf('JPY'));
       await wrapper.vm.$nextTick();
       expect(wrapper.findByDataTest('ec-currency-input__amount').element.value).toEqual('11,111');
     });
-
-    // test JPY currency
   });
 });
 
 function selectItem(wrapper, index) {
-  wrapper.findAll('.ec-dropdown-search__item').at(index).trigger('click');
+  wrapper.findByDataTest(`ec-dropdown-search__item--${index}`).trigger('click');
 }
--- a/src/components/ec-currency-input/ec-currency-input.spec.js
+++ b/src/components/ec-currency-input/ec-currency-input.spec.js
@@ -1,0 +1,33 @@
+import Vue from 'vue';
+import { mount, createLocalVue } from '@vue/test-utils';
+import EcCurrencyInput from './ec-currency-input.vue';
+
+describe('EcCurrencyInput', () => {
+  const MockedEcPopover = Vue.extend({
+    methods: {
+      update: jest.fn(),
+    },
+    template: '<div data-popover-stub><slot /><slot name="popover" /></div>',
+  });
+
+  const currencies = [{ text: 'GBP' }, { text: 'EUR' }, { text: 'USD' }];
+
+  function mountCurrencyInput(props, mountOpts) {
+    const localVue = createLocalVue();
+
+    return mount(EcCurrencyInput, {
+      localVue,
+      propsData: { ...props },
+      stubs: { EcPopover: MockedEcPopover },
+      ...mountOpts,
+    });
+  }
+
+  it('should render properly', () => {
+    const wrapper = mountCurrencyInput({
+      currencies,
+      value: {},
+    });
+    expect(wrapper.element).toMatchSnapshot();
+  });
+});
--- a/src/components/ec-currency-input/ec-currency-input.spec.js
+++ b/src/components/ec-currency-input/ec-currency-input.spec.js
@@ -16,6 +16,28 @@ describe('EcCurrencyInput', () => {
     const localVue = createLocalVue();
 
     return mount(EcCurrencyInput, {
+      sync: false,
+      localVue,
+      propsData: {
+        currencies,
+        value: {},
+        ...props,
+      },
+      stubs: { EcPopover: MockedEcPopover },
+      ...mountOpts,
+    });
+  }
+
+  function mountCurrencyInputAsTemplate(template, props, mountOpts) {
+    const localVue = createLocalVue();
+
+    const Component = localVue.extend({
+      components: { EcCurrencyInput },
+      template,
+    });
+
+    return mount(Component, {
+      sync: false,
       localVue,
       propsData: { ...props },
       stubs: { EcPopover: MockedEcPopover },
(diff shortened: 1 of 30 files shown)


src/components/ec-input-field/ec-input-field.vue
```diff
@@ -26,6 +26,8 @@
       class="ec-input-field__input"
       :data-test="$attrs['data-test'] ? `${$attrs['data-test']} ec-input-field__input` : 'ec-input-field__input'"
       :class="{
+        'ec-input-field__input--is-in-group-right': isInGroup === 'right' ? true : false,
```
matallo: isn't this redundant? if true then true?

src/components/ec-currency-input/ec-currency-input.spec.js
```diff
@@ -24,10 +46,94 @@ describe('EcCurrencyInput', () => {
...
-      currencies,
-      value: {},
-    });
+    const wrapper = mountCurrencyInput();
+
     expect(wrapper.element).toMatchSnapshot();
   });
+
+  describe(':props', () => {
+    it('should render the label when the label  is given', () => {
```
matallo: double space

src/components/ec-currency-input/ec-currency-input.spec.js
```diff
@@ -24,10 +46,94 @@ describe('EcCurrencyInput', () => {
...
+      const wrapper = mountCurrencyInput({ label: 'Currency Input' });
+      expect(wrapper.findByDataTest('ec-currency-input__label-text').exists()).toBe(true);
+    });
+
+    it('should render the note when the note  is given', () => {
```
matallo: double space

src/components/ec-currency-input/ec-currency-input.spec.js
```diff
@@ -24,10 +46,94 @@ describe('EcCurrencyInput', () => {
...
+      const wrapper = mountCurrencyInput({ note: 'max 80 chars' });
+      expect(wrapper.findByDataTest('ec-currency-input__note').exists()).toBe(true);
+    });
+
+    it('should render the error when the errorMessage is given', () => {
+      const wrapper = mountCurrencyInput({ errorMessage: 'Random message' });
+
+      expect(wrapper.find('.ec-currency-input__currencies .ec-input-field__input--has-error').exists()).toBe(true);
```
matallo: not sure this is the best selector :thinking: 

it might be a smell that classnames could be improved, but happy to move along if anybody else disagrees

for reference, check how this is done in Rails https://guides.rubyonrails.org/v3.2/active_record_validations_callbacks.html#customizing-error-messages-css

src/components/ec-input-field/ec-input-field.vue
mcibique: well, you can write both statement into single line:
```
`ec-input-field__input--is-in-group-${isInGroup}`: !!isInGroup,
```

src/components/ec-input-field/ec-input-field.vue
```diff
@@ -96,9 +98,18 @@ export default {
       type: Number,
       default: 20,
     },
+    isInGroup: {
+      type: String,
+    },
+    idFromParent: {
```
mcibique: Why do you need this? The name of the prop indicates it's a code smell and I don't see any purpose.

src/components/ec-currency-input/ec-currency-input.vue
```diff
@@ -0,0 +1,176 @@
...
+    errorMessage: {
+      type: String,
+    },
+    currencies: {
+      type: Array,
+    },
+  },
+  data() {
+    return {
+      currenciesOnFocus: false,
```
mcibique: currenciesOnFocus -> currenciesHasFocus

src/components/ec-dropdown/ec-dropdown.vue
```diff
@@ -11,22 +11,27 @@
...
       :disabled="disabled"
+      :icon-size="24"
       data-test="ec-dropdown__input"
       readonly
       icon="simple-arrow-drop-down"
-      :icon-size="24"
+      :is-in-group="isInGroup"
       @mousedown="onMousedown"
-      @focus="onFocus"
+      @focus="onFocus, $emit('onFocus', true)"
```
mcibique: 1. emitted event should be in kebab-case
2. emitted event should not use prefix `on`.
3. don't pass true/false, use rather focus and blur events. It's not intuitive for any developer when you fire focus with false when input gets blurred. All of them are used to focus/blur events.
4. Don't put two statements into one single line. I noticed this trend in our code, it needs to stop! It makes templates very polluted, unclear and this logic should not be in templates.

src/components/ec-currency-input/ec-currency-input.vue
```diff
@@ -0,0 +1,176 @@
...
+    prop: 'value',
+    event: 'change',
+  },
+  props: {
+    value: {
+      type: Object,
+    },
+    locale: {
+      type: String,
+      default: 'en',
```
mcibique: I'm not sure we should have default value here.

src/components/ec-currency-input/ec-currency-input.vue
```diff
@@ -0,0 +1,176 @@
...
+    },
+    isInvalid() {
+      return !!this.errorMessage;
+    },
+    inputModel: {
+      get() {
+        return this.value;
+      },
+      set(value) {
+        this.$emit('change', value);
```
mcibique: I don't think so this get/set is necessary, you don't modify it anywhere.

src/components/ec-currency-input/ec-currency-input.vue
```diff
@@ -0,0 +1,176 @@
...
+      >{{ note }}</span>
+    </label>
+
+    <div
+      ref="popperWidthReference"
+      class="ec-currency-input__input-group"
+    >
+      <ec-dropdown
+        v-model="inputModel.currency"
+        :class="{'ec-currency-input__currencies--on-focus': currenciesOnFocus}"
```
mcibique: `--is-focused`

src/components/ec-currency-input/ec-currency-input.vue
```diff
@@ -0,0 +1,176 @@
...
+        :id-from-parent="id"
+        :items="currencies"
```
mcibique: I would expect passing array of currencies `['GBP', 'EUR']` via props and then this component to convert it into object that ec-dropdown can understand. Then, when you select new value in the dropdown you should update the model with currency value and not with entire item.

src/components/ec-currency-input/ec-currency-input.story.js
```diff
@@ -0,0 +1,39 @@
...
+    },
+    props: {
+      label: {
+        default: text('label', 'Text input'),
+      },
+      note: {
+        default: text('note', 'Max 80 chars'),
+      },
+      locale: {
+        default: select('Locale', ['en', 'es', 'de-ch', 'jp'], 'en'),
```
mcibique: `locale`

src/components/ec-currency-input/ec-currency-input.story.js
```diff
@@ -0,0 +1,39 @@
...
+      },
+      errorMessage: {
+        default: text('error message', ''),
+      },
+    },
+    template: `
+      <div style="max-width: 600px; margin: 60px auto">
+        <ec-currency-input v-model="value" :currencies="currencies" v-bind="$props" class="ec-mt--20 ec-mb--20" />
+      </div>
```
mcibique: Show the value of the `value` in the template. So we see it is updating properly.

src/components/ec-currency-input/ec-currency-input.vue
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  &__currencies {
+    margin-right: -1px;
+    min-width: 100px;
+    max-width: 20%;
+  }
+
+  &__currencies--on-focus {
+    z-index: $z-index-level-1;
```
mcibique: I think we need to make ec-currency-input's z-index to be 0, so this CSS will not compete with anything outside of this component. Just in case.

src/components/ec-currency-input/ec-currency-input.vue
mcibique: I think what you need here are two models:
```
currencyModel: {
  get() { return this.value.currency };
  set(item) { this.$emit('change', { ...this.value, currency: item.value }) }
},
amountModel: {
  get() { return this.value.amount };
  set(value) { this.$emit('change', { ...this.value, amount: value }) }
}
```

src/components/ec-currency-input/ec-currency-input.story.js
```diff
@@ -0,0 +1,39 @@
...
+    components: { EcCurrencyInput },
+    data() {
+      return {
+        value: {},
+        currencies: currencies.map(currency => ({ text: currency })),
+      };
+    },
+    props: {
+      label: {
+        default: text('label', 'Text input'),
```
mcibique: `'Currency input'`

src/components/ec-currency-input/ec-currency-input.story.js
```diff
@@ -0,0 +1,39 @@
...
+      },
+      note: {
+        default: text('note', 'Max 80 chars'),
```
mcibique: I would add better text here, this component doesn't really have maxlength.

src/components/ec-input-field/ec-input-field.vue
stavros-tomas: @mcibique maybe there is a better way. I want to connect the currency label with an input field (for-id)

mcibique: Okay now I understand what is this for. Rename this prop to `id`, because every prop is `FromParent`, we don't need that. And then rename the computed prop to `inputId`. 

BTW, In inputId, you can do this:
```
return this.id || `ec-input-field-${this._uid}`;
```

src/components/ec-currency-input/ec-currency-input.vue
```diff
@@ -0,0 +1,189 @@
...
+        :popper-modifiers="popperModifier"
+        :popover-options="popoverOptions"
+        class="ec-currency-input__currencies"
+        is-in-group="right"
+        is-search-enabled
+        :error-message="errorMessage"
+        @focus="currenciesHasFocus = true"
+        @blur="currenciesHasFocus = false"
+      />
+      <ec-amount-input
```
mcibique: You have to pass selected currency from currencyModel into this component too. Amount input relies of currency. If currency doesn't have decimals, it will not allow you to type decimals.

src/components/ec-currency-input/ec-currency-input.spec.js
```diff
@@ -0,0 +1,160 @@
...
+      expect(wrapper.findAllByDataTest('ec-input-field__error-text').length).toBe(0);
+    });
+
+    it('should render without any currencies if the currencies prop is empty', () => {
+      const wrapper = mountCurrencyInput({ currencies: [] });
+      expect(wrapper.findByDataTest('ec-dropdown-search__no-items').exists()).toBe(true);
+      expect(wrapper.findByDataTest('ec-dropdown').element).toMatchSnapshot();
+    });
+
+    it.each('should render the amount correctly when we swap locale', async () => {
```
mcibique: You don't need `.each`.

src/components/ec-currency-input/ec-currency-input.spec.js
```diff
@@ -0,0 +1,160 @@
...
+        },
+      );
+
+      expect(wrapper.findByDataTest('ec-currency-input__amount').element.value).toEqual('11,111.11');
+      selectItem(wrapper, 3);
+      await wrapper.vm.$nextTick();
+      expect(wrapper.findByDataTest('ec-currency-input__amount').element.value).toEqual('11,111');
+    });
+
+    // test JPY currency
```
mcibique: Remove comment.

src/components/ec-currency-input/ec-currency-input.spec.js
```diff
@@ -0,0 +1,160 @@
...
+        {},
+        {
+          data() {
+            return { currencies, value: { currency: 'GBP', amount: 11111.11 } };
+          },
+        },
+      );
+
+      expect(wrapper.findByDataTest('ec-currency-input__amount').element.value).toEqual('11,111.11');
+      selectItem(wrapper, 3);
```
mcibique: We should make clear that JPY is on 3rd index in the list:
```js
selectItem(wrapper, currencies.indexOf('JPY'));
```

src/components/ec-currency-input/ec-currency-input.spec.js
```diff
@@ -0,0 +1,160 @@
...
+      await wrapper.vm.$nextTick();
+      expect(wrapper.findByDataTest('ec-currency-input__amount').element.value).toEqual('11,111');
+    });
+
+    // test JPY currency
+  });
+});
+
+function selectItem(wrapper, index) {
+  wrapper.findAll('.ec-dropdown-search__item').at(index).trigger('click');
```
mcibique: Each item in the dropdown has data-test like this:
```
:data-test="`ec-dropdown-search__item ec-dropdown-search__item--${index}`"
```
so you can change this to:
```js
wrapper.findByDataTest(`ec-dropdown-search__item--${index}`).trigger('click');
```

src/components/ec-currency-input/ec-currency-input.vue
mcibique: We decided to make it default just like in the `ec-amount-input`.

src/components/ec-input-field/ec-input-field.spec.js
```diff
@@ -44,11 +44,23 @@ describe('EcInputField', () => {
...
+    const wrapper = mountInputField({ errorMessage: 'error msg', isInGroup: 'right' });
+    expect(wrapper.element).toMatchSnapshot();
+  });
+
+  it('renders properly with the id from the parent', () => {
+    const wrapper = mountInputField({ id: 'id-test' });
+    expect(wrapper.element).toMatchSnapshot();
+  });
+
+  // Todo errorid as well
```
mcibique: Remove comment

src/components/ec-input-field/ec-input-field.story.js
```diff
@@ -59,6 +59,9 @@ stories
       iconDate: {
         default: text('icon', '', GROUPS.DATE),
       },
+      isInGroup: {
+        default: text('is in group', ''),
```
mcibique: Maybe make this an array `['', 'left', 'right']` so it's clear for anybody what can be passed into this prop.

tests/setup/after-env.js
```diff
@@ -2,9 +2,12 @@ import { Wrapper, WrapperArray } from '@vue/test-utils';
 
 require('snapshot-diff/extend-expect');
 
+const findAllByDataTest = function findAllByDataTest(dataTest) {
```
mcibique: you can remove `const findAllByDataTest = ` bit. You declare named function, you can use it directly.

PR #104, opened 2020-04-03
ONL-4376 add the new features to the table component

jaimeperezpedra:
In this ticket we add these features:

- New event when you click to the row called 'row-click'.
- Vertical and Horizontal scroll bar when you set the height of the table or when you set the max-width of some columns and the table cannot cover all the columns with space.
- New info icon with a tooltip on the header.
- The possibility to make sticky the first or the last column of the table.
- Change the style of the table on the header capitalize and the paddings also the paddings of the cells and between the title.
- Overwrite the defaults styles of the scroll-bars.

Working with Maroua

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ebury/chameleon-components",
-  "version": "0.1.73",
+  "version": "0.1.74",
   "main": "src/main.js",
   "sideEffects": false,
   "author": "Ebury Team (http://labs.ebury.rocks/)",
--- a/src/components/ec-table-head/ec-table-head.spec.js
+++ b/src/components/ec-table-head/ec-table-head.spec.js
@@ -72,4 +72,35 @@ describe('EcTableHead', () => {
     wrapper.findByDataTest('ec-table-head__cell--0').findByDataTest('ec-table-sort__icon').trigger('click');
     expect(wrapper.emitted('sort')).toEqual([['column-a']]);
   });
+
+  it('should render info icon with a tooltip as expected', () => {
+    const wrapper = mountEcTableHead({
+      columns: columns.map(column => ({ ...column, tooltip: 'tooltip example' })),
+    });
+
+    expect(wrapper.element).toMatchSnapshot();
+  });
+
+  it('the first th should have the ec-table-head__cell--sticky-left class if stickyColumn prop is left', () => {
+    const wrapper = mountEcTableHead({
+      columns, stickyColumn: 'left',
+    });
+
+    expect(wrapper.findByDataTest('ec-table-head__cell--0').classes('ec-table-head__cell--sticky-left')).toBe(true);
+    expect(wrapper.findByDataTest('ec-table-head__cell--2').classes('ec-table-head__cell--sticky-right')).toBe(false);
+    wrapper.setProps({ stickyColumn: 'right' });
+    expect(wrapper.findByDataTest('ec-table-head__cell--2').classes('ec-table-head__cell--sticky-right')).toBe(true);
+    expect(wrapper.findByDataTest('ec-table-head__cell--0').classes('ec-table-head__cell--sticky-left')).toBe(false);
+  });
+
+  it('should render the style with the min-width on the column that have the prop given', () => {
+    const wrapper = mountEcTableHead({
+      columns: columns.map(column => ({
+        ...column,
+        minWidth: column.name === 'column-b' ? '250px' : null,
+      })),
+    });
+
+    expect(wrapper.findByDataTest('ec-table-head__cell--1').attributes('style')).toBe('min-width: 250px;');
+  });
 });
--- a/src/components/ec-table-head/ec-table-head.spec.js
+++ b/src/components/ec-table-head/ec-table-head.spec.js
@@ -1,6 +1,7 @@
 import { mount } from '@vue/test-utils';
 import * as SortDirection from '../../enums/sort-direction';
 import EcTableHead from './ec-table-head.vue';
+import { withMockedConsole } from '../../../tests/utils/console';
 
 describe('EcTableHead', () => {
   const columns = [
@@ -29,6 +30,14 @@ describe('EcTableHead', () => {
   it('should render as expected', () => {
     const wrapper = mountEcTableHead({ columns });
     expect(wrapper.element).toMatchSnapshot();
+  });
+
+  it('should throw if the prop stickyColumn have a invalid value', () => {
+    withMockedConsole((errorSpy) => {
+      mountEcTableHead({ columns, stickyColumn: 'test' });
+      expect(errorSpy).toHaveBeenCalledTimes(1);
+      expect(errorSpy.mock.calls[0][0]).toContain('Invalid prop: custom validator check failed for prop "stickyColumn"');
+    });
   });
 
   it('should have class ec-table-head__cell--text-center if type of content is icon', () => {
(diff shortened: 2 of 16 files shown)


jafcalvente:
Great work!
One thing. It sounds to me that UX wanted the tooltip to appear when hovering, both over the "i" icon and in the header text. I prefer it in this way, but we could check with them or pass them the link to see if they are ok with it.

src/scss/tools/_typography.scss
```diff
@@ -89,11 +89,18 @@
 
 @mixin input-label {
   font-size: 14px;
-  line-height: 20px;
```
jafcalvente, line 92: I think remove this line is going to change the line-height of all the current input's labels from 20px to 24px. Is that what we really want?

src/components/ec-table-head/ec-table-head.vue
```diff
@@ -42,6 +61,12 @@ export default {
       type: Array,
       default: () => [],
     },
+    stickyColumn: {
+      type: String,
+      validator(value) {
+        return ['left', 'right'].includes(value);
```
mcibique, line 67: I think this validator is not tested.

src/components/ec-table/ec-table.vue
```diff
@@ -72,6 +89,13 @@ export default {
       type: Boolean,
       default: false,
     },
+    maxHeight: Number,
+    stickyColumn: {
+      type: String,
+      validator(value) {
+        return ['left', 'right'].includes(value);
```
mcibique, line 96: I think this validator is not tested.

src/components/ec-table-head/ec-table-head.spec.js
```diff
@@ -72,4 +72,35 @@ describe('EcTableHead', () => {
...
+
+  it('should render the style with the min-width on the column that have the prop given', () => {
+    const wrapper = mountEcTableHead({
+      columns: columns.map(column => ({
+        ...column,
+        minWidth: column.name === 'column-b' ? '250px' : null,
+      })),
+    });
+
+    expect(wrapper.findByDataTest('ec-table-head__cell--1').attributes('style')).toBe('min-width: 250px;');
```
mcibique, line 104: You should also test `ec-table-head__cell--0` not having the style.

src/components/ec-table-head/ec-table-head.vue
```diff
@@ -56,21 +81,56 @@ export default {
...
+  }
+
   &__cell {
-    padding: 0 0 8px 16px;
+    padding: 16px 0 16px 16px;
     text-align: left;
-    border-bottom: 1px solid $level-6-disabled-lines;
+    box-shadow:
+      inset 0 1px 0 $level-6-disabled-lines,
+      inset 0 -1px 0 $level-6-disabled-lines;
```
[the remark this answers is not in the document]
matallo: Been here before, main difference I've found was opacity behavior, and box model block display; used both in the past indistinctly so I'm not very strong opinionated

src/components/ec-table-head/ec-table-head.vue
```diff
@@ -56,21 +81,56 @@ export default {
...
+    min-width: 100px;
+    position: sticky;
+    border-collapse: separate;
+    top: 0;
+    background: white;
```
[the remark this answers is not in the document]
matallo: better then, use `#FFF`

src/components/ec-table-sort/ec-table-sort.vue
```diff
@@ -1,34 +1,26 @@
 <template>
-  <div
+  <a
```
matallo, line 2: can this be a `button`?

src/components/ec-table/ec-table.spec.js
```diff
@@ -239,4 +239,48 @@ describe('EcTable', () => {
     wrapper.findByDataTest('ec-table-head__cell--0').findByDataTest('ec-table-sort__icon').trigger('click');
     expect(wrapper.emitted('sort')).toEqual([['lorem']]);
   });
+
+  it('the first td should have the ec-table__cell--sticky-left class if stickyColumn prop is left', () => {
```
matallo, line 243: similarly as comment above

src/components/ec-table/ec-table.vue
```diff
@@ -116,10 +148,17 @@ export default {
     }
   }
 
+  &__title {
+    @include h3;
```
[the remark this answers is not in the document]
matallo: does this change come from design?

src/scss/tools/_typography.scss
```diff
@@ -89,11 +89,18 @@
...
   font-weight: bold;
   color: $level-3-body-and-headings;
 }
 
+@mixin table-headers {
+  font-size: 14px;
+  line-height: 20px;
+  font-weight: bold;
+  text-transform: unset;
```
matallo, line 100: I know I suggested doing some transformation (from previous uppercase to Titlecase) maybe, shall we deal with this at the component level if any?

src/components/ec-table-head/ec-table-head.vue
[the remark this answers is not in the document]
jaimeperezpedra: Because the border doesn't work with the position sticky, we have to use box-shadow.

src/components/ec-table-head/ec-table-head.vue
```diff
@@ -56,21 +81,56 @@ export default {
...
+    z-index: $z-index-level-1;
+
+    @include table-headers;
 
-    @include mini-header;
+    &--sticky-left {
+      left: 0;
+      z-index: $z-index-level-2;
```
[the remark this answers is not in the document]
jaimeperezpedra: image
This is the reason why we have more levels on the right and on the left.

src/components/ec-table/ec-table.vue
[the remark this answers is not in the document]
jaimeperezpedra: Yes it cames from UX

src/components/ec-table-sort/ec-table-sort.vue
jaimeperezpedra: Why should have it as a button?

src/components/ec-table/ec-table.vue
```diff
@@ -116,10 +148,17 @@ export default {
...
+
+    padding-bottom: 16px;
+  }
+
   &__cell {
     @extend %common-cell-layout;
 
-    padding: 8px 0 8px 16px;
+    min-width: 100px;
+    padding: 16px 0 16px 16px;
```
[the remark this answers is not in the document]
jaimeperezpedra: The paddings are fine I checked with Uve, and for the footer, we need the designs for that we will work on it when we will work in the pagination ticket

src/components/ec-table-sort/ec-table-sort.vue
matallo: I like this rule of thumb from https://css-tricks.com/a-complete-guide-to-links-and-buttons/

> Are you making a JavaScript-powered clickable action? Use a button

this link doesn't point to any url, let's make it a button

jaimeperezpedra:
@jafcalvente Yes I checked with Uve and she is fine with only the tooltip on the info icon

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@ebury/chameleon-components",
-  "version": "0.1.73",
+  "version": "0.1.74",
```
jmesao, line 3: @jaimeperezpedra it is not necessary to commit this if later we'll execute "npm run release" command?

jaimeperezpedra: I do the release command before to merge the PR
I don't know if this is what you are talking about

jmesao: Ok, got it, You execute separated commands to do the release, thx.

PR #125, opened 2020-04-16
Onl 4002 align to the right and max width

jafcalvente:
I would like to add a couple of new features to the _EcTable_ component.

- To be able to align columns to the right (by default they are all aligned to the left, except the ones containing an icon that are centered).
- To be able to define a _maxWidth_ of column. If the text inside the cell requires more space than the column offers, use the _ellipsis_.

### DIFF
--- a/src/components/ec-table-head/ec-table-head.spec.js
+++ b/src/components/ec-table-head/ec-table-head.spec.js
@@ -40,7 +40,7 @@ describe('EcTableHead', () => {
     });
   });
 
-  it('should have class ec-table-head__cell--text-center if type of content is icon', () => {
+  it('should have a column align to center if its type is icon', () => {
     const wrapper = mountEcTableHead({
       columns: columns.map(column => ({
         ...column,
@@ -48,8 +48,18 @@ describe('EcTableHead', () => {
       })),
     });
 
-    expect(wrapper.findByDataTest('ec-table-head__cell--0').classes('ec-table-head__cell--text-center')).toBe(false);
-    expect(wrapper.findByDataTest('ec-table-head__cell--1').classes('ec-table-head__cell--text-center')).toBe(true);
+    expect(wrapper.element).toMatchSnapshot();
+  });
+
+  it('should have a column align to the right if its type is currency', () => {
+    const wrapper = mountEcTableHead({
+      columns: columns.map(column => ({
+        ...column,
+        type: column.name === 'column-b' ? 'currency' : null,
+      })),
+    });
+
+    expect(wrapper.element).toMatchSnapshot();
   });
 
   it('should not render if no columns are supplied', () => {
@@ -113,5 +123,16 @@ describe('EcTableHead', () => {
 
     expect(wrapper.findByDataTest('ec-table-head__cell--1').attributes('style')).toBe('min-width: 250px;');
     expect(wrapper.findByDataTest('ec-table-head__cell--0').attributes('style')).toBe(undefined);
+  });
+
+  it('should render with max-width and ellipsis on the column that has the prop given', () => {
+    const wrapper = mountEcTableHead({
+      columns: columns.map(column => ({
+        ...column,
+        maxWidth: column.name === 'column-b' ? '250px' : null,
+      })),
+    });
+
+    expect(wrapper.element).toMatchSnapshot();
   });
 });
--- a/src/components/ec-table-head/ec-table-head.vue
+++ b/src/components/ec-table-head/ec-table-head.vue
@@ -15,7 +15,10 @@
         :colspan="column.span"
         scope="col"
       >
-        <span class="ec-table-head__cell-wrapper">
+        <span
+          class="ec-table-head__cell-wrapper"
+          :class="{'ec-table-head__cell-wrapper--justify-end': column.align === 'right'}"
+        >
           <span>
             {{ column.title }}
           </span>
@@ -79,7 +82,23 @@ export default {
       this.$emit('sort', column.name);
     },
     getWidthStyle(column) {
-      return column && column.minWidth ? { minWidth: column.minWidth } : null;
+      let widthStyle = {};
+      if (column && column.minWidth) {
+        widthStyle = {
+          ...widthStyle,
+          minWidth: column.minWidth,
+        };
+      }
+      if (column && column.maxWidth) {
+        widthStyle = {
+          ...widthStyle,
+          maxWidth: column.maxWidth,
+          overflow: 'hidden',
+          textOverflow: 'ellipsis',
+          whiteSpace: 'nowrap',
+        };
+      }
+      return Object.keys(widthStyle).length ? widthStyle : null;
     },
     getStickyColumnClass(colIndex, columns) {
       if (this.stickyColumn === 'left' && colIndex === 0) {
@@ -101,6 +120,10 @@ export default {
   &__cell-wrapper {
     display: flex;
     align-items: center;
+
+    &--justify-end {
+      justify-content: flex-end;
+    }
   }
 
   &__icon {
--- a/src/components/ec-table-head/ec-table-head.vue
+++ b/src/components/ec-table-head/ec-table-head.vue
@@ -22,7 +22,7 @@
             'ec-table-head__cell-wrapper--is-type-currency': column.type === 'currency',
           }"
         >
-          <span :class="{'ec-table-head__cell-text--is-ellipsis': column.maxWidth}">
+          <span :class="{'ec-table-head__cell-text--has-max-width': column.maxWidth}">
             {{ column.title }}
           </span>
           <ec-icon
@@ -85,20 +85,10 @@ export default {
       this.$emit('sort', column.name);
     },
     getWidthStyle(column) {
-      let widthStyle = {};
-      if (column && column.minWidth) {
-        widthStyle = {
-          ...widthStyle,
-          minWidth: column.minWidth,
-        };
+      if (column && (column.maxWidth || column.minWidth)) {
+        return { maxWidth: column.maxWidth, minWidth: column.minWidth };
       }
-      if (column && column.maxWidth) {
-        widthStyle = {
-          ...widthStyle,
-          maxWidth: column.maxWidth,
-        };
-      }
-      return Object.keys(widthStyle).length ? widthStyle : null;
+      return null;
     },
     getStickyColumnClass(colIndex, columns) {
       if (this.stickyColumn === 'left' && colIndex === 0) {
@@ -165,7 +155,7 @@ export default {
     }
   }
 
-  &__cell-text--is-ellipsis {
+  &__cell-text--has-max-width {
     @include ellipsis;
   }
 }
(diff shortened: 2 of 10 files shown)


src/components/ec-table-head/ec-table-head.vue
```diff
@@ -15,7 +15,10 @@
         :colspan="column.span"
         scope="col"
       >
-        <span class="ec-table-head__cell-wrapper">
+        <span
+          class="ec-table-head__cell-wrapper"
+          :class="{'ec-table-head__cell-wrapper--justify-end': column.align === 'right'}"
```
mcibique, line 20: I don't think this is right. Table header should not align itself according to how the column content is aligned. When we have icons, they are aligned to the center but the header stays intact.

src/components/ec-table-head/ec-table-head.vue
```diff
@@ -79,7 +82,23 @@ export default {
       this.$emit('sort', column.name);
     },
     getWidthStyle(column) {
-      return column && column.minWidth ? { minWidth: column.minWidth } : null;
+      let widthStyle = {};
```
mcibique, line 85: This is too much calculation in the JS for CSS. minWidth is okay here because it cannot be set via CSS, but rest of it can. Set a modifier class if there is a maxWidth and use ellipsis mixin in CSS.

src/components/ec-table-head/ec-table-head.vue
```diff
@@ -101,6 +120,10 @@ export default {
   &__cell-wrapper {
     display: flex;
     align-items: center;
+
+    &--justify-end {
```
mcibique, line 124: we have `--text-center` and `--text-right` in cells, we should have the same naming here. But first check with Uve if this is actually desired behaviour.

src/components/ec-table/ec-table.story.js
```diff
@@ -23,6 +23,7 @@ const columns = [
     title: 'Original amount',
     sortable: true,
     tooltip: 'This is info test',
+    align: 'right',
```
mcibique, line 26: I don't like the fact that we have two different props to set one behaviour. If `type` is icon, then use `text-center`, if `align` is right, then use `text-right`. Why not using a type? Can we give a meaningful name for the type of column you are trying to built in this PR?

Joining these two props will also eliminate a clash when you set align to right and type to icon.

src/components/ec-table/ec-table.story.js
```diff
@@ -35,6 +36,11 @@ const columns = [
     type: 'icon',
     tooltip: 'This is info test',
   },
+  {
+    name: 'text-long',
+    title: 'Text long',
```
mcibique, line 41: `Long text`.

src/components/ec-table/ec-table.vue
```diff
@@ -119,7 +122,23 @@ export default {
       this.$emit('sort', columnName);
     },
     getColumnWidth(column) {
-      return column && column.minWidth ? { minWidth: column.minWidth } : null;
+      let widthStyle = {};
```
mcibique, line 125: Same as in thead. Move this to CSS.

mcibique:
BTW, why we need this alignment? Why there is a text that must be aligned to the right? What is the requirement and why is not consistent with what we already have?

src/components/ec-table-head/ec-table-head.vue
jafcalvente: I can check it with UX, because in the designs look like the header is align according to how the column content is aligned. Even in the columns which contain icons, the header text seems to be aligned to  center.
Maybe we have here a kind of misalignment between design and implementation.
image

mcibique: I see, let's wait what they say but we have to unify this, either fixing the icon type or removing right alignment.

I also see that the type of the column may be `currency`. Again, please check with designers that currencies should be always aligned to the right. Because then we should fix this on Trade finance tables too.

I think we should abandon the `--text-right`, `--text-center` and `--flex-end` modifiers and go with better names:
```css
__cell--is-type-icon {
  text-align: center;
}

__cell--is-type-currency {
  text-align: right
}
```

src/components/ec-table/ec-table.story.js
matallo: I tend to agree with Milos approach here

src/components/ec-table-head/ec-table-head.vue
```diff
@@ -15,8 +15,14 @@
         :colspan="column.span"
         scope="col"
       >
-        <span class="ec-table-head__cell-wrapper">
-          <span>
+        <span
+          class="ec-table-head__cell-wrapper"
+          :class="{
+            'ec-table-head__cell-wrapper--text-center': column.type === 'icon',
```
mcibique: change it please to { [`ec-table-head__cell-wrapper--is-type-${column.type}`]: !!column.type }

src/components/ec-table-head/ec-table-head.vue
```diff
@@ -15,8 +15,14 @@
...
+            'ec-table-head__cell-wrapper--text-right': column.type === 'currency',
+          }"
+        >
+          <span :class="{'ec-table-head__cell-text--ellipsis': column.maxWidth}">
```
mcibique: `ec-table-head__cell-text--has-max-width`

src/components/ec-table-head/ec-table-head.vue
```diff
@@ -79,7 +85,20 @@ export default {
...
+          minWidth: column.minWidth,
+        };
+      }
+      if (column && column.maxWidth) {
+        widthStyle = {
+          ...widthStyle,
+          maxWidth: column.maxWidth,
+        };
+      }
+      return Object.keys(widthStyle).length ? widthStyle : null;
```
mcibique, line 101: The entire function can be written as:

```
if (column && (column.maxWidth || column.minWidth)) {
  return { maxWidth: column.maxWidth, minWidth: column.minWidth };
}
return null;
```

src/components/ec-table/ec-table.story.js
```diff
@@ -23,6 +23,7 @@ const columns = [
     title: 'Original amount',
     sortable: true,
     tooltip: 'This is info test',
+    type: 'currency',
```
mcibique, line 26: Could you please change the data for each row to show some currencies, e.g. `EUR 100`.

src/components/ec-table-head/ec-table-head.vue
jafcalvente: Updated. Now the ellipsis mixin is used instead.

src/components/ec-table-head/ec-table-head.vue
jafcalvente: Changed to `--is-type-icon` and `--is-type-currency`

src/components/ec-table/ec-table.story.js
jafcalvente: Great. Now the type currency is used to discriminate it.

src/components/ec-table/ec-table.story.js
jafcalvente: Updated.

src/components/ec-table/ec-table.vue
jafcalvente: Done.

src/components/ec-table-head/ec-table-head.vue
jafcalvente: After talk it with UX, it is added the alignment of the content of the columns for a new type, currency, and the header text of every column is aligned in the same way as its content.
Very well spotted!

src/components/ec-table-head/ec-table-head.vue
jafcalvente: I love this way of defining the classes according to the type, I think it's much more smart and reusable, but don't you think it can give us problems when we eventually want to include the plugin to purge css with Tailwind?

src/components/ec-table/ec-table.story.js
jafcalvente: Done.

src/components/ec-table-head/ec-table-head.vue
jafcalvente: Sublime. Updated!

src/components/ec-table-head/ec-table-head.vue
jafcalvente: Renamed.

src/components/ec-table/ec-table.vue
```diff
@@ -119,7 +123,10 @@ export default {
       this.$emit('sort', columnName);
     },
     getColumnWidth(column) {
-      return column && column.minWidth ? { minWidth: column.minWidth } : null;
+      if (column && (column.maxWidth || column.minWidth)) {
+        return { maxWidth: column.maxWidth, minWidth: column.minWidth };
```
jmesao, line 127: @jafcalvente Don't we have any problem if we pass maxWidth but no minWidth being minWidth undefined?

src/components/ec-table/ec-table.vue
```diff
@@ -35,7 +35,11 @@
               class="ec-table__cell"
               :class="[
                 getStickyColumnClass(colIndex, columns),
-                { 'ec-table__cell--text-center': columns[colIndex] && columns[colIndex].type === 'icon' }]"
+                {
+                  'ec-table__cell--is-type-icon': columns[colIndex] && columns[colIndex].type === 'icon',
+                  'ec-table__cell--is-type-currency': columns[colIndex] && columns[colIndex].type === 'currency',
```
jmesao, line 40: @jafcalvente I'm thinking "currency" is too specific, maybe would it make more sense amount type? What do you think?

src/components/ec-table/ec-table.vue
jafcalvente: We should not, because when using the Object Syntax the defined styles are only applied if the condition is true.

src/components/ec-table-head/ec-table-head.spec.js
```diff
@@ -40,16 +40,26 @@ describe('EcTableHead', () => {
     });
   });
 
-  it('should have class ec-table-head__cell--text-center if type of content is icon', () => {
+  it('should have a column align to center if its type is icon', () => {
```
jmesao, line 43: @jafcalvente do you mean ... _if type is icon_ and remove _its_?

src/components/ec-table/ec-table.vue
mcibique: https://github.com/vuejs/vue/blob/aef5b4e47811cb842315bd27d6919650da45279b/src/platforms/web/server/modules/style.js#L23

The value must be string or number, otherwise it's ignored. So minWidth set to null or undefined is just not rendered.

src/components/ec-table/ec-table.vue
jafcalvente: The name of `currency` for the type was decided in a previous discussion of this same PR, and was the indicated one in the conversation that we had with UX by Slack.
A couple of reasons because I prefer to keep it:

- It is true that `currency` may seem generic (since the cell shows Symbol + Amount), but by calling it only `amount`, we are in the same generalization. If at some point we wanted to have a type that referred only to amount, we couldn't. However, if we only wanted to see the symbol we still could have the type `symbol`.

- On the other hand, the Chameleon component that allows us to enter symbol and amount is already called `Currency Input`, so for the sake of homogeneity, it would be good to keep the same pattern.

PR #132, opened 2020-04-22
ONL-4168: fix: Fixed wrong events fired from amount input and ec-amount directive

mcibique:


src/components/ec-currency-input/ec-currency-input.story.js
```diff
@@ -38,9 +39,13 @@ stories
         default: text('error message', ''),
       },
     },
+    methods: {
+      onChanged: action('change'),
```
matallo, line 43: same as above

PR #133, opened 2020-04-23
Onl 4168 fixing wrong events fired from amount input

MaruaEl:
- We fixed the issue with the focus event in the dropdown component. Added  open event and custom "after open" required for EBO
- fixed the issue with the currency column shrinking even if a set width was there. 
- renamed focus function


src/components/ec-dropdown/ec-dropdown.spec.js
```diff
@@ -403,5 +404,7 @@ describe('EcDropdown', () => {
 });
 
 function selectItem(wrapper, index) {
+  wrapper.findByDataTest('ec-dropdown__input').trigger('mousedown');
+  wrapper.findByDataTest('ec-dropdown__input').trigger('click');
```
matallo, line 408: does it make sense to trigger both events? from MDN:

> This differs from the click event in that click is fired after a full click action occurs; that is, the mouse button is pressed and released while the pointer remains inside the same element. mousedown is fired the moment the button is initially pressed.

maybe we should stick to one?

jaimeperezpedra: We need this `mousedown` first because we change on the dropdown when we fire the focus event. If you check on the dropdown method(`onFocus`) you will see that we are checking the data `shouldEmitFocus` if this is true, if not we are not firing the event.

PR #138, opened 2020-04-27
ONL-4415: feat: Added purgecss.

mcibique:


postcss.config.js
```diff
@@ -11,4 +8,28 @@ module.exports = ({ file }) => {
       'postcss-preset-env': {},
     },
   };
+
+  // Don't purge the story for TailwindCSS, otherwise it won't be able to get the full list of available utility classes.
+  const shouldPurgeFile = file.basename !== 'tailwind.story.css';
+  const isProd = process.env.NODE_ENV === 'production';
+  if (isProd && shouldPurgeFile) {
```
mgascam, line 15: Happy that finally we can exclude the tailwind story from purging

PR #150, opened 2020-05-01
ONL-4434: feat: Migrated generic folder from SASS to PostCSS

mcibique:


src/styles/generic/css-reset.css
```diff
@@ -0,0 +1,141 @@
+/* stylelint-disable */
+
+/* http://meyerweb.com/eric/tools/css/reset/
+   v2.0 | 20110126
+   License: none (public domain)
+
+   Thank you Eric :)
```
matallo, line 7: :heart:

src/styles/generic/fonts.css
```diff
@@ -0,0 +1,316 @@
+/* https://fonts.googleapis.com/css?family=Roboto+Condensed|Roboto:300,400,500,700 */
+
+/* cyrillic-ext */
+@font-face {
+  font-family: 'Roboto';
+  font-style: normal;
+  font-weight: 300;
+  src: local('Roboto Light'), local('Roboto-Light'), url('https://fonts.gstatic.com/s/roboto/v20/KFOlCnqEu92Fr1MmSU5fCRc4EsA.woff2') format('woff2');
```
matallo, line 8: This is the first time I see `local()` used, awesome! :open_mouth:

mcibique:
> Just a minor thing, do we always want to fetch the Roboto asset instead of using a (more performant) system typeface?

Hmmm, we can ask @victoriaportillo but I'm not sure if it's a good idea. We also have condensed version of the Roboto font which might not be used yet, but it can be in the future. In that case we don't have alternative in system fonts. However, our tailwind config is ready:
```fontFamily: {
      sans: [
        // 'system-ui',
        // '-apple-system',
        // 'BlinkMacSystemFont',
        // '"Segoe UI"',
        'Roboto',
        // '"Helvetica Neue"',
        // 'Arial',
        // '"Noto Sans"',
        'sans-serif',
        // '"Apple Color Emoji"',
        // '"Segoe UI Emoji"',
        // '"Segoe UI Symbol"',
        // '"Noto Color Emoji"',
      ],
    },
```
just uncomment it and it will work. Currently only Roboto and Roboto Condensed is used and sans-serif is a backup for both of them.

mcibique:
I'm going to merge this because I want to progress to the 4th layer, if we decide to change the system fonts, I think we should be doing that in a separate ticket. We will get decision from designers tomorrow.

PR #151, opened 2020-05-01
ONL-4416: feat: Added theme picker.

mcibique:


matallo:
shall we make some small edit in any component to confirm it works properly? I bet it does, though

mcibique:
It can be verified in Colors story:

https://chameleon-git-onl-4416-theme-picker-in-storybook.ebury.now.sh/?path=/story/css-colors--all

I'll add example of how to use it in component in ONL-4420.

PR #155, opened 2020-05-04
ONL-3972: tests: Added visual regression tests.

mcibique:
I might do a demo of this tomorrow after the standup, to cover some configuration options, decisions and any questions.


matallo:
> I might do a demo of this tomorrow after the standup, to cover some configuration options, decisions and any questions.

what about frontend meeting?

tests/integration/specs/visual-regression/visual-regression.spec.js
```diff
@@ -0,0 +1,132 @@
...
+      Cypress.spec.name,
+      'Visual regression tests cannot run in All Specs mode because of bugs in Cypress.\nsee https://github.com/cypress-io/cypress/issues/3090.\n',
+    ).not.to.equal('All Specs');
+
+    Cypress.log({
+      displayName: 'INIT',
+      message: 'Obtaining all stories from storybook',
+    });
+    cy.visit('/');
+    cy.get('img[alt="Chameleon components"]', { log: false }).should((img) => {
```
matallo, line 13: I find this intriguing! what's the motivation? Happy to keep it thought :smile:

mcibique: Yeah, this one is very tricky. The motivation is to wait until Storybook is fully initialized. But Storybook doesn't have `data-test` attributes and most of the classes on elements are generated from css-modules. So there is not much to use for selecting elements.

I found a "better" solution than checking the header logo has been loaded:
```
    // wait for storybook to initialize. initialization is done when left menu is loaded
    // we can check it by waiting on menu items.
    cy.get('a[id^=explore]');
```
^^ that will check the left menu items have been loaded, which means that Storybook is ready. It still bad when we compare it to approach via `data-test`, but that will do.

matallo:
we can create separate tasks for this but for OS consistency between visual regression tests we should run the tests from a container

as said, we're probably fine now

mcibique:
Moved the tests into docker using headless browsers.

mcibique:
I have created ONL-4501 for dealing with the -u flag and to investigate running `cypress open` instead of `cypress run` in the container.

PR #164, opened 2020-05-06
ONL-4443: fix: Fixing currency input again

mcibique:
It is now official: I, personally, from the bottom of my heart, hate this component. :)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@ebury/chameleon-components",
-  "version": "0.1.89",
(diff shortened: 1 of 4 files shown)


mcibique:
Yesterday I fixed two issues and created a new one. When the amount in currency input is empty, and you pick a currency, it will insert 0 into the amount input. It should stay empty. You can see that issue in master deployment: 
1. https://chameleon.ebury.now.sh/?path=/story/currency-input--basic
2. pick GBP
3. it automatically adds 0 to the input -> that's wrong

PR #165, opened 2020-05-06
ONL-4435: feat: Added new tw-grid based on the bootstrap grid system.

mcibique:


src/scss/components/ec-grid/ec-grid.story.js
```diff
@@ -1,176 +1,176 @@
-import { storiesOf } from '@storybook/vue';
```
matallo, line 1: can we get ride of the file? I'm missing why we comment the whole thing out

jaimeperezpedra: I have the same question here, for the rest looking good!

mcibique: Yeah I forgot to remove to remove this story. I removed the TF example because I think we should not have any reference to EBO and internal app logic in here.

mcibique:
@matallo all comments have been addressed.

PR #177, opened 2020-05-14
ONL-4504: feat: Added abortable fetch, empty  and error state to smart table.

mcibique:


src/components/ec-smart-table/ec-smart-table.spec.js
```diff
@@ -9,101 +9,200 @@ describe('EcSmartTable', () => {
...
         expect(wrapper.findByDataTest('ec-table-head').element).toMatchSnapshot('After sorting different column ([ASC, null, ASC])');
-        sortColumnByIndex(wrapper, 0);
+        await sortColumnByIndex(wrapper, 0);
         expect(wrapper.findByDataTest('ec-table-head').element).toMatchSnapshot('After sorting different column #2 ([DESC, null, ASC])');
       });
     });
   });
 });
+
+function flushPromises() {
```
jmesao, line 200: I saw in Vue-test-utils a similar thing but using an npm package flush-promises, but that great do that here directly :)

src/components/ec-smart-table/ec-smart-table.spec.js
```diff
@@ -9,99 +10,188 @@ describe('EcSmartTable', () => {
...
 
-  function sortColumnByIndex(wrapper, index) {
-    wrapper.findByDataTest(`ec-table-head__cell--${index}`).findByDataTest('ec-table-sort__icon').trigger('click');
+  async function mountEcSmartTableWithResolvedData(resolvedData, props, mountOpts) {
+    const resolvedDataSource = {
+      fetch: jest.fn().mockResolvedValue(resolvedData),
+    };
+    const wrapper = mountEcSmartTable({ ...props, dataSource: resolvedDataSource }, mountOpts);
+    await flushPromises();
+    wrapper.vm.$forceUpdate();
```
jafcalvente, line 43: Out of curiosity. Why do we need to execute `$forceUpdate`? Waiting for resolution of promises was not enough in this case then?

src/main.js
```diff
@@ -25,9 +25,15 @@ export { default as EcNavigationLink } from './components/ec-navigation-link';
...
 export { default as EcPopover } from './components/ec-popover';
 export { default as EcPrivacyPolicy } from './components/ec-privacy-policy';
+export { default as EcSmartTable } from './components/ec-smart-table';
 export { default as EcSubmenu } from './components/ec-submenu';
 export { default as EcTable } from './components/ec-table';
 export { default as EcToaster } from './components/ec-toaster';
 export { default as EcTooltip } from './directives/ec-tooltip';
 export { default as EcUserInfo } from './components/ec-user-info';
+
+export { default as ecWithAbortableFetch } from './hocs/ec-with-abortable-fetch';
```
jafcalvente, line 35: So must we use camelCase for HOCs instead of PascalCase?

src/hocs/ec-with-loading/ec-with-loading.js
```diff
@@ -0,0 +1,35 @@
...
+  name: 'EcWithLoading',
+  functional: true,
+  props: {
+    isLoading: {
+      type: Boolean,
+      default: false,
+    },
+    isLoadingTransparent: {
+      type: Boolean,
+      default: true,
```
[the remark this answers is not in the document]
mcibique: That's true. However, if you don't set a default value for Boolean, it will become `false`. So it will not trigger the default value in EcLoading.

I was thinking to remove any constrains from HOC props and just let it pass as it is:
```
isLoading: {},
isLoadingTransparent: {},
loadingIconSize: {},
```

That would move the logic back to the EcLoading code and no code duplication is needed.

src/components/ec-smart-table/ec-smart-table.spec.js
mcibique: I couldn't figure out 100% why is this happening, but it has something to do with combination of functional and non-functional components when rendering EcSmartTable and turning the sync mode off when mounting components:
```
  withEcSmartTableContainer(
    withAbortableFetch(
      withSorting( // functional
        withEcSmartTableRenderer( // functional, including withLoading
          EcTable,
```
Without this, the table ends up always in initial loading state regardless of how many flushes/nextTicks I do.

src/main.js
mcibique: It is a function that wraps a Component and produces HOC so it makes sense to name it in camelCase IMO. The function itself is not a component, but a "decorator" or "enhancer" or a "component factory".

PR #205, opened 2020-05-27
Onl 4554 migrate checkbox

jafcalvente:


CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+# 0.1.98 - Minor changes in tailwind.config.js
```
jafcalvente, line 1: @mcibique when you get a chance I'd like to know if you see any drawbacks to these changes I made to the tailwind config.

src/components/ec-checkbox/ec-checkbox.vue
```diff
@@ -111,102 +117,86 @@ export default {
...
   &__check-icon-wrapper {
-    cursor: pointer;
-    background: $ec-checkbox-default-background-color;
-    width: 20px;
-    height: 20px;
-    flex-shrink: 0;
-    border: 2px solid $ec-checkbox-default-border-color;
-    position: relative;
-    margin-right: 8px;
+    background: white;
```
jaimeperezpedra, line 137: This could be `@apply tw-bg-light`

mcibique, line 137: `tw-bg-light`

src/components/ec-checkbox/ec-checkbox.vue
```diff
@@ -111,102 +117,86 @@ export default {
...
+      @apply tw-border-error;
     }
   }
 
   &__check-icon {
-    fill: $ec-checkbox-default-background-color;
-    position: absolute;
-    left: 50%;
-    top: 50%;
+    fill: white;
```
mcibique, line 191: `tw-fill-light`

src/components/ec-checkbox/ec-checkbox.vue
```diff
@@ -111,102 +117,86 @@ export default {
...
     transform: translate(-50%, -50%);
```
mcibique, line 192: I think this could be `tw--translate-x-1/2 tw--translate-y-1/2`. Please test it first.

src/components/ec-checkbox/ec-checkbox.story.js
```diff
@@ -5,26 +5,26 @@ import EcCheckbox from './ec-checkbox.vue';
 const stories = storiesOf('Checkbox', module);
 
 stories
-  .add('basic', () => ({
+  .add('all', () => ({
     components: { EcCheckbox },
     props: {
       valueFromPropsChecked1: {
-        default: boolean('Checkbox 1'),
+        default: boolean('"Label and Error messages coming from props" Checkbox'),
```
mcibique, line 12: I found this labels very confusing, you can make it better by using knob groups: https://www.npmjs.com/package/@storybook/addon-knobs#categorization.

src/components/ec-checkbox/ec-checkbox.story.js
```diff
@@ -45,73 +45,21 @@ stories
...
-    components: { EcCheckbox },
-    data() {
-      return {
         errorMessage: 'An error has occurred',
         errorMessageMultiline: 'Lorem ipsum dolor sit amet consectetur adipisicing elit. Provident eos consequatur quas reiciendis aliquid ipsam ea pariatur dolorem, molestias maiores.',
       };
     },
     template: `
-      <div class="ec-m--24" style="max-width:500px;">
+      <div class="tw-m-24" style="max-width:500px;">
```
mcibique, line 53: Could you try `tw-max-w-screen-sm`? It's not exactly the same and it might affect the generated screenshots. Just give a try.

src/components/ec-checkbox/ec-checkbox.story.js
```diff
@@ -137,28 +85,75 @@ stories
...
+          :error-message="valueFromPropsErrorMessage">
+        </ec-checkbox>
+
+        <h3> Label and Error messages coming from template</h3>
+        <ec-checkbox
+          v-if="!valueFromPropsHasError"
+          v-model="checkbox2"
+          :disabled="valueFromPropsDisabled">
+          <template #label>
+            I accept the <a href="#" @click.stop.prevent="$emit('OpenTerms')"> terms and conditions </a>
```
mcibique, line 134: It seems like this slips through the previous reviews: it should follow our conventions: `open-terms`.

src/components/ec-checkbox/ec-checkbox.vue
jafcalvente: To be honest, I'm not sure this should be so, since this color, before the migration, was defined in the variable `$ec-checkbox-default-background-color`, which had as value `$white`, and the white color should always be `white`, I think

jafcalvente: I answered Jaime this about this topic

> To be honest, I'm not sure this should be so, since this color, before the migration, was defined in the variable $ec-checkbox-default-background-color, which had as value $white, and the white color should always be white, I think

So, we need to change `white` to `tw-bg-light` for any reason? I have doubts about this.

src/components/ec-checkbox/ec-checkbox.vue
jafcalvente: Nop, I tried it :-(
We can use both `tw--translate-x-1/2` and `tw--translate-y-1/2` separately, but if we want to use them together, we'd need the utility class `tw-transform`, that is deactivated in `corePlugins`, because it is not compatible with IE11: https://tailwindcss.com/docs/translate

src/components/ec-checkbox/ec-checkbox.vue
jmesao, line 191: @jafcalvente I think we could do @apply tw-fill-light;

src/components/ec-checkbox/ec-checkbox.vue
mcibique: Nope, white color should not be treated as a special color. We have branding now, every color should be changeable. I wanted to get rid of the naming of "$white", so it can be tweaked to any color we want. I couldn't find better name than `light` for `$white` and `dark` for `$black`.
If you have:
```
.my-component {
  background-color: var(--some-color); // blue for example
  color: white;
}
```
and then in branding I'll change `--some-color` to `#fff`, you will get white text on white background.

src/components/ec-checkbox/ec-checkbox.vue
mcibique: cool, leave it as it is then

src/components/ec-checkbox/ec-checkbox.vue
mcibique: I think better names for them would be `key-7` and `key-8`. Because there are nothing but regular colors we use.

jafcalvente: Understood. Thank you for the explanation, it make totally sense.

mcibique: @online I have a suggestion here: I'd completely drop $white/$black and `ec-color-light/ec-color-dark` and go for these two colors instead:
```
--ec-gray-color-level-7: #F4F5F6
--ec-gray-color-level-1: #1C2022
```
because:
1. F4F5F6 - This is very close to white, we use it for the background of the body.
2. 1C2022 - We don't use black directly anywhere, designers don't use black color at all (it's too dark). We need black only to be a base color for shadows and backdrops (e.g `rgba(0,0,0, 0.5)`) so it doesn't matter if the base color is little bit lighter than black.

This basically means we would have no "color" meaning in the variable names, e.g. imagine if we want to create night mode of EBO, I still have to do this:
```
:root {
  --ec-color-light: black; // or something very dark
  --ec-color-dark: white; // or something very light
}
```
After dropping these we would do:
```
:root {
  --ec-gray-color-7: black; // or something very dark
  --ec-gray-color-1: white; // or something very light
}
```
Which is absolutely fine.

WDYT?

matallo: Makes sense to me, let's add these to the gray variants :+1:

matallo:
I remember discussing we should have the stories for visual regression testing in columns for better snapshots performance, however they are in a single column now. Is this quicker?

mcibique:
The viewport before snapshot is set to 1280x720px (it's 1280x694px on FF for unknown reason). If story doesn't fit into it, Cypress takes snapshot of what's visible, then scrolls the story by 720px, takes another snapshot and at the end it will join the generated snapshots together. This takes some time, it's not very significant on small stories, but on very large (e.g. button's 1280x30000px) it can take 40s-70s.

src/components/ec-checkbox/ec-checkbox.story.js
jafcalvente: Done

src/components/ec-checkbox/ec-checkbox.vue
mcibique: ooof, that's too much change. I think we have to go for `gray-8`, `gray-9/0` then.

src/components/ec-checkbox/ec-checkbox.story.js
jafcalvente: I tested and yes, we get a slight difference between both generated screenshots, since the `tw-max-w-screen-sm` give us 76px more, and it makes that the two examples of multiline label test appear on 2 lines instead of on 3 lines. In my POV, the objective of these two tests is fulfilled anyway, since we are seeing them with a line break, so I would see right to make use of this utility class, changing the visual regression test
image

jafcalvente: Finally I've included the `tw-max-w-screen-sm` class, and replaced the Lorem ipsum text by other longer, so we can maintain a very similar snapshot (with 3-line text).

src/components/ec-checkbox/ec-checkbox.story.js
jafcalvente: Done

src/components/ec-checkbox/ec-checkbox.vue
jafcalvente: Done

jafcalvente: Done.

src/components/ec-checkbox/ec-checkbox.vue
jafcalvente: Done

PR #216, opened 2020-06-01
ONL-4559 migrate loading component

jaimeperezpedra:
Migrate Loading component.

New classes:
tw-rotate-360


mcibique:
There are also hard-coded styles in the story which should be changed to tw-* classes, e.g.
```
 <div style="height: 100vh">
```

PR #217, opened 2020-06-01
ONL-4576 - Using tw classes in main container component

MaruaEl:
Sub task: https://fxsolutions.atlassian.net/browse/ONL-4576
- Migrating the ec-main-component to use Tailwind utility classes
- added data-test attribute


src/components/ec-main-container/ec-main-container.vue
```diff
@@ -1,8 +1,16 @@
 <template>
-  <div class="ec-main-container">
+  <div class="tw-p-24 tw-min-h-full ec-main-container">
```
matallo: taking into account we keep the `. ec-main-container` class, wouldn't it be interesting to move these utility classes with `@apply` ?

PR #221, opened 2020-06-04
ONL-4585 feat: Allow the sorting order of a column to be configurable

mgascam:
**TL,DR;**

- Adds `defaultSortCycle` property to `EcWithSorting`. This property controls the default sort cycle of all the columns, if none is applied to specific column. It is an array like: `[null, 'asc','desc']` (default)

- Column object definition can override the default for any specific column, providing a `sortCycle` property

This PR adds the possibility to configure the cycle in which the sorts are applied to the table when a user clicks on the table column. To represent this cycle, I am using an array. The first element of the array represents the state where no sort is applied, and the elements 1 and 2 represents the states after the user clicks. For instance, the current sorting cycle goes like this: 'none', 'ascending', 'descending'. It will be represented as `[null, 'asc', 'desc']`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ebury/chameleon-components",
-  "version": "0.1.100",
+  "version": "0.1.101",
   "main": "src/main.js",
   "sideEffects": false,
   "author": "Ebury Team (http://labs.ebury.rocks/)",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ebury/chameleon-components",
-  "version": "0.1.101",
+  "version": "0.1.102",
   "main": "src/main.js",
   "sideEffects": false,
   "author": "Ebury Team (http://labs.ebury.rocks/)",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ebury/chameleon-components",
-  "version": "0.1.101",
+  "version": "0.1.100",
   "main": "src/main.js",
   "sideEffects": false,
   "author": "Ebury Team (http://labs.ebury.rocks/)",
--- a/src/components/ec-smart-table/__snapshots__/ec-smart-table.spec.js.snap
+++ b/src/components/ec-smart-table/__snapshots__/ec-smart-table.spec.js.snap
@@ -103,7 +103,9 @@ exports[`EcSmartTable should render resolved data properly 1`] = `
     class="ec-loading__content ec-loading__content--is-transparent"
     data-test="ec-loading__content"
   >
-    <div>
+    <div
+      defaultsortcycle=",asc,desc"
+    >
       <!---->
        
       <div
--- a/src/components/ec-smart-table/__snapshots__/ec-smart-table.spec.js.snap
+++ b/src/components/ec-smart-table/__snapshots__/ec-smart-table.spec.js.snap
@@ -103,9 +103,7 @@ exports[`EcSmartTable should render resolved data properly 1`] = `
     class="ec-loading__content ec-loading__content--is-transparent"
     data-test="ec-loading__content"
   >
-    <div
-      defaultsortcycle=",asc,desc"
-    >
+    <div>
       <!---->
        
       <div
--- a/src/components/ec-smart-table/ec-smart-table.story.js
+++ b/src/components/ec-smart-table/ec-smart-table.story.js
@@ -20,6 +20,7 @@ const columns = [
     name: 'original-amount',
     title: 'Original amount',
     sortable: true,
+    sortCycle: [null, 'desc', 'asc'],
   },
   {
     name: 'repayment-date',
@@ -99,6 +100,12 @@ stories
       fetchEmptyList: {
         default: boolean('fetchEmptyList', false),
       },
+      defaultSortCycle: {
+        default: select('Sort Cycle by default', {
+          ascFirst: [null, 'asc', 'desc'],
+          descFirst: [null, 'desc', 'asc'],
+        }),
+      },
     },
     methods: {
       onSort: action('sort'),
@@ -162,6 +169,7 @@ stories
             :sticky-column="stickyColumn || null"
             :error-message="errorMessage || undefined"
             :empty-message="emptyMessage || undefined"
+            :default-sort-cycle="defaultSortCycle"
             @sort="onSort"
             @abort="onAborted"
             @error="onError">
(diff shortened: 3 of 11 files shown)


src/hocs/ec-with-sorting/ec-with-sorting.spec.js
```diff
@@ -1,7 +1,7 @@
 import { mount, createLocalVue, createWrapper } from '@vue/test-utils';
 import * as SortDirection from '../../enums/sort-direction';
 import withSorting from './ec-with-sorting';
-
+// TODO test defaultSort
```
mgascam, line 4: This is a left over, it is already tested

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@ebury/chameleon-components",
-  "version": "0.1.100",
+  "version": "0.1.101",
```
jaimeperezpedra, line 3: Recommendation:
I usually bump the version after get the approvals and just before merge my PR with the last version on master, if you bump the version right now can happen that other people merge another PR update the version and you will have conflicts with the versions

src/components/ec-smart-table/__snapshots__/ec-smart-table.spec.js.snap
```diff
@@ -103,7 +103,9 @@ exports[`EcSmartTable should render resolved data properly 1`] = `
     class="ec-loading__content ec-loading__content--is-transparent"
     data-test="ec-loading__content"
   >
-    <div>
+    <div
+      defaultsortcycle=",asc,desc"
```
mcibique, line 107: This is not a valid HTML attribute, you have to stop passing the defaultSortCycle from withSorting to the wrapped component. You can do that in the `createHOCc` call in withSorting:
```
// this bit determines what properties are being passed to the wrapped component. By default, all props coming to the withSorting are passed down to the wrapped component. On top of that, you can add more props, e.g. sorts will be passed like this:
props: {
  sorts() {
      return this.internalSorts;
    },
}
```
^^ to fix this you have to do this. change the props to a function:
```
props(props) {
      return { 
        ...props, // pass everything you've got
        sorts: this.internalSorts, // pass additional prop
        defaultSortCycle: null, // don't pass defaultSortCycle to the wrapped component
      };
    },
```

src/hocs/ec-with-sorting/ec-with-sorting.js
```diff
@@ -13,40 +13,41 @@ const withSorting = createHOCc({
       type: Array,
       default: () => [],
     },
+    defaultSortCycle: {
```
mcibique, line 16: I'd remove the word `default` from the name. We do not expect any other SortCycle to be passed in the future.

src/hocs/ec-with-sorting/ec-with-sorting.js
```diff
@@ -13,40 +13,41 @@ const withSorting = createHOCc({
...
+      type: Array,
+      default: () => [null, SortDirection.ASC, SortDirection.DESC],
```
mcibique, line 18: I think null should be always first and I wouldn't make its position configurable. I cannot imagine passing something like
`[SortDirection.ASC, null, SortDirection.DESC]`.

src/hocs/ec-with-sorting/ec-with-sorting.js
```diff
@@ -64,8 +65,9 @@ const withSorting = createHOCc({
     },
   },
   listeners: {
-    sort(columnName) {
-      this.sortBy(columnName);
+    sort(column) {
+      const { name: columnName, sortCycle } = column;
+      this.sortBy(columnName, sortCycle);
```
mcibique, line 70: I'm thinking about creating and exporting an enum sort-direction-cycle.js:
```
export const LOWEST_FIRST = [ASC, DESC]
export const HIGHEST_FIRST= [DESC, ASC]
```

so then we can do in EBO:
```
columns: [{
  name: 'currency',
  title: 'Currency column',
  sortable: true,
  sortCycle: SortDirectionCycle.HIGHEST_FIRST,
}]
```
Or we can declare then in EBO only but more specific:
```
export const NUMBERS = [DESC, ASC]
```
Or we can use both approaches:
```
export const NUMBERS = HIGHEST_FIRST // HIGHEST_FIRST comes from chameleon, NUMBERS is declared in EBO.
```

You can also reuse the enum for default of value of sortCycle prop:
```
props: {
  sortCycle: {
    default = () => SortDirectionCycle.LOWEST_FIRST
}
}

File: src/components/ec-smart-table/ec-smart-table.story.js [hunk 7]
```diff
@@ -99,6 +100,12 @@ stories
       fetchEmptyList: {
         default: boolean('fetchEmptyList', false),
       },
+      defaultSortCycle: {
+        default: select('Sort Cycle by default', {
```
mcibique, line 104: Use name of the prop as a label just like the other props do.

src/components/ec-smart-table/__snapshots__/ec-smart-table.spec.js.snap
mgascam: Will do thanks

package.json
mgascam: Thanks for the advice, I will follow your suggestion

src/hocs/ec-with-sorting/ec-with-sorting.js
mgascam: Valid point

- [ ] Rename

src/hocs/ec-with-sorting/ec-with-sorting.js
mgascam: Yep makes sense

- [ ] remove 'null' position

src/hocs/ec-with-sorting/ec-with-sorting.js
mgascam: I really like the idea thanks

- [ ] Declare enum

src/components/ec-smart-table/ec-smart-table.story.js
mgascam: - [ ] Rename

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@ebury/chameleon-components",
-  "version": "0.1.101",
+  "version": "0.1.100",
```
matallo, line 3: careful, current version is 101

mgascam: Yep, thanks for the heads up

PR #225, opened 2020-06-05
main container fix css

jaimeperezpedra:
**BEFORE**
image
**AFTER**
image


mcibique:
I think this solution is quite complicated, why have you not just wrapped the content slot with this?
```
<div class="tw-col-full">
 <slot />
</div>
```

PR #226, opened 2020-06-05
ONL-4402 add tooltip to input filed component label

MaruaEl:
- This pr will add a tooltip to the input filed's label

image


src/components/ec-input-field/ec-input-field.story.js
```diff
@@ -136,6 +136,10 @@ stories
           <div class="ec-col-4 ec-p--24">
             <ec-input-field readonly placeholder="My input" value="Lorem ipsum dolor sit amet consectetur adipisicing elit. Provident eos consequatur quas reiciendis aliquid ipsam ea pariatur dolorem, molestias maiores." label="Read only input with long text" :icon="iconText" :is-in-group="isInGroup" @change="onChange" @input="onInput" />
           </div>
+
+          <div class="ec-col-3 ec-p--24">
+            <ec-input-field placeholder="An other input" v-model="valueText" label="Input tooltip on the label" :is-in-group="isInGroup" tooltipLabel="I'm a tooltip text" @change="onChange" @input="onInput" />
```
mcibique, line 141: `placeholder="Another input"`
tooltipLabel should be `tooltip-label`

src/components/ec-input-field/ec-input-field.vue
```diff
@@ -11,9 +11,19 @@
     >
       <span
         v-if="label"
-        class="ec-input-field__label-text"
+        :class="{'ec-input-field__label-text': true, 'ec-input-field__label-text--tooltip': tooltipLabel}"
```
mcibique, line 14: We should use `--is-`, `--can-`, '--has-' prefixes for modifiers, e.g. `--has-tooltip`.

src/components/ec-input-field/ec-input-field.vue
```diff
@@ -82,6 +94,10 @@ export default {
       default: '',
       type: String,
     },
+    tooltipLabel: {
```
mcibique, line 97: We should write a test for the new prop. We don't have to test the tooltip, but we should test the icon is there if it's set.

src/components/ec-input-field/ec-input-field.story.js
mcibique, line 141: I'd add possibility to set that value from story props too.

src/components/ec-input-field/ec-input-field.vue
```diff
@@ -205,11 +221,19 @@ $ec-input-field-invalid-color: $color-error !default;
...
+  }
+
   &__label-text {
     @include input-label;
 
     flex-grow: 1;
     margin-right: 8px;
+
+    &--tooltip {
+      flex-grow: 0;
```
mcibique, line 235: I think that enforcing this might cause an issue when the input has label-text, note and label-tooltip. Can you please create that combination in the story and test it?

The `flex-grow: 1` is there to push the note to the right, or push it to the new line when label and note are to long.

MaruaEl: image

image

image

This are I think all the combinations possibly... not sure if it ever happen to have a note with a tooltip?

mcibique: It usually contains validation instructions, so it could be possible. What would happen if we move the tooltip icon inside of `ec-input-field__label-text`?
```
  <span class="ec-input-field__label-text">Short label <ec-icon ... /></span>
```
and keep label-text growing.

MaruaEl: image 
Was one of my first solutions but then I went mad with the alignment of the icon, I couldn't find a way to align it correctly to the text

mcibique: You can add to the icon:
```
display: inline-block;
vertical-align: baseline? bottom? text-top? // one of those
```

or you can make `ec-input-field__label-text` display: flex or inline-flex and position children inside using flex properties.

MaruaEl: ohh, ok in the end `ec-input-field__label-text display: flex` works with the icon `align-self: center;`
I'll update the tests and push again - thanks

src/components/ec-input-field/ec-input-field.spec.js
```diff
@@ -155,4 +156,9 @@ describe('EcInputField', () => {
     const wrapper = mountInputField({ icon: 'simple-check' }, { attrs: { disabled: true } });
     expect(wrapper.element).toMatchSnapshot();
   });
+
+  it('renders properly when the given prop is set', () => {
+    const wrapper = mountInputField({ labelTooltip: 'Testing the labelTooltip prop' });
+    expect(wrapper.props().labelTooltip).toBe('Testing the labelTooltip prop');
```
mcibique, line 162: This test doesn't check the logic in the template (`v-if="labelTooltip"`) it tests that Vue can set props. That prop might not be used properly in the template or in computed props or other methods. We should check what was rendered in the wrapper by checking:
```
```
We want to see that the icon is not there in all other tests except for this specific one.

src/components/ec-input-field/ec-input-field.spec.js
```diff
@@ -155,4 +156,9 @@ describe('EcInputField', () => {
     const wrapper = mountInputField({ icon: 'simple-check' }, { attrs: { disabled: true } });
     expect(wrapper.element).toMatchSnapshot();
   });
+
+  it('renders properly when the given prop is set', () => {
```
mcibique, line 160: renders properly when the labelTooltip prop is set

PR #234, opened 2020-06-12
Onl 4613 migrate privacy policy

jmesao:


src/components/ec-privacy-policy/ec-privacy-policy.vue
```diff
@@ -31,49 +32,41 @@ export default {
...
     }
   }
 
   &__title {
-    @include h4;
+    @apply tw-h4;
 
-    @media screen and (max-width: $ec-privacy-policy-max-width) {
-      display: none;
+    @media screen and (max-width: var(--ec-body-max-width)) {
```
mcibique, line 61: I'm pretty sure you cannot use `var()` in the media query and I see that the design is broken because of that. There is a solution for this: `env()` but it is not supported by IE11, doesn't have polyfill, only postcss plugin which requires these values to be resolved during the build. We need to investigate on Monday what to do here. Hardcoding the values is our plan C. Using `env()` without supporting the IE11 is the plan B.

https://caniuse.com/#feat=css-env-function
https://developer.mozilla.org/en-US/docs/Web/CSS/env

PR #235, opened 2020-06-12
ONL-4614 migrate toaster

jaimeperezpedra:


src/components/ec-toaster/ec-toaster.vue
```diff
@@ -53,63 +53,58 @@ export default {
...
+    @apply tw-flex tw-flex-col tw-items-end;
+    @apply tw-relative;
+    @apply tw-overflow-visible;
+    @apply tw-p-24;
+    @apply tw-m-0;
+    @apply tw-pointer-events-none;
   }
 
   &__item {
     display: inline-block;
```
matallo, line 78: 🔥

src/components/ec-toaster/ec-toaster.vue
```diff
@@ -53,63 +55,56 @@ export default {
...
+    @apply tw-shadow-level-2;
   }
 }
 
 .ec-toaster-items-transition-enter,
 .ec-toaster-items-transition-leave-to {
-  opacity: 0;
-  transform: translateX(100%);
+  @apply tw-opacity-0;
+  @apply tw--translate-x-full;
```
mcibique, line 104: `tw-translate-x-full`

src/components/ec-toaster/ec-toaster.story.js
```diff
@@ -73,15 +73,13 @@ stories.add('basic', () => ({
...
-        right: 0,
-        top: 0,
-        width: '100%',
+      <div class="tw-fixed tw-right-0 tw-top-0 tw-w-full" style={{
         maxWidth: '400px',
       }}>
         <EcToaster messages={this.messages} onRemove={message => this.removeMessage(message)} />
-        <button style="position:fixed;top:50%;left:50%;transform: translate(-50%, -50%);" class="ec-btn ec-btn--md ec-btn--rounded ec-btn--primary" onClick={this.addMessage}>Add another message</button>
+        <button
+          class="ec-btn ec-btn--md ec-btn--rounded ec-btn--primary tw-fixed tw-left-1/2 tw-top-1/2 tw--translate-x-1/2 tw--translate-y-1/2"
```
mcibique, line 81: `tw-translate-x-1/2 tw--translate-y-1/2` doesn't work the same way. This generates two transform properties, one of them will be ignored then, because element can have only one. Either leave the transform in CSS form (no TW) or refactor the story. I don't understand why the button is positioned this way? Why is it not outside of the fixed div and positioned to the center using flex for instance. That would be much better solution.

PR #251, opened 2020-06-19
Onl 4568 colors gray scale chameleon

MaruaEl:
We are going to eliminate, as colors of the Tailwind theme, both dark and light, and they are going to be part of the gray scale, and become gray-0 and gray-8 respectively.
This change implies the refactoring of its use, both in Chameleon component and in EBO.


stavros-tomas:
Good job Maroua :) 

Just a small point which I am not sure also, seems like classes 

".tw-bg-dark" ".tw-bg-gray-0"
".tw-border-dark" ".tw-border-gray-0" etc..

Do the same thing, shall we keep it like that?

matallo:
Well spotted @stavros-tomas let's remove the dark and white classes 👍

PR #252, opened 2020-06-19
ONL-4584: feat: Added button category for both modal buttons.

mcibique:


src/components/ec-modal/ec-modal.vue
```diff
@@ -57,31 +57,37 @@
             :show="isLoadingNegativeButton"
             :transparent="!isLoadingNegativeButton"
           >
-            <button
+            <ec-btn
```
matallo, line 60: niiice ✨

PR #273, opened 2020-06-25
Onl 4669 inline actions migration

jaimeperezpedra:
- Migrate the classes from css to tw
- Add the possibility to add href and download attrs to the items

The regression tests are green!


src/components/ec-inline-actions/ec-inline-actions.vue
```diff
@@ -16,14 +16,17 @@
             :key="item.text"
             v-ec-tooltip.left="item.tooltip"
             class="ec-inline-actions__item"
-            :data-test="`ec-inline-actions__item ec-inline-actions__item-${index}${indexList}`"
+            :data-test="`ec-inline-actions__item ec-inline-actions__item--${index}-${indexList}`"
```
jaimeperezpedra, line 19: I added '-' to a better understanding of the data test

src/components/ec-inline-actions/ec-inline-actions.vue
```diff
@@ -74,61 +77,78 @@ export default {
...
+    @apply tw-border-none;
+    @apply tw-btn-text;
+    @apply tw-cursor-pointer;
+    @apply tw-py-8 tw-px-16;
+    @apply tw-flex tw-items-center;
 
     &:hover {
-      background: $level-7-backgrounds;
+      @apply tw-bg-gray-7;
+      @apply tw-no-underline;
```
jaimeperezpedra, line 127: This is needed for the anchor underline

src/components/ec-inline-actions/ec-inline-actions.vue
```diff
@@ -74,61 +77,78 @@ export default {
...
     },
   },
+  methods: {
+    componentTag(item) {
+      if (item.href) {
+        return 'a';
+      }
+      return 'button';
+    },
+    getDownloadText(item) {
```
mcibique, line 87: I'm not sure here. Shouldn't the logic be:
1. if there is a href and download, return download
2. it there is a href and download is an empty string, return empty string <- MDN says download attribute doesn't have to have always a value
3. if there is a href and download is undefined or null, return null
4. if there is no href, return null

src/components/ec-inline-actions/ec-inline-actions.vue
```diff
@@ -74,61 +77,78 @@ export default {
...
+      const hasHrefAndDowload = !!item.href && !!item.download;
+      if (hasHrefAndDowload) {
+        return item.download;
+      }
+      if (item.href) {
+        return '';
```
mcibique, line 93: This will make every anchor downloadable. You cannot make just plain anchor without download attribute. I'd do this only if download is set to an empty string, otherwise null.

Please write a test for it.

src/components/ec-inline-actions/ec-inline-actions.vue
```diff
@@ -16,14 +16,17 @@
...
           >
-            <button
+            <component
+              :is="componentTag(item)"
               v-ec-close-popover="!item.disabled"
               class="ec-inline-actions__button"
               data-test="ec-inline-actions__button"
               :class="item.disabled ? 'ec-inline-actions__button--disabled' : ''"
               :disabled="item.disabled"
+              :href="!!item.href ? item.href : null"
```
mcibique, line 28: disabled doesn't work on `<a>`.

src/components/ec-inline-actions/ec-inline-actions.vue
```diff
@@ -74,61 +77,78 @@ export default {
...
-    cursor: pointer;
-    padding: 8px 16px;
-    display: flex;
-    align-items: center;
+    @apply tw-text-left;
+    @apply tw-w-full;
+    @apply tw-bg-gray-8;
+    @apply tw-text-gray-3;
+    @apply tw-border-none;
+    @apply tw-btn-text;
```
mcibique, line 120: :+1: Like it. Make sure the line-height: 22px and font-weight: 400 doesn't break the design.

src/components/ec-inline-actions/ec-inline-actions.vue
```diff
@@ -74,61 +77,78 @@ export default {
...
-    width: 100%;
-    background: $white;
-    color: $level-3-body-and-headings;
-    border: none;
-    font-size: 16px;
-    cursor: pointer;
-    padding: 8px 16px;
-    display: flex;
-    align-items: center;
+    @apply tw-text-left;
```
mcibique, line 115: You can join tw-text-left tw-text-gray-3 tw-btn-text;

src/components/ec-inline-actions/ec-inline-actions.vue
jaimeperezpedra: I'm also adding the case that the item it's disabled, in that case, we will return null.

And the 3 and the 4 for me it's the same if you have the download undefined or null you will return null, doesn't matter the href.

src/components/ec-inline-actions/ec-inline-actions.vue
jaimeperezpedra: It doesn't break it!

src/components/ec-inline-actions/ec-inline-actions.story.js
```diff
@@ -43,7 +46,7 @@ stories.add('basic', () => ({
   template: `
   <div style="display: flex; height: 100vh">
     <div style="margin: auto">
```
stavros-tomas, line 48: 2 style tags here and line above :)

PR #275, opened 2020-06-29
ONL-4667 migration to tw

MaruaEl:
- you'll notice `class="tw-my-64 tw-mx-auto" style="max-width: 600px;"` margin 64 instead of 60px let me know if you prefer to keep it 60. I've also left the custom mac-width.


src/components/ec-currency-input/ec-currency-input.vue
```diff
@@ -186,70 +186,61 @@ export default {
...
 
   &__input-group {
-    display: flex;
-    flex-direction: row;
+    @apply tw-flex tw-flex-row;
   }
 
   &__currencies {
-    $currencies-width: 104px;
+    --currencies-width: 104px;
```
mcibique, line 198: local CSS variables are not supported by polyfills: https://www.npmjs.com/package/css-vars-ponyfill (Limitations):
> Custom property declaration support is limited to :root and :host rulesets

You can create this var globally: `--ec-currency-input-currencies-width: 104px`

src/components/ec-currency-input/ec-currency-input.vue
```diff
@@ -186,70 +186,61 @@ export default {
...
-<style lang="scss">
-@import '../../scss/tools/index';
-@import '../../scss/settings/index';
-
-$ec-currency-input-invalid-color: $color-error !default;
-
+<style>
 .ec-currency-input {
-  z-index: 0; // When in currencies dropdown is focused gets a high z-index. We set the index here of the whole component to 0 so we eliminate the chance of competing with other components
+  z-index: 0;
```
mcibique, line 191: Can we keep that comment in the code?

src/components/ec-currency-input/ec-currency-input.story.js
```diff
@@ -43,8 +43,8 @@ stories
       onChange: action('change'),
     },
     template: `
-      <div style="max-width: 600px; margin: 60px auto">
-        <ec-currency-input v-model="value" :currencies="currencies" v-bind="$props" @change="onChange" class="ec-mt--20 ec-mb--20" />
+      <div class="tw-my-64 tw-mx-auto" style="max-width: 600px;">
```
mcibique, line 46: You can also replace the max-width: 600px with: `tw-max-w-screen-sm`. It's not exactly the same but we can get rid of the style. 

But I wonder that visual regression tests didn't fail after this change.

src/components/ec-currency-input/ec-currency-input.vue
MaruaEl: Thank you Milos, I've created that variable in layout.css

src/components/ec-currency-input/ec-currency-input.story.js
MaruaEl: will change it and update regression tests

src/components/ec-currency-input/ec-currency-input.vue
```diff
@@ -186,70 +186,59 @@ export default {
...
 
   &__input-group {
-    display: flex;
-    flex-direction: row;
+    @apply tw-flex tw-flex-row;
   }
 
   &__currencies {
-    $currencies-width: 104px;
+    width: var(--currencies-width);
```
mcibique, line 198: Don't forget to update references of `--currencies-width` to `--ec-currency-input-currencies-width`.

stavros-tomas:
It seems we have a visual bug. [Previously]( https://chameleon-git-onl-4270-currency-component.ebury.now.sh/?path=/story/currency-input--basic) when selecting a currency the currency dropdown was remaining highlighted (all borders blue). Now the right hand side border is gray. 

@mcibique shall we make a quick fix in this PR like:
`    onCurrencyChange(evt) {
      this.currenciesHasFocus = true;
      this.$emit('change', evt);
      this.$emit('currency-change', evt);
    },`

MaruaEl:
image
@stavros-tomas do you mean this?

stavros-tomas:
Yes that's it. In previous version of this when selecting a currency/focusing becomes like this
image

mcibique:
Not sure if `onCurrencyChange` function is the right place to fix it. If you change the value of the input via JS, e.g. `this.currencyInputModel = { 'currency': 'EUR' }`, that might trigger the change event too. Try it before you fix it.

PR #294, opened 2020-07-13
Onl 4683 visual fixes

mcibique:
Small fixed required for new TF Gov schemes:
1. ec-main-container without intro have a huge bottom margin
2. fixed submenu not being tabbable (a11y). 
3. fixed submenu jumping on small screens when changing the tab.

### DIFF
--- a/src/components/ec-main-container/ec-main-container.spec.js
+++ b/src/components/ec-main-container/ec-main-container.spec.js
@@ -31,7 +31,7 @@ describe('EcMainContainer', () => {
     expect(wrapper.element).toMatchSnapshot();
   });
 
-  it('should display only the title when the titleIntro is not given', () => {
+  it('should render without title intro when the titleIntro prop is not given', () => {
     const propsData = {
       title: 'Trade Finance',
     };
--- a/src/components/ec-main-container/ec-main-container.spec.js
+++ b/src/components/ec-main-container/ec-main-container.spec.js
@@ -31,6 +31,17 @@ describe('EcMainContainer', () => {
     expect(wrapper.element).toMatchSnapshot();
   });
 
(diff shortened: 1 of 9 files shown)


src/components/ec-main-container/ec-main-container.spec.js
```diff
@@ -31,6 +31,17 @@ describe('EcMainContainer', () => {
     expect(wrapper.element).toMatchSnapshot();
   });
 
+  it('should display only the title when the titleIntro is not given', () => {
```
matallo, line 34: I like to write the specs on top of each other, thus I'd phrase this as "it renders properly" then a "it renders with titleIntro"

src/components/ec-submenu/ec-submenu.vue
```diff
@@ -2,26 +2,31 @@
...
     v-if="submenu && submenu.length > 0"
     class="ec-submenu"
+    :data-test="$attrs['data-test'] ? `${$attrs['data-test']} ec-submenu` : 'ec-submenu'"
   >
     <div class="ec-submenu__header-container">
       <ul class="ec-submenu__header">
         <li
           v-for="(menuItem, index) in submenu"
           :key="index"
-          :data-test="'ec-submenu__header-item-' + index"
```
jaimeperezpedra, line 11: I guess this remove of the index it's because we want to check better the title instead of the item right?

mcibique: I moved the `@click` events from li to anchor because of a11y and I wanted to move the data-test too, because usually we want to target in tests an element which has a click handler. I still left a data-test here, but it's not that important for us.

jaimeperezpedra: Cool! thx for the explanation!

PR #296, opened 2020-07-20
ONL-4673 ec-popover migration

MaruaEl:
https://fxsolutions.atlassian.net/browse/ONL-4673

### DIFF
--- a/src/styles/components/v-tooltip.css
+++ b/src/styles/components/v-tooltip.css
@@ -1,3 +1,6 @@
+
+@import '../../styles/tools/ec-z-level-all.css';
+
 .ec-popover {
   @apply tw-block;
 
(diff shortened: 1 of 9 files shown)


src/styles/components/v-tooltip.css
```diff
@@ -0,0 +1,33 @@
+.ec-popover {
+  @apply tw-block;
+
+  /* base z-indexing for popover and related modifiers */
+  @apply tw-z-level-1;
```
mcibique, line 5: This is not the same as `@include z-index-all-levels` in scss. That mixin generates this:
```
@mixin z-index-all-levels {
  @include z-index-level-1;

  &--level-2 {
    @include z-index-level-2;
  }

  &--level-3 {
    @include z-index-level-3;
  }

  &--modal {
    @include z-index-modal;
  }

  &--tooltip {
    @include z-index-tooltip;
  }

  &--notification {
    @include z-index-notification;
  }
}
```

MaruaEl: Thank you Milos for spotting it! I've added the migrated mixin as well and added it to the file

PR #297, opened 2020-07-21
ONL-4676 menu migration

jaimeperezpedra:
**Layout Menu migration.**

There is some unit test that we can fix it(because they are using classes) once we do the migration of the [navigation-link](https://fxsolutions.atlassian.net/browse/ONL-4678).


src/components/ec-menu/ec-menu.vue
```diff
@@ -50,34 +53,29 @@ export default {
...
-  margin: 0;
-  display: flex;
-  flex-direction: column;
-  flex-wrap: wrap;
-  align-items: flex-start;
-  justify-content: flex-start;
+  @apply tw-list-none;
+  @apply tw-p-0;
+  @apply tw-m-0;
+  @apply tw-flex tw-flex-row tw-flex-wrap;
```
matallo, line 61: wasn't it column? 🤔

jaimeperezpedra: done

src/components/ec-menu/ec-menu.vue
```diff
@@ -50,34 +53,29 @@ export default {
...
+  @apply tw-m-0;
+  @apply tw-flex tw-flex-column tw-flex-wrap;
+  @apply tw-items-start tw-justify-start;
 
   &--horizontal {
-    flex-direction: row;
-    justify-content: center;
-    align-items: center;
+    @apply tw-flex-row;
+    @apply tw-items-start tw-justify-start;
```
mcibique, line 66: I think both should be `*-center`

jaimeperezpedra: done

src/components/ec-menu/ec-menu.story.js
```diff
@@ -39,10 +39,10 @@ stories
       },
     },
     template: `
-      <div style="margin: 0 auto;" :style="{ maxWidth: width + 'px' }">
+      <div class="tw-my-auto tw-mx-auto" :style="{ maxWidth: width + 'px' }">
```
jmesao, line 42: my is 0 no auto, continue having the same behavior?

jaimeperezpedra: Yes it has the same behaviour, anyway I'll fix it

PR #301, opened 2020-07-22
ONL-4675 -> migrated tooltip and re-styled the story using tw classes

MaruaEl:
ONL-4675 -> [https://fxsolutions.atlassian.net/browse/ONL-4675](https://fxsolutions.atlassian.net/browse/ONL-4675)

- migrated tooltip styles to use tw
- re-styled story to use tw classes - is a bit different from before but I thought that is still make sense
- please note that I'm still running the regression tests on my local in case deployment fails.


matallo:
needs work

Screenshot 2020-07-22 at 10 53 49

vs

Screenshot 2020-07-22 at 10 53 42

MaruaEl:
Hi Carlos, 
thank you - I'll fix it. My bad, I thought that it didn't need to be exactly the same but to just showcase the functionality. 
Will update the pr asap

src/styles/components/v-tooltip.css
```diff
@@ -1,6 +1,111 @@
...
+      @apply tw-bg-gray-8;
+      @apply tw-text-gray-3;
+      @apply tw-border tw-border-solid tw-border-gray-6;
+    }
+  }
+
+  &__arrow {
+    @apply tw-z-tooltip;
+    @apply tw-w-0;
+    @apply tw-h-0;
```
jmesao, line 29: @MaruaEl You could group tw-h-0 and put it in the same line with tw-w-0

src/styles/components/v-tooltip.css
```diff
@@ -1,6 +1,111 @@
...
+    @apply tw-border-solid;
+    @apply tw-absolute;
+    @apply tw-border-gray-2;
```
jmesao, line 32: @MaruaEl You could group this and put it in the same line with tw-border-solid

src/styles/components/v-tooltip.css
MaruaEl: Thanks will do! v

matallo:
now we're talking

postcss.config.js
```diff
@@ -28,6 +28,10 @@ module.exports = ({ file }) => {
         /data-v-.*/,
         /^ec-/,
       ],
+      whitelistPatternsChildren: [
+        /^ec-/,
+        /^ebo-/,
```
mcibique, line 33: We have to do this in ebo repo. remove the ebo- bit from here. chameleon has no clue about EBO.

PR #302, opened 2020-07-23
Onl 4674 migrate table component tw

mgascam:
Table component migration to TW. There are some rules that I wasn't sure how to proceed, so I didn't apply the migration:
- `position: sticky` is not available as a utility class, should I add it?
- `min-width: 100px` it does not seem to align to the grid, Should I add `tw-min-w-104`?
-`border-collapse` is not available

I had some issues with the screenshots missing most of the text. I tried to `waitOn` but it didn't work. It seemed to me like the browser didn't have enough time to render the fonts. So I added a new property called `waitForDOM` to be able to increase the waiting for this story without affecting others. Please let me know if there is a better way to handle this


src/components/ec-table/ec-table.vue
```diff
@@ -137,81 +137,81 @@ export default {
...
 
-<style lang="scss">
-@import '../../scss/settings/colors/index';
-@import '../../scss/tools/index';
+<style>
+@import '../../styles/tools/scrollbars';
+
+.common-cell-layout {
+  &:last-of-type {
+    padding-right: 16px;
```
jaimeperezpedra, line 145: Here you can use  tw-pr-16

src/components/ec-table/ec-table.vue
```diff
@@ -137,81 +137,81 @@ export default {
...
+  @apply tw-pt-16;
+  @apply tw-w-full;
+  @apply tw-relative;
+
   border-spacing: 0;
-  border: none;
-  padding-top: 16px;
-  width: 100%;
-  position: relative; // We need to reset the z-index so the sticky columns and header will not compete with the outer world
-  z-index: 0;
```
jaimeperezpedra, line 156: Any reason to remove the z-index?

src/components/ec-table-footer/ec-table-footer.vue
```diff
@@ -23,25 +23,23 @@ export default {
...
 
   &__cell {
-    padding: 0 16px;
-    width: 100%;
+    @apply tw-py-0 tw-px-16;
+    @apply tw-w-full;
+    @apply tw-bottom-0;
+    @apply tw-bg-gray-7;
+
     border-radius: 0 0 5px 5px;
```
jaimeperezpedra, line 37: Here you can use this class tw-rounded-b

src/components/ec-table-head/ec-table-head.vue
```diff
@@ -102,61 +102,61 @@ export default {
 };
 </script>
 
-<style lang="scss">
-@import '../../scss/tools/typography';
-@import '../../scss/settings/index';
+<style>
+/* @import '../../scss/tools/typography'; */
```
jaimeperezpedra, line 106: remove the commented code

src/components/ec-table-sort/ec-table-sort.vue
```diff
@@ -70,33 +70,31 @@ export default {
...
+    @apply tw-text-key-3;
+    @apply tw-outline-none;
   }
 
   &__icon {
-    fill: currentColor;
+    @apply tw-fill-current;
 
     &--asc {
       transform: rotateX(180deg);
```
jaimeperezpedra, line 92: I guess we can add some class for theses cases

jaimeperezpedra:
I think you can add the position: sticky
I show more times the min-width: 100px, maybe will be nice to add a new number to the classes

src/components/ec-table-sort/ec-table-sort.vue
```diff
@@ -70,33 +70,31 @@ export default {
 };
 </script>
 
-<style lang="scss">
-@import '../../scss/settings/colors/index';
-@import '../../scss/tools/accessibility';
+<style>
 
 .ec-table-sort {
-  line-height: 0;
```
matallo, line 78: have we gotten rid of this?

src/components/ec-table/ec-table.story.js
```diff
@@ -73,8 +73,11 @@ const data = [
 const stories = storiesOf('Table', module);
 
 stories
```
matallo, line 75: I think the new way of navigating tables is an improvement, kudos 👏

src/components/ec-table/ec-table.story.js
```diff
@@ -95,116 +98,55 @@ stories
...
     methods: {
       onSort: action('sort'),
       onRowClick: action('onRowClick'),
     },
     template: `
-      <div style="display: flex; height: 100vh">
-        <div style="margin: auto 20px; width: 90vw" class="ec-card" >
+      <div class="tw-flex tw-flex-col tw-h-screen">
+        <h2 class="feature-title tw-m-24">Basic</h2>
+        <div class="tw-my-auto tw-mx-20 ec-card" style="width: 90vw">
```
matallo, line 111: any reason why keeping the width and not making it 100%? Can't remember why

src/components/ec-table/ec-table.vue
```diff
@@ -137,81 +137,81 @@ export default {
 };
 </script>
 
-<style lang="scss">
-@import '../../scss/settings/colors/index';
-@import '../../scss/tools/index';
+<style>
+@import '../../styles/tools/scrollbars';
```
matallo, line 141: nice one, just importing what we're actually using 👍

src/components/ec-table/ec-table.vue
matallo: it's just below, isn't it? (Happened to me too)

src/components/ec-table/ec-table.vue
```diff
@@ -137,81 +137,81 @@ export default {
...
+  @apply tw-relative;
+
   border-spacing: 0;
-  border: none;
-  padding-top: 16px;
-  width: 100%;
-  position: relative; // We need to reset the z-index so the sticky columns and header will not compete with the outer world
-  z-index: 0;
-
-  %common-cell-layout {
```
matallo, line 158: But why do we remove this?

MaruaEl:
Hi Miguel, 
can you help me understand the difference between this branch basic version: 
image

and the previous? 
image


I'm referring to the gray bg in the first column - thanks

matallo:
@MaruaEl you can check that with the knobs

src/components/ec-table/ec-table.vue
jaimeperezpedra: Then use the tw-z-0 class

src/components/ec-table-sort/ec-table-sort.vue
mgascam: Glups, fixed it

src/components/ec-table/ec-table.story.js
mgascam: I am not sure...I tried with 100% and the tables overflow the viewport (at least in my laptop)

src/components/ec-table/ec-table.vue
mgascam: I think we don't have that class available. I checked and we are using the same css rule in src/components/ec-currency-input/ec-currency-input.vue that is already migrated.

I am happy to add the `tw-z-level-0` class if we agree.

src/components/ec-table/ec-table.story.js
jaimeperezpedra: This because if you set to 100% it will break the table on the story

src/components/ec-table-sort/ec-table-sort.vue
mgascam: Yep, I checked and we only have this instance so far. Maybe in the future?

src/components/ec-table/ec-table.vue
mgascam: I moved it to the top of the style block. The compiler was complaining that I could not define it here

src/components/ec-table/ec-table.vue
mgascam: Done thanks!

src/components/ec-table/ec-table.vue
mgascam: Done thanks!

src/components/ec-table-footer/ec-table-footer.vue
mgascam: Done thanks!

src/components/ec-table-head/ec-table-head.vue
mgascam: Done thanks!

src/components/ec-table/ec-table.vue
```diff
@@ -137,81 +137,80 @@ export default {
...
-    padding: 16px 0 16px 16px;
-    border-bottom: 1px solid $level-6-disabled-lines;
-    vertical-align: middle;
-
-    @include body-text;
+    @apply .common-cell-layout;
+    @apply tw-pt-16 tw-pr-0 tw-pb-16 tw-pl-16;
+    @apply tw-border-b tw-border-solid tw-border-gray-6;
+    @apply tw-align-middle;
+    @apply tw-body-text;
```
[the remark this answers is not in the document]
mgascam: Yes it is there please double check:
@include body-text;

mgascam:
> A couple of things:
> 
> * The basic table had the fixed column (the first one) with gray background color, but now it appears without it)
>   image
> * The tables into the firefox snapshot are not being displayed properly.

Thanks for your feedback, please find below my comments:

#1 I changed the way we are showcasing the sticky column functionality. In the previous version, the tables in the stories had the sticky column enabled by default, and you can play with the knobs to enable/disable. Now the only table that has the sticky enabled by default is the one that is showing that scenario (the one at the bottom). The rest have the sticky disabled by default, and you can enable with the knobs. Please let me know if this solution works for you.

#2 I had some issues with screenshots. The thing is that the stories do not have enough time to render the font. I tried an approach of adding a longer delay but it was discarded in favor of a better solution: install the Roboto font in the container. This is going to be done in a different task, once all the migrations are ready

jafcalvente:
Yes, both behaviors were already commented on. I didn't remember them. Sorry for wasting your time :-(

PR #303, opened 2020-07-23
ONL-4678 navigation link migration

jaimeperezpedra:


src/components/ec-navigation-link/ec-navigation-link.vue
```diff
@@ -102,74 +108,68 @@ export default {
...
 
     &:hover,
     &:focus {
-      color: $ec-navigation-link-text-color;
+      @apply tw-text-gray-8;
     }
 
     &:focus {
-      background-color: rgba($level-4-tech-blue, 0.9);
+      background: hsla(var(--ec-key-color-level-4), 0.9);
```
jafcalvente, line 172: In the SASS version, the CSS property used was `background-color`, but now the shorthand `background` is the used one. Did you have to do this change because of something?

jaimeperezpedra: Nop, it's a mistake, already fixed

matallo:
Transitions don't look exactly right, to be honest previous ones seemed too slow for me, can you review with @victoriaportillo ?

PR #305, opened 2020-07-24
ONL 4677 - migrate layout navigation

stavros-tomas:
Migrate navigation component to use Tailwind


tailwind.config.js
```diff
@@ -388,6 +388,7 @@ module.exports = {
     zIndex: {
       0: 0,
       auto: 'auto',
+      navigation: 1,
```
stavros-tomas, line 391: @mcibique not sure if we can remove this. You remember why we added it in the navigation?

stavros-tomas: Talked with Milos this is resolved.

src/components/ec-navigation/ec-navigation.vue
```diff
@@ -80,6 +80,12 @@ export default {
 </script>
 
 <style>
+:root {
```
MaruaEl, line 83: nice

PR #306, opened 2020-07-24
ONL-4679 migrated layout user info

MaruaEl:
https://fxsolutions.atlassian.net/browse/ONL-4679

image


src/styles/settings/layout.css
```diff
@@ -5,4 +5,5 @@
   --ec-modal-lg-max-width: 1100px;
   --ec-currency-input-currencies-width: 104px;
   --ec-tooltip-width-value: 6px;
+  --ec-user-info-avatar-size: 48px;
```
mcibique, line 8: I don't think this is a global variable that anybody can re-use. Just leave it in ec-user-info.vue style. you can declare
```
:root {
  --ec-user-info-avatar-size: 48px;
}
```

there. I'm also not sure why the other variables ended it this file, like ec-modal, ec-currency and ec-tooltip.

mcibique: You can even use TW for this:

:root {
  --ec-user-info-avatar-size: theme('spacing.48');
}

just like in https://github.com/Ebury/chameleon/blob/master/src/components/ec-dropdown-search/ec-dropdown-search.vue#L310

MaruaEl:
I'm updating visual regression tests - then I'll merge - thanks

PR #314, opened 2020-07-31
ONL-4790 fix: prevent double click on panel ctas

mgascam:
Opening this PR for discussion of a fix for [ONL-4790](https://fxsolutions.atlassian.net/browse/ONL-4790). This bug can be reproduced by double clicking on the back CTA of the panel. Using the Vue devtools you can see that the event is triggered twice. The first one is handled correctly, but not the second one. 
Screenshot_20200731_091718

I propose to avoid the double click by using the `once` directive. About the root cause, I am not 100% sure. I belive it has to do with the fact that the second event is not handled. Using the debugger tools, I can see  the line that throws when the second event is emitted:
Screenshot_20200731_092027

### DIFF
--- a/src/components/ec-panel/ec-panel.vue
+++ b/src/components/ec-panel/ec-panel.vue
@@ -18,7 +18,7 @@
           aria-label="Go back"
           class="ec-panel__header-action ec-panel__header-action--back"
           data-test="ec-panel__header-action--back"
-          @click.stop.prevent="goBack"
+          @click.once.stop.prevent="goBack"
         >
           <ec-icon
             class="ec-panel__header-icon"
@@ -32,7 +32,7 @@
           aria-label="Close panel"
           class="ec-panel__header-action ec-panel__header-action--close"
           data-test="ec-panel__header-action--close"
-          @click.stop.prevent="closePanel"
+          @click.once.stop.prevent="closePanel"
         >
           <ec-icon
             class="ec-panel__header-icon"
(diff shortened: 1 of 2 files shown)


src/components/ec-panel/ec-panel.vue
```diff
@@ -32,7 +32,7 @@
           aria-label="Close panel"
           class="ec-panel__header-action ec-panel__header-action--close"
           data-test="ec-panel__header-action--close"
-          @click.stop.prevent="closePanel"
+          @click.once.stop.prevent="closePanel"
```
mcibique, line 35: I think we need to address this problem differently. I see two issues with fixing it like this:
1. `closePanel` emits `close` event and it's up to the parent to decide if the panel should get closed or not. If parent will decide to not closing the panel, then it stays open and it will the close event will not be fired in the future.
2. Our code should be robust enough to handle multiple callback calls. It should not rely on the fact that the close event will be emitted 0..1 times. We should definitely check what is the problem in EBO's code and only if we cannot fix that, we will re-consider this approach.

I still don't understand what caused the error in EBO.

matallo: I think like Milos, we should dig deeper into it, I'd prefer not to ship this fix.

mgascam: Yep, agree, this is more a workaround than a proper fix. I don't really understand the problem. Anyway I thought about sharing the PR to help understand the problem.

PR #325, opened 2020-08-07
feat(ONL-3957): Add keyboard navigation for dropdown search component

jafcalvente:
The `EcDropdownSearch` (and by extension `EcDropdown`) component has been provided with keyboard navigation. This navigation is based on the native `select` tag.

Its main features are:
- The dropdown can gain focus, like any other input, button or anchor, by pressing the `tab` key. Once it gets the focus, another press of the `tab` key will make the focus go to the next element.

- If the dropdown is **not** in multiple selection mode, we can select an item using the `up` and `down` arrow keys without opening the dropdown.

- The dropdown is not automatically opened after gaining focus. This requires pressing the `enter` or `space` key.

- Once opened, we will move between the elements with the `up` and `down` arrow keys.

- As we navigate through the options, the scroll will be updated, so at all times the element we have active will be visible.

- If there was an element already selected before opening the dropdown, when opened the scroll will be updated to make the element visible.

- If the dropdown is **not** in multiple selection mode, navigating between the items will cause the item to be selected automatically.

- If the dropdown is open and is **not** in multiple selection mode, we can close it using the `enter` or `space` keys.

- If the dropdown is in multiple selection mode, navigating between the elements will make them appear in the same style as if we were doing `mouseover`, but it will not select them. We can select them by using the `enter` or `space` keys.

- Keyboard events are of the `keydown` type (not `keypress`, which is deprecated, nor `keyup`).


src/components/ec-dropdown-search/ec-dropdown-search.vue
```diff
@@ -297,6 +347,112 @@ export default {
...
+    updateScroll() {
+      const currentItemIndex = this.filteredItems.indexOf(this.currentActiveItem);
+      const $elItems = this.$refs.itemElements;
+
+      if ($elItems && $elItems.length && currentItemIndex >= 0) {
+        const itemTop = $elItems[currentItemIndex].offsetTop;
+        const itemBottom = itemTop + $elItems[currentItemIndex].offsetHeight;
+        const visibleWindowHeight = this.$refs.itemsOverflowContainer.clientHeight;
+
+        if (!this.visibleWindow) {
```
mcibique, line 435: I'm not sure that remembering the position of scrolled "window" is the right thing to do. I found multiple issues when this visibleWindow values go out of sync because I scrolled the dropdown and then I closed it without selecting a value. I think the internal state only complicates this. 

You need 4 variables:
```
let itemTopEdge = $elItems[currentItemIndex].offsetTop;
let itemBottomEdge = itemTopEdge + $elItems[currentItemIndex].offsetTop;

let containerTopEdge = this.$refs.itemsOverflowContainer.scrollTop;
let containerBottomEdge = containerTopEdge  + this.$refs.itemsOverflowContainer.clientHeight;

if (itemTopEdge < containerTopEdge) {
  this.$refs.itemsOverflowContainer.scrollTop = itemTopEdge 
} else if (itemBottomEdge  > containerBottomEdge) {
  this.$refs.itemsOverflowContainer.scrollTop = itemBottomEdge
} else {
  // do nothing, both edges of the item are visible
}
```
^^ something like this. I haven't run that code!

BTW, maybe we should also check that container's clientHeight is lower than container's scrollHeight and do the maths only if so. Otherwise there is no scrollbar needed and all items are visible.

src/components/ec-dropdown-search/ec-dropdown-search.vue
```diff
@@ -186,11 +199,18 @@ export default {
...
+      type: Boolean,
+      default: false,
+    },
   },
   data() {
     return {
       isOpen: false,
       filterText: '',
+      visibleWindow: null,
+      active: null,
```
mcibique, line 212: I'd name it lastHighlightedItem and set it in `select()` method regardless whether it is multiple or single. That means you don't need currentActiveItem computed property

src/components/ec-dropdown-search/ec-dropdown-search.vue
```diff
@@ -186,11 +199,18 @@ export default {
       type: String,
       default: 'No results found',
     },
+    isMultiple: {
```
mcibique, line 202: I think this is something we should discuss in the standup. Basically this component is not aware of that there is multiple options available, the whole logic is in ec-dropdown.vue. I understand that you need to know here what mode the parent is in but that's the problem. Child should not be aware of the parent. Three options how to fix this:
1. Move multiple logic into this component
2. Remove the support of multiple options completely, follow YAGNI principle. The multiple options is speced but not used anywhere in our code. We have to maintain it and I see that it complicates lots of logic in this ticket. I implemented it because it was 5 minute task while implementing ec-dropdown. We can remove that functionality and rework it when needed. 
3. We can scope out the multiple support from this ticket and do it separately with option 1.

^^ @matallo

src/components/ec-dropdown-search/ec-dropdown-search.vue
```diff
@@ -262,15 +287,34 @@ export default {
     isEmpty() {
       return this.isSearchEnabled && this.filteredItems.length === 0;
     },
+    isFirstItemCustom() {
```
mcibique, line 290: Maybe name it `isFirstItemSelectable`?

src/components/ec-dropdown-search/ec-dropdown-search.vue
```diff
@@ -297,6 +347,112 @@ export default {
...
+          this.visibleWindow.top = 0;
+          this.visibleWindow.bottom = visibleWindowHeight;
+        } else if (itemTop <= this.visibleWindow.top) {
+          this.visibleWindow.top = itemTop;
+          this.visibleWindow.bottom = this.visibleWindow.top + visibleWindowHeight;
+        }
+        this.$refs.itemsOverflowContainer.scrollTop = this.visibleWindow.top;
+      }
+    },
+    removeActiveItem() {
```
mcibique, line 453: `clearLastHighligthedItem` would be a better name. Remove indicates it is going to be removed from the list and user cannot select it anymore.

tests/utils/html-element.js
```diff
@@ -0,0 +1,13 @@
+export function setUpHTMLElement(options) {
```
mcibique, line 1: I'm not sure this will be useful for anything else than testing the scroll functionality in dropdown-search, I'd keep it in the spec.js file then, maybe rename it to `mockHtmlElementPosition`. I'm not sure how the teardown of this mock will look like.

src/components/ec-dropdown-search/ec-dropdown-search.vue
```diff
@@ -297,6 +347,112 @@ export default {
     hasCta() {
       return !!this.$scopedSlots.cta;
     },
+    onArrowKey(key) {
+      const currentItemIndex = this.filteredItems.indexOf(this.currentActiveItem);
+      let nextItem;
+
+      if (currentItemIndex >= 0) {
+        if (key === ARROW_DOWN) {
+          const selectableItems = this.filteredItems.slice(currentItemIndex + 1).filter(item => !item.disabled);
```
mcibique, line 356:
```
const nextItem = this.filteredItems.find((item, i) => !item.disabled && i > currentItemIndex); 
// ^^ does not iterate over array twice and breaks the loop when first item matches. Returns null if no item matches.

...
if (nextItem) {
  this.activateItemViaKeyboardNavigation(nextItem);
}
```

src/components/ec-dropdown-search/ec-dropdown-search.vue
```diff
@@ -297,6 +347,112 @@ export default {
     hasCta() {
       return !!this.$scopedSlots.cta;
     },
+    onArrowKey(key) {
+      const currentItemIndex = this.filteredItems.indexOf(this.currentActiveItem);
```
mcibique, line 351: `const currentHighligthedItemIndex = ...`

src/components/ec-dropdown-search/ec-dropdown-search.vue
```diff
@@ -297,6 +347,112 @@ export default {
...
+      let nextItem;
+
+      if (currentItemIndex >= 0) {
+        if (key === ARROW_DOWN) {
+          const selectableItems = this.filteredItems.slice(currentItemIndex + 1).filter(item => !item.disabled);
+          if (selectableItems.length) {
+            [nextItem] = selectableItems;
+          }
+        } else {
+          const selectableItems = this.filteredItems.slice(0, currentItemIndex).filter(item => !item.disabled);
```
mcibique, line 361: Same thick as above:
```
const nextItem = this.filteredItems.find((item, i) => !item.disabled && i < currentItemIndex); 
```

src/components/ec-dropdown-search/ec-dropdown-search.vue
```diff
@@ -297,6 +347,112 @@ export default {
...
+          if (selectableItems.length) {
+            nextItem = selectableItems[selectableItems.length - 1];
+          }
+        }
+      } else {
+        const selectableItems = this.filteredItems.filter(item => !item.disabled);
```
mcibique, line 367: Use `.find()` instead of `.filter()`

src/components/ec-dropdown-search/ec-dropdown-search.vue
```diff
@@ -297,6 +347,112 @@ export default {
...
+        }
+      } else {
+        this.show();
+      }
+    },
+    onSearchFieldEnterKeyDown() {
+      this.closeViaKeyboardNavigation();
+    },
+    onTabKeyDown() {
+      this.closeViaKeyboardNavigation();
```
mcibique, line 400: I think we should be able to close the dropdown on ESC too. That's the default behaviour of `<select>`.

src/components/ec-dropdown-search/ec-dropdown-search.vue
jafcalvente: Yes, I like this form much better, especially for not having to save the state and can calculate it dynamically. Thank you!

src/components/ec-dropdown-search/ec-dropdown-search.vue
```diff
@@ -262,15 +287,34 @@ export default {
...
+      }
     },
     show() {
-      this.isOpen = true;
-      this.$emit('open');
+      if (!this.isOpen) {
+        const focusedElement = this.$refs.popover.$el.querySelector(':focus');
+        if (focusedElement) {
+          // necessary to regain the focus after tab/enter keyboard event if search feature is active
+          this.initialFocusedElementType = focusedElement.localName;
```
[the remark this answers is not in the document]
jafcalvente: Much clearer this way!

matallo:
Awesome PR, very good work. I'm doing a proper review, but first I wanted to mention that when trying the component in Chameleon I missed being able to focus on the CTA with the keyboard navigation, maybe we can implement that in a further future story.

src/components/ec-dropdown/ec-dropdown.spec.js
```diff
@@ -217,187 +209,21 @@ describe('EcDropdown', () => {
       });
 
       it('should emit open event when the input is clicked(click mock)', () => {
-        const wrapper = mountDropdownMultipleValue({ items });
+        const wrapper = mountDropdownSingleValue({ items });
```
matallo, line 212: nitpicking, shall we rename this to `mountDropdown` as is?

matallo:
Hey Jose, incredible work. The solution is novel but I can see some quirks that I'd rather fully address in a new issue.

When navigating with the arrow keys (and the dropdown closed), those who have a CTA seem to not update. Also, I wouldn't really navigate to the CTAs (being a link or a button) with the arrow keys but with the tab key, which is the _native_ behaviour. Similarly to the searchbox, which currently autofocus but I do believe one should focus/blur with the tab.

As said, the work done in the PR goes the extra mile, and I see this complete, let's address those comments (CTA functionality) in a separate task or let's have a quick 2 min call.

jafcalvente:
Ok team, finally the CTA keyboard navigation will be implemented in a separate story (https://fxsolutions.atlassian.net/browse/ONL-4816). I've removed code regarding this last feature and now the code should work properly only with the selectable items. Thank you all for your feedback!

PR #350, opened 2020-08-15
feat(ONL-4779) Inline Input Field component

jafcalvente:
New component in order to allow edit a field and save/cancel the changes.

The component requires at least a `label`, passed as an attribute, and some content, passed inside the component tag.

The behavior is defined by the `is-editable` attribute.
- If it is `false`, the component just displays the label, using the `tw-mini-header`, class and the content.

- If it is `true` it will require also the `value` attribute. Then we'll see the text passed via the `value` attribute and, at their right, a button to enter into the edit mode, which allows changing the information using an input field.

The buttons to accept or cancel the changes in the edit mode emit events, and so the parent will be able to handle the information properly.

The component has integrated keyboard navigation.


src/components/ec-inline-input-field/components/read-only/read-only.vue
```diff
@@ -0,0 +1,77 @@
...
+      />
+    </button>
+  </div>
+</template>
+
+<script>
+import EcIcon from '@/components/ec-icon';
+
+export default {
+  name: 'EcInlineInputFieldReadOnly',
```
matallo, line 28: I don't think it's the best naming, actually I think it is a bit misleading the fact that it is editable it makes it not read only (another thing is the component state).

I'd use something like `value-text.vue` or similar, not totally sure.

src/components/ec-inline-input-field/ec-inline-input-field.story.js
```diff
@@ -0,0 +1,88 @@
...
+            :value="value"
+            @cancel="onCancel"
+            @edit="onEdit"
+            @submit="onSubmit"
+          >
+            {{ value }}
+          </ec-inline-input-field>
+        </div>
+        <div class="tw-col-full md:tw-col-4">
+          <ec-inline-input-field label="Inline input field - Uneditable">
```
matallo, line 72: Capitalize for consistency

src/components/ec-inline-input-field/components/loading/loading.vue
```diff
@@ -0,0 +1,43 @@
+<template>
+  <div
+    class="tw-flex"
+    data-test="ec-inline-input-field-loading"
+  >
+    <div class="tw-flex tw-flex-grow tw-truncate tw-border tw-border-solid tw-border-gray-6 tw-rounded">
```
mcibique, line 6: this is chameleon, we should not use tw classes directly on elements, create BEM classes for it.

src/components/ec-inline-input-field/components/loading/loading.vue
```diff
@@ -0,0 +1,43 @@
...
+    </div>
+  </div>
+</template>
+
+<script>
+import EcIcon from '@/components/ec-icon';
+import EcLoading from '@/components/ec-loading';
+
+export default {
+  name: 'EcInlineInputField',
```
mcibique, line 34: EcInlineInputFieldLoading

src/components/ec-inline-input-field/components/edit/edit.vue
```diff
@@ -0,0 +1,118 @@
...
+      value: null,
+    };
+  },
+  watch: {
+    status: {
+      immediate: true,
+      handler() {
+        this.value = this.originalValue;
+        this.$nextTick(() => {
+          this.$refs.input.$el.querySelector('input').focus();
```
mcibique, line 75: I'm not very big fan of parent being aware of what's in the children. I know it was used like this somewhere in ec-dropdown, and I see how quickly bad design spreads. There is not much to do here, just make it a little better, we can create a method in ec-input-field:
```
methods: {
  focus() {
    this.$refs.input.focus();
  }
}
```
and then here you have that function exposes and you can use it in this component as `this.$refs.input.focus()`. I'm not still satisified with this solution but at least we got rid of `querySelector('input')`.

src/components/ec-inline-input-field/components/read-only/read-only.vue
```diff
@@ -0,0 +1,77 @@
...
+  components: { EcIcon },
+  props: {
+    value: {
+      default: '',
+      type: String,
+    },
+    gainFocus: {
```
mcibique, line 35: The same problem with focus here, this is not a real property. I'd rather use the same approach which I suggested above.
Create a method in this component:
```
focus() {
  this.$nextTick(() => {
     this.$refs.editButton.focus();
   });
}
```
add ref to this component in the parent and call that function from the parent:
```
    cancel(data) {
      this.$emit('cancel');
      this.$refs.readOnly.focus();
    },
```

src/enums/input-status.js
```diff
@@ -0,0 +1,3 @@
+export const EDITING = 'editing';
```
mcibique, line 1: I wouldn't put this into src/enums. Who is going to reuse this? It is internal state enum of the ec-inline-input-field, keep it in its folder.

src/components/ec-inline-input-field/ec-inline-input-field.vue
```diff
@@ -0,0 +1,101 @@
...
+  },
+  computed: {
+    isEditing() {
+      return this.isEditable && this.status === EDITING;
+    },
+    isReadOnly() {
+      return this.isEditable && this.status === READ_ONLY;
+    },
+  },
+  watch: {
```
mcibique, line 78: This is the reason why I suggested to use v-model. You wouldn't need this intermediate innerValue and this watcher. You shouldn't be updating the innerValue at all. After submit you should just notify the parent via `this.$emit('submit', data.value);` and let the parent update the value and pass it back to this component. v-model is suitable for it.

src/components/ec-inline-input-field/ec-inline-input-field.vue
```diff
@@ -0,0 +1,101 @@
...
+  props: {
+    label: {
+      default: '',
+      type: String,
+    },
+    isEditable: {
+      type: Boolean,
+      default: false,
+    },
+    status: {
```
mcibique, line 52: All components we have are using `is-loading` prop. I'd continue with that. My problem is how to use this code in my component:
```
<ec-inline-input-field :status="LOADING">
...
import { LOADING, EDITING } from '@/enums';
...
export default {
  data() {
    return { LOADING, EDITING }
  }
}
```
I'd prefer two booleans I can set in the template:
```
<ec-inline-input-field :is-loading="true" :is-editing="true">
```

src/components/ec-inline-input-field/ec-inline-input-field.vue
```diff
@@ -0,0 +1,101 @@
+<template>
+  <div
+    class="tw-truncate"
+    data-test="ec-inline-input-field"
+  >
+    <div class="tw-mini-header">{{ label }}</div>
```
mcibique, line 6: This div should be a label in editing mode and it should point to the ID of the <ec-input-field>. Otherwise we will have just an empty input without a label. Accessibility panel complains about it too.

src/components/ec-inline-input-field/components/read-only/read-only.vue
```diff
@@ -0,0 +1,77 @@
...
+
+<style>
+.ec-inline-input-field-read-only {
+  &__action {
+    @apply tw-border-0;
+    @apply tw-bg-transparent;
+    @apply tw-outline-none;
+  }
+
+  &__action:focus &__edit-icon {
```
mcibique, line 68: Same problem with BEM here, we should not be nesting classes.

src/components/ec-inline-input-field/components/read-only/read-only.vue
```diff
@@ -0,0 +1,77 @@
+<template>
+  <div
+    class="tw-w-full tw-flex tw-justify-between tw-items-center"
+    data-test="ec-inline-input-field-read-only"
+  >
+    <span class="tw-truncate">{{ value }}</span>
+    <button
```
mcibique, line 7: make it `type="button"` otherwise it would submit the `<form>` if included in it.

src/components/ec-inline-input-field/components/read-only/read-only.vue
```diff
@@ -0,0 +1,77 @@
...
+      ref="editButton"
+      class="ec-inline-input-field-read-only__action"
+      data-test="ec-inline-input-field-read-only__action"
+      @click="edit"
+      @keydown.enter.space.prevent="edit"
```
mcibique, line 12: You can remove this line, it is done by the browser for you.

src/components/ec-inline-input-field/components/edit/edit.vue
```diff
@@ -0,0 +1,118 @@
...
+      data-test="ec-inline-input-field-edit__input"
+      @keydown.enter="submitViaKeyboard"
+      @keydown.esc="cancelViaKeyboard"
+    />
+    <div class="tw-flex tw-ml-20 tw-mt-12">
+      <button
+        class="ec-inline-input-field-edit__action"
+        data-test="ec-inline-input-field-edit__submit-action"
+        @click="submit"
+        @keydown.enter.space.prevent="submitViaKeyboard"
```
mcibique, line 19: You don't need this function as well.  keep `@click="submit"` and you can check in submit that $event has a `code` prop set. keydown fires KeyboardEvent which has it, mouse click fires MouseEvent which doesn't.

src/components/ec-inline-input-field/components/edit/edit.vue
```diff
@@ -0,0 +1,118 @@
...
+          class="ec-inline-input-field-edit__action-icon"
+          name="simple-check"
+          :size="16"
+        />
+      </button>
+      <button
+        class="ec-inline-input-field-edit__action tw-ml-8"
+        data-test="ec-inline-input-field-edit__cancel-action"
+        @click="cancel"
+        @keydown.enter.space.prevent="cancelViaKeyboard"
```
mcibique, line 31: You don't need this one as well.

src/components/ec-inline-input-field/ec-inline-input-field.vue
```diff
@@ -0,0 +1,101 @@
...
+    <div v-if="isEditable">
+      <ec-inline-input-field-read-only
+        v-if="isReadOnly"
+        :value="value"
+        :gain-focus="gainFocus"
+        @edit="edit"
+      />
+      <ec-inline-input-field-edit
+        v-else-if="isEditing"
+        :original-value="value"
```
mcibique, line 16: Why not using name `value`? Or v-model would be even better.

src/components/ec-inline-input-field/components/edit/edit.vue
```diff
@@ -0,0 +1,118 @@
...
+  data() {
+    return {
+      value: null,
+    };
+  },
+  watch: {
+    status: {
+      immediate: true,
+      handler() {
+        this.value = this.originalValue;
```
mcibique, line 73: I disagree with this line. Ff status changes, why are you updating completely different data with value from different property?

src/components/ec-inline-input-field/components/edit/edit.vue
mcibique, line 75: Second though here, why are you focusing the input every time the status changes? If submitting takes 30 seconds, user might be doing something else. After submitting is done, parent will change the status from LOADING to IS_EDITING and it will re-focus the input. I'm not sure that's right.

src/components/ec-inline-input-field/ec-inline-input-field.story.js
```diff
@@ -0,0 +1,88 @@
+import { storiesOf } from '@storybook/vue';
+import { text, boolean } from '@storybook/addon-knobs';
+import { action } from '@storybook/addon-actions';
+import { EDITING, LOADING, READ_ONLY } from '@/enums/input-status';
+import EcInlineInputField from './ec-inline-input-field.vue';
+
+const stories = storiesOf('Inline input Field', module);
```
mcibique, line 7: Inline Input Field

src/components/ec-inline-input-field/ec-inline-input-field.story.js
```diff
@@ -0,0 +1,88 @@
...
+
+stories
+  .add('basic', () => ({
+    components: { EcInlineInputField },
+    props: {
+      valueFromKnob: {
+        default: text('Initial value', 'Initial value'),
+      },
+      isEditable: {
+        default: boolean('Is editable?', true),
```
mcibique, line 17: We don't add `?` in the knob labels.

src/components/ec-inline-input-field/ec-inline-input-field.story.js
```diff
@@ -0,0 +1,88 @@
...
+      },
+      error: {
+        default: boolean('Error when saving?', false),
```
mcibique, line 20: `Show error when saving.`

MaruaEl:
Hi @jafcalvente!
Just a small thing, I was wondering - what the design team said about switching between the pencil icon and the confirm delete icons? 

Because it jumps between one and the other? 

image
image

jafcalvente:
@MaruaEl, They commented that, when we are in the initial state and we press the pencil button we show the input and the two buttons. When we press the confirm icon we show the spinner inside the input, and the buttons disabled, and when the operation is finished we go back to the initial state, no matter if the operation was successful or not.

src/components/ec-inline-input-field/components/read-only/read-only.vue
jafcalvente: Initially, I used the `gainFocus` data due to the submit process. In the case of cancel process there would be no problem, because it is an action that occurs immediately: the user cancels, the read-only mode is displayed and the button can gain focus immediately. However, when we have a submit, the read-only mode is not shown immediately; we will go to the loading mode and then to the read-only mode, so I thought I needed that data to know, at the moment the component becomes read-only, if the pencil button should gain the focus or not.
Anyway, this functionality has been removed at all because I realized by another of yours comments that it doesn't make sense try to put the focus in the pencil icon programmatically because the submit is an asynchronous operation, and as you said, user can already be working in a different part when we retrieve the response.

src/enums/input-status.js
jafcalvente: The file has been removed.

src/components/ec-inline-input-field/ec-inline-input-field.vue
jafcalvente: Absolutely right. It makes perfect sense to use `v-model` for a case like this. I'm still not used to thinking about this directive at a custom component level.
Even so, I think I would still need that `innerValue` (I've renamed it to `valueForLoading` for clarity). I mean, initially, I had to include that way of preserving the inner temporal value because of the "loading" mode. When we are in that mode we need to show what the user wrote when they were in edit mode without changing really the global `value`, since it is the parent's job, so I think I need that data to show it in the "loading" mode during the time the `submit` was emitted and we are waiting for the parent to save the new `value` or discard it.

src/components/ec-inline-input-field/components/edit/edit.vue
jafcalvente: This functionality was removed (reason explained in a previous comment). Anyway, like curiosity, it seems that if you want to use the `$event.code` you need to specify the type of event within the tag. Otherwise (if we only define the `@click`, the callback is called, but even when it is done through the keyboard, what is received is finally a `MouseEvent` (instead of a `KeyboardEvent`).

src/components/ec-inline-input-field/components/edit/edit.vue
jafcalvente: Removed. Now is working with `v-model`.

src/components/ec-inline-input-field/components/edit/edit.vue
jafcalvente: Each time the `status` was changed (and this was set to a different value than `editing`) the component was removed (due to the `v-if` directive in the parent component), so it should only be entered here when the component is re-mounted. But the truth is that it was a horrible solution, taking into account that we already have a hook precisely for that purpose. I have taken that piece of code to the "mounted" method.

matallo (approved):

stavros-tomas (approved):
Great work Jose :)

matallo (approved):

mcibique (approved):

PR #359, opened 2020-08-24
fix: Removed webpack aliases

mcibique:
After installing the 1.0.5 in EBO, webpack started throwing errors that `@/enums/key-code` doesn't exist. It's because we should not be using webpack aliases in this library because `@/` alias can change the value in the application.

We could use them in stories and spec.js but then we cannot spot problem that the alias is being used in the "production" code (vue and js files). I decided to remove them completely from storybook and jest. We were using it only on a few places and most of the usage didn't make sense anyway.

### DIFF
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -1,11 +1,9 @@
 /* eslint-disable import/no-extraneous-dependencies */
-const path = require('path');
 const StyleLintPlugin = require('stylelint-webpack-plugin');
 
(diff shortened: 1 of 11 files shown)


earnubs (approved):
LGTM

jafcalvente (approved):
After seeing this, I also updated my PR. Great to know it. It makes sense, but I hadn't thought about it.

matallo (approved):

PR #372, opened 2020-08-27
ONL 4853 - inline input copy success error

stavros-tomas:
Success error state for the inline input component

some extra bits where added as well

- Circle CI failing fix
- clear all mocks on jest config


src/components/ec-inline-input-field/ec-inline-input-field.spec.js
```diff
@@ -140,4 +158,87 @@ describe('EcInlineInputField', () => {
...
+        },
+      );
+      await wrapper.vm.$nextTick();
+
+      expect(wrapper.element).toMatchSnapshot();
+      expect(wrapper.findByDataTest('ec-inline-input-field-copy').exists()).toBeTruthy();
+    });
+
+    it('should show the success tooltip after succesfully triggering the copy method', async () => {
+      mockClipboardCopySuccess();
```
stavros-tomas, line 193: This does make the promise to be resolved and enter inside "then" where the value of "isCopied" changes from "false" to "true". If console log it we can see this is the case. The tooltip though during testing doesn't see this change. Similar issue with the error tooltip which even though the test pass is a false positive.

src/components/ec-inline-input-field/ec-inline-input-field.spec.js
```diff
@@ -140,4 +158,87 @@ describe('EcInlineInputField', () => {
...
+    it('should throw an error if the tooltip props were not given', () => {
+      withMockedConsole((errorSpy) => {
+        mountInlineInputField(
+          {
+            isEditable: false,
+            isCopiable: true,
+          },
+        );
+        expect(errorSpy).toHaveBeenCalledTimes(2);
+        expect(errorSpy.mock.calls[0][0]).toContain('Invalid prop: type check failed for prop "tooltipTextSuccess"');
```
stavros-tomas, line 172: Most likely this need some work as well

src/components/ec-inline-input-field/components/copy/copy.vue
```diff
@@ -25,19 +35,51 @@
...
     copy() {
-      clipboardCopy(this.value);
+      clipboardCopy(this.value)
+        .then(() => {
+          this.isCopied = true;
+          this.showTooltip = true;
+        })
+        .catch(() => {
+          this.isCopied = false;
+          this.showTooltip = true;
```
jmesao, line 81: @stavros-tomas Could you move this line to a .finally?

src/components/ec-inline-input-field/ec-inline-input-field.spec.js
```diff
@@ -142,29 +157,88 @@ describe('EcInlineInputField', () => {
...
+      await wrapper.vm.$nextTick();
+
+      expect(clipboardCopy).toHaveBeenCalledTimes(1);
+      expect(wrapper.findByDataTest('ec-inline-input-field-copy__icon').attributes('mocked-tooltip-content')).not.toBe(tooltipTextError);
+      expect(wrapper.findByDataTest('ec-inline-input-field-copy__icon').attributes('mocked-tooltip-content')).toBe(tooltipTextSuccess);
+      expect(wrapper.findByDataTest('ec-inline-input-field-copy__icon').attributes('mocked-tooltip-classes')).not.toBe('ec-tooltip--bg-error');
+      expect(wrapper.findByDataTest('ec-inline-input-field-copy__icon').attributes('mocked-tooltip-classes')).toBe('ec-tooltip--bg-success');
+    });
+
+    it('should show the error tooltip after unsuccesfully triggering the copy method', async () => {
```
mcibique, line 214: unsuccessfully

src/components/ec-inline-input-field/ec-inline-input-field.spec.js
```diff
@@ -142,29 +157,88 @@ describe('EcInlineInputField', () => {
...
+          tooltipTextError,
+        },
+      );
+      await wrapper.vm.$nextTick();
+
+      expect(wrapper.element).toMatchSnapshot();
+      expect(wrapper.findByDataTest('ec-inline-input-field-copy').exists()).toBeTruthy();
+    });
+
+    it('should show the success tooltip after succesfully triggering the copy method', async () => {
```
mcibique, line 192: successfully

src/components/ec-inline-input-field/ec-inline-input-field.spec.js
mcibique: It is not updating because the mock for the tooltip doesn't update its values. It's simple, it only has `bind` so it works only if component is mounted. For updates you have to change the mock:
```
jest.mock('../../directives/ec-tooltip', () => {
  const MockedTooltipDirective = {
    bind(el, { value }) {
      if (value.content) {
        el.setAttribute('mocked-tooltip-content', value.content);
        el.setAttribute('mocked-tooltip-classes', value.classes);
      }
    },
    update(el, { value }) {
      if (value.content) {
        el.setAttribute('mocked-tooltip-content', value.content);
        el.setAttribute('mocked-tooltip-classes', value.classes);
      }
    },
  };
  return MockedTooltipDirective;
});
```

src/components/ec-inline-input-field/ec-inline-input-field.spec.js
```diff
@@ -1,11 +1,32 @@
...
+    bind(el, { value }) {
+      // this can be undefined that's why we check it
+      if (value.content) {
+        el.setAttribute('mocked-tooltip-content', value.content);
+        el.setAttribute('mocked-tooltip-classes', value.classes);
+      }
+    },
+    update(el, { value }) {
+      if (value.content) {
+        el.setAttribute('mocked-tooltip-content', value.content);
```
jmesao, line 18: @stavros-tomas Those lines are repeated on lines 12 and 13, could we create a fn?

mcibique:
@matallo I have fixed jest running out of memory when running on CI. It's because jest automatically determines how many workers it can uses by `#cpus / 2` -> which yields 1 on CI. That's not enough, I increased it 2 by setting max-workers=2 in package.json. 

Additionally, I have stopped Cypress to install and download the binary when running on CI. We do not run Cypress on CI anyway, so that should speed up the installation.

matallo (approved):

src/components/ec-inline-input-field/ec-inline-input-field.spec.js
stavros-tomas: Thank you @jmesao :)

jmesao (approved):

mcibique (approved):

PR #387, opened 2020-09-03
feat(ONL-4824): Adding Chinese language fonts

jafcalvente:


src/styles/generic/fonts.story.js
```diff
@@ -46,6 +46,62 @@ stories.add('fonts', () => ({
         <p style={style}>
           ABCČĆDĐEFGHIJKLMNOPQRSŠTUVWXYZŽabcčćdđefghijklmnopqrsštuvwxyzžАБВГҐДЂЕЁЄЖЗЅИІЇЙЈКЛЉМНЊОПРСТЋУЎФХЦЧЏШЩЪЫЬЭЮЯабвгґдђеёєжзѕиіїйјклљмнњопрстћуўфхцчџшщъыьэюяΑΒΓΔΕΖΗΘΙΚΛΜΝΞΟΠΡΣΤΥΦΧΨΩαβγδεζηθικλμνξοπρστυφχψωάΆέΈέΉίϊΐΊόΌύΰϋΎΫΏĂÂÊÔƠƯăâêôơư1234567890‘?’“!”(%)[#]@/&\-+÷×=®©$€£¥¢:;,.*
         </p>
+        <div class="tw-mini-header">English</div>
```
mcibique, line 49: You can remove the lorem ipsum paragraphs now, and also the paragraph with special characters. Most of them are already covered in specific languages.

I'd create at the bottom and section "Special characters" and only add remaining that are not covered:
`1234567890‘?’“!”(%)[#]@/&\-+÷×=®©$€£¥¢:;,.*`

src/styles/generic/fonts.css
```diff
@@ -63,6 +63,24 @@
   unicode-range: U+0000-00FF, U+0131, U+0152-0153, U+02BB-02BC, U+02C6, U+02DA, U+02DC, U+2000-206F, U+2074, U+20AC, U+2122, U+2191, U+2193, U+2212, U+2215, U+FEFF, U+FFFD;
 }
 
+/* chinese-simplified */
+@font-face {
+  font-family: 'Noto Sans SC';
+  font-style: normal;
+  font-weight: 300;
+  src: local('Noto Sans SC Light'), local('NotoSansSC-Light'), url('https://fonts.gstatic.com/ea/notosanssc/v1/NotoSansSC-Light.woff2') format('woff2');
```
mcibique, line 71: Holly molly, it's 1.7 MB each! Is it because of all Chinese characters? Can we check if there is version of Noto Sans without Latin, Greek and Japanese, Korean. Roboto has 10 KB in comparison.

I see that the Traditional text in the story need three fonts to render properly:
```
Noto Sans SC Regular—Network resource(57 glyphs)
Noto Sans TC Regular—Network resource(16 glyphs)
Roboto—Network resource(9 glyphs)
```
^^ that's 3.5 MB of fonts to render. and if the font is bold, that's another 3.5 MB to download.

jafcalvente: I downloaded the font files and, based on the size, it seems that other Asian characters than Chinese are already excluded.
Comparing both of them, local files and remote ones, the local ones are a bit lighter, but not too much (probably the remote ones referenced through fonts.gstatic include the Latin characters). It looks like each woff2 chinese file has at least 1.4 MB: https://google-webfonts-helper.herokuapp.com/fonts/noto-sans-sc?subsets=chinese-simplified

So far I haven't been able to find lighter files than these. Here are some screenshots with the weights of the different files (we are importing only the last two, simplify and traditional Chinese):
image

jafcalvente: Finally, we go for the approach of using system fonts. Removed the fonts for the Chinese language.

mcibique (approved):

PR #392, opened 2020-09-07
Onl 4868 - Adding validations to ec-inline-input

stavros-tomas:
This is part of of the story of adding validation to ec-inline-input

Now we can pass the prop errorMessage and display the validation error in similar fashion with the ec-input-field.


src/components/ec-inline-input-field/components/edit/edit.vue
```diff
@@ -101,6 +106,8 @@ export default {
   &__actions {
     @apply tw-flex tw-items-center;
     @apply tw-ml-20;
+
+    height: 42px; /* height of input field */
```
stavros-tomas, line 110: When error message is visible the height of input field changes. This makes the icons expand as well in height and thy stop being "relatively centered" with the input field box. This will force them to keep the same height always. If you don't like this solution maybe we can amend for something else.

stavros-tomas:
Cypress refuses to update the firefox snap. If someone can push it would be great.

src/components/ec-inline-input-field/components/edit/edit.vue
mcibique: It should be at least extracted into a variable, because nobody will remember to update this if the height has changed.

ec-input-field already has:
```
:root {
  --ec-input-field-icon-area-size: 42px;
}
```
you can expand that to 
```
:root {
  --ec-input-field-height: 42px;
  --ec-input-field-icon-area-size: var(--ec-input-field-height);
}
```
and then re-use the new variable in your edit.vue.

matallo (approved):

mcibique (approved):

PR #395, opened 2020-09-09
Onl 4867 give input components is sensitive config

stavros-tomas:
I added a "isSensitive" prop to the following components.

-ec-input
-ec-inline-input (copy,edit,loading,value text sub-components)
-ec-amount
-ec-currency-input
-ec dropdown
-ec-dropdown-search

This prop will apply our "sensitive class" if it set.

I created a config.file where you can set the class. If you try to use the prop without to set the class will throw an error.

To showcase this on storybook I included an isSensitive knob on the relavant stories and added this plugin https://www.npmjs.com/package/tailwindcss-filters to apply a blur to the input when the knob is true. The blurring doesn't happen on production.


src/components/ec-dropdown/ec-dropdown.spec.js
```diff
@@ -133,6 +136,11 @@ describe('EcDropdown', () => {
       const wrapper = mountDropdown({ items, isLoading: true });
       expect(wrapper.element).toMatchSnapshot();
     });
+
+    it('should render with a sensitive class when isSensitive prop is set to true when isSensitive prop is set to true', () => {
```
matallo, line 140: duplicated spec text or (as I see in the component below) you wanted to mention about the nested `isSensitive`?

matallo (changes requested):
Screenshot 2020-09-09 at 15 42 45

something seems to be missing

also, shall we add the sensitive knob to the Storybook stories?

postcss.config.js
```diff
@@ -15,7 +15,10 @@ module.exports = ({ file }) => {
   if (isProd && shouldPurgeFile) {
     // see https://purgecss.com/guides/vue.html for reference
     config.plugins['@fullhuman/postcss-purgecss'] = {
-      content: ['./src/**/*.vue', './src/**/*.story.js'],
+      content: [
+        './src/**/*.vue',
+        './src/**/*.story.js',
+        './.storybook/*.js'],
```
stavros-tomas, line 21: @jaimeperezpedra This will change to './.preview/*.js' after you merge your pr

matallo (approved):

postcss.config.js
jaimeperezpedra: the postcss.config.js is not changing, I changed the config.js from the .storybook folder

jaimeperezpedra (approved):

PR #414, opened 2020-09-18
ONL-4887 Added whitespace no wrap to column headers

pelocho:


src/components/ec-table-head/ec-table-head.vue
```diff
@@ -152,8 +155,12 @@ export default {
     }
   }
 
-  &__cell-text--has-max-width {
-    @apply tw-truncate;
+  &__cell-text {
+    @apply tw-whitespace-no-wrap;
```
mcibique, line 159: I'm not sure why this was added.

mcibique (commented):
There are failing tests in the build.

src/components/ec-table-head/ec-table-head.vue
pelocho: This is because when the headers are too long in some languages, the text line breaks. That way we avoid it.

matallo (approved):

jaimeperezpedra (approved):

MaruaEl (approved):

jafcalvente (approved):

jmesao (approved):

PR #437, opened 2020-10-02
ONL-5041 add new icons for fees status

MaruaEl:
- Hello, this Pr will added two new icons, one for the Partially paid status, one for the Returned status. The Partially paid icon needs to have 2 different colors, so we hardcoded the constant one directly in the svg file - as that color will always be gray and only the first part of the icon will change. 
image
the second icon is:
image


src/assets/icons/rounded/partial.svg
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="16px" height="16px" viewBox="0 0 16 16" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <title>Graphics / 00 Icon / Rounded / Partial</title>
+    <g id="Graphics-/-00-Icon-/-Rounded-/-Partial" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
+        <g id="Group" transform="translate(1.000000, 1.070889)">
+            <path d="M6,13.8582223 C2.60770586,13.3729926 0,10.4555831 0,6.92911114 C0,3.40263923 2.60770586,0.485229661 6,2.30926389e-14 L6,13.8582223 Z" id="Left-side" fill="#15BC8B"></path>
+            <path class="tw-fill-gray-4" d="M8,2.30926389e-14 C11.3922941,0.485229661 14,3.40263923 14,6.92911114 C14,10.4555831 11.3922941,13.3729926 8,13.8582223 L8,2.30926389e-14 Z" id="Right-side" fill="#6C7F84"></path>
```
MaruaEl, line 7: Here we have the hardcoded class

jaimeperezpedra (approved):

jafcalvente (changes requested):
You need to update the visual regression snapshots with these two new rounded icons.

mcibique:
The class tw-fill-gray-4 is always from removed from the final CSS because purgecss cannot find it used anywhere. Purgecss doesn't run on your local but it does run when deploying to now.sh so the class has no effect there.

You have to add SVGs to be scanned into `postcss.config.js`:
```js
config.plugins['@fullhuman/postcss-purgecss'] = {
      content: [
        './src/**/*.vue',
        './src/**/*.story.js',
        './src/**/*.svg',
        './.storybook/*.js'
      ],
```

You can check how is it working/not working by setting 
```
// const isProd = process.env.NODE_ENV === 'production';
const isProd = true;
```
in the same file. You have to restart `npm run start` every time you change that file.

MaruaEl:
> The class tw-fill-gray-4 is always removed from the final CSS because purgecss cannot find it used anywhere. Purgecss doesn't run on your local but it does run when deploying to now.sh so the class has no effect there.
> 
> You have to add SVGs to be scanned into `postcss.config.js`:
> 
> ```js
> config.plugins['@fullhuman/postcss-purgecss'] = {
>       content: [
>         './src/**/*.vue',
>         './src/**/*.story.js',
>         './src/**/*.svg',
>         './.storybook/*.js'
>       ],
> ```
> 
> You can check how it is working/not working by setting
> 
> ```
> // const isProd = process.env.NODE_ENV === 'production';
> const isProd = true;
> ```
> 
> in the same file. You have to restart `npm run start` every time you change that file.

Good spot!

pelocho (approved):

MaruaEl:
> You need to update the visual regression snapshots with these two new rounded icons.

All done Jose - thank you for your review

matallo (approved):

jafcalvente:
Hey @MaruaEl, it's true that you already have the chrome snapshots uploaded but we still need also the firefox ones

jmesao (approved):

MaruaEl:
> Hey @MaruaEl, it's true that you already have the chrome snapshots uploaded but we still need also the firefox ones

I think now should be it @jafcalvente

PR #451, opened 2020-10-05
ONL-4900: feat: Added pagination

mcibique:
EcSmartTable has a new prop: `isPaginationEnabled`, set to false by default. Currently balances table in EBO doesn't need pagination so I made it optional. 

I didn't create a story for ec-table-pagination because it doesn't make sense to reuse this component anywhere else outside of the table (just like we don't have ec-table-head, ec-table-footer, etc.)

The biggest development issue is that we would have to provide a translations for current page string for every table we use in EBO. The text says `1 of 10 pages (67 payments)`. So every table will require its own translation. The default value is:
```
      <slot
        name="pages"
        v-bind="{ page, numberOfItems, totalPages, total }"
      >
        <span>{{ page }}&nbsp;of&nbsp;{{ totalPages }} pages</span>
        <span>({{ total }}&nbsp;items)</span>
      </slot>
```
^^ `&nbsp;` are there because I don't want the breaking spaces between `1 of 6` and `XX items`. I really don't want to break this into 3 slots and having these slots wrapped in `tw-whitespace-nowrap`. And I don't want to have 3 props for 3 texts either.

### DIFF
--- a/src/components/ec-smart-table/ec-smart-table.spec.js
+++ b/src/components/ec-smart-table/ec-smart-table.spec.js
@@ -262,7 +262,7 @@ describe('EcSmartTable', () => {
       expect(wrapper.findByDataTest('ec-table-footer').element).toMatchSnapshot();
     });
 
-    it('should render footer slot inside of the pagination when pagination is not enabled', async () => {
+    it('should render footer slot inside of the pagination when pagination is enabled', async () => {
       const wrapper = await mountEcSmartTableWithResolvedData(lotsOfData, {
         columns,
         isPaginationEnabled: true,
--- a/src/components/ec-smart-table/ec-smart-table.spec.js
+++ b/src/components/ec-smart-table/ec-smart-table.spec.js
@@ -11,8 +11,8 @@ describe('EcSmartTable', () => {
   ];
 
   const data = {
-    count: 3,
-    total: 3,
+    count: 1,
+    total: 1,
     items: [[1, 2, 3]],
   };
 
@@ -29,6 +29,9 @@ describe('EcSmartTable', () => {
   function mountEcSmartTable(props, mountOpts) {
     return mount(EcSmartTable, {
       propsData: { dataSource, ...props },
+      stubs: {
+        EcPopover: true,
+      },
       ...mountOpts,
     });
   }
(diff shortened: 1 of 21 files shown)


src/components/ec-table-pagination/ec-table-pagination.spec.js
```diff
@@ -0,0 +1,130 @@
...
+    const wrapper = mountEcTablePagination({ numberOfItems: 50 });
+    expect(wrapper.findByDataTest('ec-table-pagination__page-size').element).toMatchSnapshot();
+  });
+
+  it('should display next and prev buttons disabled if pagination has only one page', () => {
+    const wrapper = mountEcTablePagination({ numberOfItems: 50, total: 50, page: 1 });
+    expect(wrapper.findByDataTest('ec-table-pagination__actions').element).toMatchSnapshot();
+  });
+
+  it('should display only next button disabled if pagination is on the last page', () => {
```
matallo, line 62: personally I'd join these with a describe "when pagination is on the last page" with the same above, reusing the wrapper in a `beforeAll`

this works too 👍

matallo (approved):
I'm sincerely feeling goosebumps seeing this component work ❤️

stavros-tomas (approved):

src/components/ec-table-pagination/ec-table-pagination.vue
```diff
@@ -0,0 +1,224 @@
...
+    },
+    pageSizeModel: {
+      get() {
+        return this.pageSizeItems.find(item => item.value === this.numberOfItems);
+      },
+      set(pageSizeItem) {
+        this.$emit('change', 1, pageSizeItem.value);
+      },
+    },
+    seletedPageSizeText() {
```
MaruaEl, line 143: `selectedPageSizeText` typo?

src/components/ec-table-pagination/ec-table-pagination.vue
```diff
@@ -0,0 +1,224 @@
...
+      <ec-dropdown-search
+        v-model="pageSizeModel"
+        :items="pageSizeItems"
+        :is-search-enabled="false"
+      >
+        <button
+          class="ec-table-pagination__action ec-table-pagination__action--page-size"
+          data-test="ec-table-pagination__action--page-size"
+        >
+          {{ seletedPageSizeText }}
```
MaruaEl, line 24: same as below?

mgascam (approved):

src/components/ec-smart-table/ec-smart-table.spec.js
```diff
@@ -195,4 +198,98 @@ describe('EcSmartTable', () => {
...
+          footer() {
+            return (<div>My custom footer</div>);
+          },
+        },
+      });
+
+      expect(wrapper.findByDataTest('ec-table-footer').element).toMatchSnapshot();
+    });
+
+    it('should render footer slot inside of the pagination when pagination is not enabled', async () => {
```
jmesao, line 265: @mcibique Do you mean when pagination is enabled?

src/hocs/ec-with-pagination/ec-with-pagination.spec.js
```diff
@@ -0,0 +1,51 @@
...
+
+    return { hocWrapper, componentWrapper };
+  }
+
+  it('should render properly', () => {
+    const { hocWrapper } = mountEcWithPagination();
+    expect(hocWrapper.element).toMatchSnapshot();
+  });
+
+  it('should update the pagination whe pagination event is emitted', async () => {
```
jmesao, line 31: when*

PR #471, opened 2020-10-12
ONL-5119 - Fix false triggering of an action when disabled

stavros-tomas:
An inline action when has disabled set to true it shouldn't trigger the action.

This is behaving correctly when the action renders as a button.
When it renders as a link though it does not (if we set href, then will render as a link).

This PR is for fixing the "href/disabled" scenario.


src/components/ec-inline-actions/ec-inline-actions.vue
```diff
@@ -88,6 +88,9 @@ export default {
     },
   },
   methods: {
+    allowedActions(item) {
```
mcibique, line 91: `allowedActions` is a misleading name. It sounds like a getter which gives you all allowed actions. 

What about `callAction(item)` or doAction(item)

stavros-tomas: Amended to "doAction". Thank you

matallo (approved):

MaruaEl (approved):

PR #483, opened 2020-10-19
ONL-4910 ecFilterPopover

MaruaEl:
image

Create a box that shows the list of all the options for a specific filter and allow users to select/unselect the items
eg. Click on a filter item (e.g. fee type and a list of options will appear in this box) that you can select/deselect eg via checkboxes

- should contain an empty `ec-popover` with slot for the content
- should have the trigger with label, badge icon and expand icon (https://fxsolutions.atlassian.net/browse/ONL-4909)
- should hold in its state if it’s open/closed.
- should have props label and number-of-selected-filters
Scrollbar is not exactly the same as the design - shall we bring it up with Uve? 
Screenshot 2020-10-19 at 16 58 35


mcibique:
The trigger has a focus and hover states that are not implemented:
Screenshot from 2020-10-20 09-10-39

When there is no space on the left, the popover is displayed overflowed.
Screenshot from 2020-10-20 09-13-18

src/components/ec-filter-popover/ec-filter-popover.story.js
```diff
@@ -0,0 +1,115 @@
+import { text } from '@storybook/addon-knobs';
+import EcFilterPopover from './ec-filter-popover.vue';
+import EcCheckbox from '../ec-checkbox/ec-checkbox.vue';
+
+export default { title: 'Filter Popover' };
```
mcibique, line 5: Can we create a separate section for filters, just like we have for layout and CSS? Something like `Filters / Filter Popover`

src/components/ec-filter-popover/ec-filter-popover.story.js
```diff
@@ -0,0 +1,115 @@
...
+  },
+  template: `
+    <div class="tw-flex tw-flex-row tw-m-auto">
+      <ec-filter-popover :label="labelOne">
+        <div slot="filter">
+          <ec-checkbox
+            v-model="item.checkbox"
+            v-for="(item, index) in itemListOne"
+            :key="index"
+            class="tw-col-12"
```
mcibique, line 74: `col-` classes should be used as children of tw-grid. There isn't any in this template. What exactly is this for?

src/components/ec-filter-popover/ec-filter-popover.vue
```diff
@@ -0,0 +1,104 @@
+<template>
+  <div
+    class="ec-filter-popover"
+    data-test="ec-filter-popover"
+  >
+    <ec-popover
+      popover-class="ec-popover"
```
mcibique, line 7: Why do you have to re-declare this? I think the default value is already set to ec-popover.

src/components/ec-filter-popover/ec-filter-popover.vue
```diff
@@ -0,0 +1,104 @@
...
+      placement="bottom"
+      trigger="click"
```
mcibique, line 9: I think default trigger is click.

src/components/ec-filter-popover/ec-filter-popover.vue
```diff
@@ -0,0 +1,104 @@
...
+      level="modal"
```
mcibique, line 10: Why it must be a modal?

src/components/ec-filter-popover/ec-filter-popover.vue
```diff
@@ -0,0 +1,104 @@
...
+      @update:open="clickTrigger"
+    >
+      <div class="ec-filter-popover__filter-item">
```
mcibique, line 13: "Item" should be an item of a list. This div is a trigger which opens the popover content, I'd name it `ec-filter-popover__trigger`.

src/components/ec-filter-popover/ec-filter-popover.vue
```diff
@@ -0,0 +1,104 @@
...
+        <a class="ec-filter-popover__label">{{ label }}</a>
```
mcibique, line 14: Anchor should have href="#", or this should be a `<button>` without background, padding and border.

src/components/ec-filter-popover/ec-filter-popover.vue
```diff
@@ -0,0 +1,104 @@
...
+    },
+    numberOfSelectedFilters: {
+      type: Number,
+      required: false,
+      default: 0,
+    },
+  },
+  data() {
+    return {
+      isOpen: false,
```
mcibique, line 57: This boolean is not used anywhere, so it can be removed.

src/components/ec-filter-popover/ec-filter-popover.vue
```diff
@@ -0,0 +1,104 @@
...
+    };
+  },
+  methods: {
+    clickTrigger(status) {
```
mcibique, line 61: ClickTrigger is a noun, functions should have a verb in the name or in this case, it should be `onOpen`. But I'm still not sure why do you need to emit isOpen event from this component? If the other components don't need this, follow YAGNI principle and remove it.

src/components/ec-filter-popover/ec-filter-popover.vue
```diff
@@ -0,0 +1,104 @@
...
+      this.isOpen = status;
+      this.$emit('isOpen', this.isOpen);
```
mcibique, line 63: You can emit status directly, the name of the event should be just `open`, so we can use it in the parent as
```
<ec-filter-popover @open="doSomething()">
```
instead of
```
<ec-filter-popover @is-open="doSomething()">
```

src/components/ec-filter-popover/ec-filter-popover.vue
```diff
@@ -0,0 +1,104 @@
...
+  &__filter-content {
+    @apply tw-bg-gray-8;
+    @apply tw-px-20 tw-py-4;
+    @apply tw-border-gray-6 tw-border-solid tw-border;
+    @apply tw-overflow-y-scroll;
+
+    @mixin small-scrollbar;
+
+    width: var(--ec-filter-popover-width);
+    height: var(--ec-filter-popover-height);
```
mcibique, line 101: Not every component must have this height, e.g. date range filter:
Screenshot from 2020-10-20 09-28-04

I think it should be only max-height, and we will let the specific filters to specify the fixed height if needed.

src/components/ec-filter-popover/index.js
```diff
@@ -0,0 +1 @@
+export { default } from './ec-filter-popover.vue';
```
mcibique, line 1: We should re-export it in main.js as well.

src/components/ec-filter-popover/ec-filter-popover.spec.js
```diff
@@ -0,0 +1,51 @@
+import { shallowMount } from '@vue/test-utils';
+import EcFilterPopover from './ec-filter-popover.vue';
+import { withMockedConsole } from '../../../tests/utils/console';
+
+const label = 'Test label';
+
+function shallowMountEcFilterPopover(props) {
+  return shallowMount(EcFilterPopover, {
```
mcibique, line 8: We should use shallowMount only as a last option, in this case, I'd rather use mount() and stub EcPopover instead. In the future, this component will have ec-number-badge and that component will be stubbed by shallowMount too, so we cannot test what it renders here. With mount and stubbed EcPopover, we will not have that problem.

src/components/ec-filter-popover/ec-filter-popover.vue
```diff
@@ -0,0 +1,104 @@
...
+      popover-class="ec-popover"
+      placement="bottom"
+      trigger="click"
+      level="modal"
+      @update:open="clickTrigger"
+    >
+      <div class="ec-filter-popover__filter-item">
+        <a class="ec-filter-popover__label">{{ label }}</a>
+        <!-- <ec-badge-number size="20" value="numberOfSelectedFilters" />
+        TODO Badge with https://fxsolutions.atlassian.net/browse/ONL-4909 -->
```
mcibique, line 16: For initial implementation, we can add 
```
<span v-if="numberOfSelectedFilters > 0">({{ numberOfSelectedFilters }})</span>
```

So we can test numberOfSelectedFilters prop in specs, in SB story

src/components/ec-filter-popover/ec-filter-popover.vue
mcibique, line 14: Can we add a data-test for this anchor so we have something to target in integration tests.

src/components/ec-filter-popover/ec-filter-popover.story.js
```diff
@@ -0,0 +1,115 @@
...
+
+export default { title: 'Filter Popover' };
+
+export const Primary = () => ({
+  components: { EcFilterPopover, EcCheckbox },
+  data() {
+    return {
+      itemListOne: [{
+        name: 'Test Bank Name',
+        checkbox: null,
```
mcibique, line 13: This prop should be called `selected` or `selectedValue`.

src/components/ec-filter-popover/ec-filter-popover.story.js
```diff
@@ -0,0 +1,115 @@
...
+      default: text('beneficiary', 'Beneficiary'),
+    },
+    // numberOfSelectedFilters: {
+    //   default: number('numberOfSelectedFilters', 0),
+    // } TODO with https://fxsolutions.atlassian.net/browse/ONL-4909
+  },
+  template: `
+    <div class="tw-flex tw-flex-row tw-m-auto">
+      <ec-filter-popover :label="labelOne">
+        <div slot="filter">
```
mcibique, line 69: You can also do `<template #filter>` instead of adding more divs.

mcibique (commented):
There are some changes in package-lock.json that should not be committed.

src/components/ec-filter-popover/ec-filter-popover.vue
```diff
@@ -0,0 +1,104 @@
...
+  &__icons {
+    @apply tw-flex-shrink-0;
+    @apply tw-self-center;
+    @apply tw-mr-24;
+  }
+
+  &__filter-content {
+    @apply tw-bg-gray-8;
+    @apply tw-px-20 tw-py-4;
+    @apply tw-border-gray-6 tw-border-solid tw-border;
```
mcibique, line 95: I think the border radius is also needed for these borders, otherwise it looks like this:

Screenshot from 2020-10-20 09-46-28

src/components/ec-filter-popover/ec-filter-popover.vue
```diff
@@ -0,0 +1,104 @@
...
+
+  &__icons {
+    @apply tw-flex-shrink-0;
+    @apply tw-self-center;
+    @apply tw-mr-24;
+  }
+
+  &__filter-content {
+    @apply tw-bg-gray-8;
+    @apply tw-px-20 tw-py-4;
```
mcibique, line 94: This component should not be specifying any paddings because each filter will have specific criteria, e.g. multi value filter will need to draw full width delimiter. that won't be possible if there is a padding already here:

Screenshot from 2020-10-20 09-50-47

Same problem will be with tabs:
Screenshot from 2020-10-20 09-52-17

src/components/ec-filter-popover/ec-filter-popover.vue
```diff
@@ -0,0 +1,104 @@
...
+    @apply tw-border-gray-6 tw-border-solid tw-border;
+    @apply tw-overflow-y-scroll;
+
+    @mixin small-scrollbar;
```
mcibique, line 98: This component should not have the scrollbar, I'd add it into specific filters, e.g. multi value needs scrollbar for the multiple values checkboxes, but the search and tabs should be fixed.

Screenshot from 2020-10-20 09-53-21

src/components/ec-filter-popover/ec-filter-popover.vue
```diff
@@ -0,0 +1,104 @@
...
+  &__filter-item {
+    @apply tw-flex tw-flex-row tw-items-center;
+  }
+
+  &__label {
+    @apply tw-text-gray-3;
+    @apply tw-mr-8 tw-ml-24;
+  }
+
+  &__icons {
```
mcibique, line 86: There is no class __icons in the template. This can be removed.

src/components/ec-filter-popover/ec-filter-popover.vue
MaruaEl: In the snapshots was `popover-class=""` thought I needed to re-declare it

src/components/ec-filter-popover/ec-filter-popover.vue
MaruaEl: in snapshot was `level=""` though that I needed to specify something and thought modal was the pertinent one

src/components/ec-filter-popover/ec-filter-popover.vue
mcibique: That might have been generated before you used shallowMount(). It doesn't make sense to re-declare this, look what it generates in the browser:

Screenshot from 2020-10-20 13-07-40

ec-popover is there twice. Just remove this and update snapshots.

src/components/ec-filter-popover/ec-filter-popover.vue
mcibique: We should override the level only if it's really needed. some components that are using ec-popover might be used inside modal dialog, e.g. ec-dropdown. If ec-dropdown is not used inside of a modal, modal's z-index should be greater than z-index of ec-popover in ec-dropdown. if ec-dropdown is used inside of a modal, then z-index of modal should be smaller than z-index of ec-popover. That's why we have those levels.

In this case, ec-filter-popover will never be used inside of a modal, so it should always have default level (not set at all) so the modal which has higher z-index can be displayed over the filter.

Just don't set this value.

src/components/ec-filter-popover/ec-filter-popover.story.js
MaruaEl: is this correct or you meant `name`?

src/components/ec-filter-popover/ec-filter-popover.vue
MaruaEl: I'v added the span with some styling - but we'd have to remove this styling once we work on ONL-4909. Is that ok?

src/components/ec-filter-popover/ec-filter-popover.vue
MaruaEl: Just thought that we might need it on ebo - I'll remove it

src/components/ec-filter-popover/ec-filter-popover.story.js
MaruaEl: I was cheating - I've changed it to now - sorry

src/components/ec-filter-popover/ec-filter-popover.vue
MaruaEl: I've removed the emit for now

src/components/ec-filter-popover/ec-filter-popover.story.js
mcibique: It should be 
```
itemListOne: [{
        name: 'Test Bank Name',
        selected: false,
}, {
```

so then in the template is used like this:
```
<ec-checkbox
            v-model="item.selected"
```

You can also use isSelected or selectedValue as a name of the prop, it's up to you.

src/components/ec-filter-popover/ec-filter-popover.vue
matallo: IMO that's fine, we'll pick it up from there

src/components/ec-filter-popover/ec-filter-popover.story.js
matallo: Actually we've started doing this in other places as a rule of thumb

matallo:
check tests, the build is not passing

matallo:
I don't think this is necessary wrong, but confirm with @Ebury/ux-design that the height of the anchor has to be the same when it has filters applied and when it doesn't

Screenshot 2020-10-22 at 14 01 25

matallo (approved):

src/components/ec-filter-popover/ec-filter-popover.spec.js
```diff
@@ -0,0 +1,72 @@
...
+    const wrapper = mountEcFilterPopover({ label, numberOfSelectedFilters });
+    expect(wrapper.element).toMatchSnapshot();
+  });
+
+  it('numberOfSelectedFilters should be visible if it\'s value is greater than 0 ', () => {
+    const wrapper = mountEcFilterPopover({ label, numberOfSelectedFilters: 5 });
+    expect(wrapper.element).toMatchSnapshot();
+  });
+
+  it('should render given default slot', () => {
```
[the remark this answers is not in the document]
MaruaEl: Ops, thank you for spotting it. I'll be more careful next time

src/components/ec-filter-popover/ec-filter-popover.vue
```diff
@@ -0,0 +1,198 @@
...
+      :placement="'bottom'"
+      offset="0, 8"
+      @update:open="onOpen"
+    >
+      <div
+        class="ec-filter-popover__trigger"
+        :class="{ 'ec-filter-popover__trigger--focused': triggerIsFocused }"
+        data-test="ec-filter-popover__trigger"
+      >
+        <button
```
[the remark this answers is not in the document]
MaruaEl: True true! updating it right away

mcibique (dismissed):

matallo (approved):

PR #485, opened 2020-10-20
ONL-4920 add light mode to banner

pelocho:
image


jaimeperezpedra (dismissed):

mgascam (dismissed):

src/styles/settings/colors/reserved-color-scale.css
```diff
@@ -1,10 +1,14 @@
 :root {
   --ec-reserved-color-error: 0, 72%, 56%;
```
MaruaEl, line 2: @jaimeperezpedra  @pelocho  are this still used? the default is going to be used anywhere? From what I understood the light one is going to be the new default? or am I wrong?

MaruaEl (approved):

src/styles/settings/colors/reserved-color-scale.css
pelocho: Right now the default is being still used, so we can't remove it

src/styles/components/ec-btn/ec-btn.css
```diff
@@ -194,11 +194,11 @@
...
 
     &.ec-btn--primary-reverse {
-      @apply tw-text-gray-8;
+      @apply tw-text-info-hover;
 
       &:hover {
-        @apply tw-bg-gray-8;
-        @apply tw-text-key-4;
+        @apply tw-text-gray-8;
+        @apply tw-bg-info-hover;
```
mcibique, line 201: I think you forgot to change --secondary-reverse. Now these buttons look like this:
Screenshot from 2020-10-22 09-29-16

src/components/ec-alert/ec-alert.vue
```diff
@@ -136,11 +138,43 @@ export default {
   }
 
   &__title {
-    @apply tw-body-strong tw-text-gray-8;
+    @apply tw-body-strong;
```
mcibique, line 141: can you try this?

```
&__title {
  @apply tw-body-strong; 
}

&__subtitle {
  @apply tw-body-text;
}

&__title,
&__subtitle {
  &--success {
      @apply tw-text-success-hover;
    }

  ...
}
```

src/components/ec-alert/ec-alert.vue
```diff
@@ -153,54 +187,87 @@ export default {
...
     line-height: 0;
 
     &:focus {
       @apply tw-outline-none;
-      @apply tw-bg-gray-8;
     }
 
-    &:hover {
-      @apply tw-bg-gray-8;
+    &:hover,
```
mcibique, line 197: This block of code generates invalid CSS and doesn't work
Screenshot from 2020-10-22 09-38-33

src/styles/settings/colors/reserved-color-scale.css
```diff
@@ -1,10 +1,14 @@
...
+  --ec-reserved-color-error-hover: 0, 72%, 44%;
+  --ec-reserved-color-error-light: 0, 59%, 94%;
   --ec-reserved-color-info: 205, 100%, 56%;
-  --ec-reserved-color-info-hover: 205, 100%, 46%;
+  --ec-reserved-color-info-hover: 205, 100%, 37%;
+  --ec-reserved-color-info-light: 205, 58%, 94%;
   --ec-reserved-color-success: 162, 80%, 41%;
-  --ec-reserved-color-success-hover: 162, 80%, 31%;
+  --ec-reserved-color-success-hover: 162, 80%, 26%;
+  --ec-reserved-color-success-light: 164, 28%, 92%;
```
mcibique, line 10: I think H should be 162, just like the two other success colors. L should be 94% just like the other 3 light colors.

src/styles/settings/colors/reserved-color-scale.css
```diff
@@ -1,10 +1,14 @@
...
   --ec-reserved-color-warning: 36, 100%, 55%;
   --ec-reserved-color-warning-hover: 36, 100%, 45%;
+  --ec-reserved-color-warning-light: 37, 79%, 94%;
```
mcibique, line 13: H should be 36, just like the other 2 warning colors.

src/styles/settings/colors/reserved-color-scale.css
mcibique: I think we should rename error-hover to be error-dark and do it for every type. The hover is no longer for hover, title and subtitle uses it for text without `:hover` or `:focus`. Double check *-hover is not used in EBO and chameleon and rename them.

src/components/ec-alert/ec-alert.vue
```diff
@@ -153,54 +187,87 @@ export default {
...
+    &:focus {
```
matallo, line 198: I came to comment the same than Milos, modifiers can't be appended to pseudoclasses (well, they can with some special syntax in scss, but I really wouldn't use that here.

Could we apply the modifier, then nest the psuedoclass?

matallo (changes requested):

src/styles/settings/colors/reserved-color-scale.css
pelocho: I put the colors that UX team used in the design, but you're right!

src/styles/settings/colors/reserved-color-scale.css
pelocho: You're right! In the next commit, It will be changed

src/styles/components/ec-btn/ec-btn.css
pelocho: Yes, I forgot that one! Fix in the next commit

src/components/ec-alert/ec-alert.vue
pelocho: It works! Thank you for this solution!

src/styles/settings/colors/reserved-color-scale.css
mcibique: yeah, we had this issue in the past too. I have adjusted few colors that didn't match existing patterns. These discrepancies happens between converting colors from hex/rgb to HSL so I wouldn't be scared of slightly changing them if nobody can tell the difference anyway.

src/components/ec-alert/ec-alert.vue
pelocho: You're right, I've just tested it, and doesn't apply anything, so we can remove it!

src/components/ec-alert/ec-alert.vue
pelocho: I've just removed it! It wasn't applying anything and the component was working in the same way

src/components/ec-alert/ec-alert.vue
```diff
@@ -153,54 +174,68 @@ export default {
...
-    &:hover {
-      @apply tw-bg-gray-8;
     }
 
     &--info:focus,
     &--info:hover {
       @apply tw-fill-info;
     }
 
+    &--info {
```
matallo, line 189: specificity takes care of this (I think) but IMO we should have this prior to the psuedoclasses avobe and nest

```
    &--info {
      ...
      &:focus,
      &:hover {
      @apply tw-fill-info;
    }
```

pelocho: You're right! I've just changed it! Thanks!

matallo (approved):

mgascam (approved):

PR #500, opened 2020-10-30
ONL-4917 date range filter component

MaruaEl:
~~WIP Pr to gather feedbacks~~
~~Working on unit tests in the main time~~
~~looking at the failed build~~

Moving this branch from WIP to PR 
Running the visual regression tests for snapshots now


src/components/ec-date-range-filter/ec-date-range-filter.vue
```diff
@@ -0,0 +1,156 @@
...
+      >
+        <div
+          class="ec-date-range-filter__inputs-wrapper"
+          data-test="ec-date-range-filter__inputs-wrapper"
+        >
+          <ec-input-field
+            v-model="fromValueDate"
+            class="ec-date-range-filter__from-label"
+            type="date"
+            placeholder="DD/MM/YYYY"
```
matallo, line 20: if we're not supporting yet different locales for our date fields the least we can do is being consistent how we're showing the placeholder (see capitalization) https://bitbucket.org/fxsolutions/eburyonline/src/ba86b02b8b4b249a806b314b41682158588804c5/frontend/src/views/trade-finance/submitted-requests/create-request/components/form/form.vue#lines-265

is the time to start thinking about a date field component? it has commonalities such error messages

src/components/ec-date-range-filter/ec-date-range-filter.vue
```diff
@@ -0,0 +1,156 @@
...
+            :label="fromLabelText"
+            :error-message="errorMessageFromDate"
+            @change="onChange()"
+            @input="errorMessageFromDate = ''"
```
matallo, line 24: not sure I find it the best way to do this

src/components/ec-date-range-filter/ec-date-range-filter.vue
```diff
@@ -0,0 +1,156 @@
...
+export default {
+  name: 'EcDateRangeFilter',
+  components: { ecFilterPopover, EcInputField },
+  props: {
+    label: {
+      type: String,
+      required: true,
+      default: '',
+    },
+    value: { // Check we actually need this
```
matallo, line 61: do we? AFAIK, we need the separate from and to values, and not something we'll compose for later use, probably we can get rid of this

matallo (changes requested):

matallo:
It's very subtle but the tab reflows when selecting a date and the badge appears, the rest LGTM!

matallo (approved):

MaruaEl:
> It's very subtle but the tab reflows when selecting a date and the badge appears, the rest LGTM!

Thank you Carlos for checking it - I think the reflow will be fixed when we develop the badge - because the heigh and width of the badge should be 15px - so a bit smaller than the current one. Maybe I'll add a comment on the badge ticket so we'll be sure that that is done correctly

matallo:
Fair enough!

stavros-tomas (approved):

jaimeperezpedra (approved):

mcibique (approved):

PR #511, opened 2020-11-03
[WIP] - ONL-5053 - Metroline item

stavros-tomas:
Initial PR for metroline item.

While working on metroline container I might have to make some changes to this component. I'll keep this one open until I finish with metroline container.

PR #521, opened 2020-11-09
ONL - 5051 Metroline

stavros-tomas:
This PR includes both Metroline Container and Metroline Item

**Provide/Inject**
Instead of the classic communication of components with props/events this time we went for the provide/inject api.
With Provide we "provide" every child that is nested inside the metroline container an object called "metroline".

This object the children (in this case metroline items) can inject it to themselves and in that way they can use all the data/methods that the parent provides.

**Registering/Unregistering**
Upon creation(created hook) each metro item knows its id so it "registers" with the appropriate method that the parent provides with that particular Id. Similarly for when destroyed. 

**Id**
Visually the order of the items will always render as they are in the template. To know though which item is active and where to go next we pass as a prop an id. The metro container holds an array of those ids and from there we do the appropriate calculations for next/previous items.

This potentially could have been automated but we hard-coded it as a bulletproof solution in case one metro item is hidden and then it becomes visible. 

For example think that that we had 3 items. The second is not visible. So At the beginning we register [first, third]. Now we make the second visible. The order will be [first, third, second]. This means the order will be incorrect when we try to click "continue" to go to next item. On the contrary if we pass the id as prop we avoid this kind of issues.

@events
When we change from one item to another we emit an event called "change". When we complete the metroline we emit an event called "complete". The latter is particularly handy if we want to trigger something else in the view that holds the metroline. For example after metroline is complete show another section.

**Stories and Testing**
Because of this dependency between metro container and metro item we have 1 story and 1 test as you cannot display them individually. Also this means that I can only test them when mounted as templates.

**Metro Items**
For each slot of a metro item we expose the following
`
v-bind="{ activateItem, goToNext, isLast, isReadOnly, isNext, isActive, isCompleted }"
`

Those can be used later like this 
`   
 <template #footer-cta="{ goToNext , isReadOnly }">
     <a href v-if="!isReadOnly" @click="goToNext"> Continue </a>
     <a href="#" v-else> Do Something Else </a>
 </template>
`

OLD PR for metroline-time served its purposed and will be closed without to be merged. You can still find it [here](https://github.com/Ebury/chameleon/pull/511)


stavros-tomas:
Kudos to @mcibique for amazing provide/inject Idea

src/components/ec-metroline-item/ec-metroline-item.vue
```diff
@@ -0,0 +1,240 @@
...
+export default {
+  name: 'EcMetrolineItem',
+  components: { EcIcon },
+  inject: ['metroline'],
+  props: {
+    id: {
+      type: Number,
+      required: true,
+    },
+    badgeText: {
```
stavros-tomas, line 124: @jafcalvente @matallo Following conversation from previous PR badgeText

src/components/ec-metroline-item/ec-metroline-item.vue
```diff
@@ -0,0 +1,240 @@
...
+  &__header {
+    @apply tw-flex tw-flex-wrap tw-justify-between tw-items-center;
+    @apply tw-mb-12;
+  }
+
+  &__header-heading {
+    @apply tw-text-gray-3 tw-h4;
+
+    .ec-metroline-item--is-active & {
+      font-weight: 500; /* For this specific case only, an h4 with font-weight 500. Applying a stong/b tag in the html would produce a weight 400 so is not an option */
```
stavros-tomas, line 218: Following previous conversation. Spoke to designers and we can make h4 bolder for this case and will solve our alignment issues. Tried b/strong tag on the template before I try it via css but doesn't seem to work

src/components/ec-metroline-item/ec-metroline-item.vue
```diff
@@ -0,0 +1,240 @@
...
+        >
+          <slot
+            v-if="!isReadOnly"
+            name="header-cta"
+            v-bind="{ goTo, goToNext, status, isLast, isReadOnly }"
+          />
+
+          <slot
+            v-else
+            name="header-cta-complete"
```
stavros-tomas, line 75: @mcibique @matallo @jafcalvente I added this slot. After we complete the metroline we have an "Add description" cta. We want at the same time "edit" to don't be visible. We could hard code above and say 
```
    <div
          v-if="isCompleted && !isReadOnly"
          data-test="ec-metroline-item__header-cta"
        >
```

so this would hide the "edit" cta when metroline is complete. This though we also hide the field for "Add description"

We could also leave at the developer judgement to remember when we implement the metroline for each metro item to do

```
          <template #header-cta="{ goTo , isReadOnly }">
            <a href v-if="!isReadOnly" @click="goTo"> Edit </a>
            <a href="#" v-else> Add Description </a>
          </template>
```

But this also felt unnecessary work for the developer during implementation for each item and also bit risky in case he forgets you can complete a trade and go back to edit.

So i went for an after completion header slot, "header-cta-complete"

Let me know if you agree

jaimeperezpedra (approved):

matallo (approved):
😻

src/components/ec-metroline/ec-metroline.story.js
```diff
@@ -0,0 +1,165 @@
...
+          </template>
+
+          <template #header-cta-complete>
+            <a href="#" @click.prevent.stop class="tw-flex tw-items-center">
+              Add Description
+            </a>
+          </template>
+
+          <template #main="{ status }">
+            <div v-if="status === MetrolineItemStatus.COMPLETED" class="tw-bg-gray-7 tw-min-h-104 tw-p-24">
```
mcibique, line 54: I think a better solution would be bind isCompleted, isNext, isActive computed props from the component so you don't have to recalculate them here again. Especially MetrolineItemStatus enum doesn't have to be added to the data of the component's template.

Don't forget to bind them for every slot, just like goTo, goToNext, isReadOnly, etc.

src/components/ec-metroline/ec-metroline.story.js
```diff
@@ -0,0 +1,165 @@
...
+          </template>
+
+          <template #sub-heading>
+            <span class="tw-flex tw-items-center tw-mr-16">
+              <ec-icon name="simple-check" class="tw-fill-current tw-mr-8" :size="16" />
+              Amount fully allocated
+            </span>
+          </template>
+
+          <template #header-cta="{ goTo }">
```
mcibique, line 87: Second opinion wanted: I wrote this function but I'm still not sure about the name, I'm thinking about `activateItem` instead of `goTo`. The problem is that the usage of this function might be confusing: `@click.prevent.stop="goTo"` doesn't say where the code "goes to". It's item internals which will automatically add `goTo(item.id)`. 

I'd change only the activateItem in the bind. goTo is completely fine inside of the metroline and metroline-item.

src/components/ec-metroline/ec-metroline.vue
```diff
@@ -0,0 +1,67 @@
...
+          if (nextMetrolineItemId) {
+            this.metroline.activeItemId = nextMetrolineItemId;
+            this.$emit('change', this.metroline.activeItemId);
+          } else {
+            this.metroline.isCompleted = true;
+            this.$emit('complete');
+          }
+        },
+        goTo: (id) => {
+          this.metroline.activeItemId = id;
```
mcibique, line 53: We should add 
```
this.$emit('change', this.metroline.activeItemId);
```
just like we have in `goToNext()`.

src/components/ec-metroline/ec-metroline.spec.js
```diff
@@ -0,0 +1,369 @@
...
+          </template>
+        </ec-metroline-item>
+      </ec-metroline>`,
+      {
+        methods: {
+          onChange,
+        },
+      });
+
+      wrapper
```
mcibique, line 171: you can await the trigger('click') and drop the `await wrapper.vm.$nextTick();`.

src/components/ec-metroline/ec-metroline.spec.js
```diff
@@ -0,0 +1,369 @@
...
+          </template>
+        </ec-metroline-item>
+      </ec-metroline>`,
+      {
+        methods: {
+          onComplete,
+        },
+      });
+
+      wrapper
```
mcibique, line 210: await. Please check all trigger('click') calls in the spec file.

src/components/ec-metroline/ec-metroline.story.js
stavros-tomas: Makes sense. I was presenting yesterday to @jaimeperezpedra and momentarily he was bit confused what "goTo" will do on the template. Will swap to this on the metro item and all the binds

```   
   activateItem() {
      this.metroline.goTo(this.id);
    },
```

src/components/ec-metroline/ec-metroline.story.js
stavros-tomas: Removed status since we will bind the other 3

PR #526, opened 2020-11-11
ONL-5036 ec multiple values selection

MaruaEl:
image

Static list of filters for multiple selection


src/components/ec-multiple-values-selection/ec-multiple-values-selection.spec.js
```diff
@@ -0,0 +1,125 @@
...
+
+  it('should not be visibile if isSearchable is set to false', () => {
+    const wrapper = mountEcMultipleValuesSelection({ items, isSearchable: false });
+
+    expect(wrapper.element).toMatchSnapshot();
+  });
+
+  it('should emit selected items', async () => {
+    const wrapper = mountEcMultipleValuesSelection({ items });
+    await wrapper.findByDataTest('ec-multiple-values-selection__checkbox-select').findByDataTest('ec-checkbox__input').trigger('click');
```
MaruaEl, line 83: This is the solution found with @jaimeperezpedra as I didn't managed to get to trigger the click with only ec-multiple-values-selection__checkbox-select. What do you think guys?

stavros-tomas: As long as you selecting with data-test should be fine. Double findByDataTest is ok. Did the the same with Metroline and we do the same as well at the smart table spec.

src/components/ec-multiple-values-selection/ec-multiple-values-selection.vue
```diff
@@ -0,0 +1,275 @@
...
+    emptyMessage: {
+      type: String,
+    },
+    emptyIcon: {
+      type: String,
+    },
+    isSearchable: {
+      type: Boolean,
+    }, // TODO with https://fxsolutions.atlassian.net/browse/ONL-4912
+    canSelectAll: {
```
stavros-tomas, line 162: Maybe something with "is" since a boolean like the above "isSearchable"?

matallo: while I usually agree, what would this be? `isSelectAllable` on the other hand we need to keep sane naming and a bit of `canSomething` won't hurt and improve readability? Other thoughts?

MaruaEl: isSelectable? is english?

stavros-tomas:
There is a weird bug not sure if we want to address.

Open in a relative reduced height browser
image

Close and start clicking
image

It will increase the items selected

src/components/ec-multiple-values-selection/ec-multiple-values-selection.vue
```diff
@@ -0,0 +1,275 @@
...
+      type: Array,
+      required: true,
+      default: () => ([]),
+    },
+    value: {
+      type: Array,
+      required: false,
+      default: () => ([]),
+    },
+    loading: {
```
matallo, line 145: `isLoading`?

src/components/ec-multiple-values-selection/ec-multiple-values-selection.vue
```diff
@@ -0,0 +1,275 @@
...
+        data-test="ec-multiple-values-selection__search-field"
+        @change="$emit('search', query)"
+      />
+    </div>
+
+    <!-- The above will be completed with TODO with https://fxsolutions.atlassian.net/browse/ONL-4912.
+    Visible only if we pass dataSource and if isSearchable is set to true.
+    With static items list it won't be available-->
+    <ec-loading
+      :show="loading"
```
matallo, line 29: same as below

src/components/ec-multiple-values-selection/ec-multiple-values-selection.vue
```diff
@@ -0,0 +1,275 @@
...
+    >
+      <div
+        v-if="noResults"
```
matallo, line 32: `hasNoResults`, `!hasResults`, `isEmpty` better!

src/components/ec-multiple-values-selection/ec-multiple-values-selection.vue
```diff
@@ -0,0 +1,275 @@
...
+    notSelectedItems() {
+      const itemsSet = new Set(this.selectedItems.map(item => item.value));
+      return this.items.filter(item => !itemsSet.has(item.value));
+    },
+    // TODO ONL-4911
+    allItemsAreSelected() {
+      return this.notSelectedItems.length === 0;
+    },
+    // TODO ONL-4919
+    noResults() {
```
matallo, line 190: I'm pretty sure we can split this with the `emptyMessage` and `isEmpty` ? ps: naming to be decided

matallo (commented):
mostly it makes sense, but I'm finding a lot of issues regarding naming, myabe I think the best would be to do a quick call / poll and refine them

matallo (approved):

stavros-tomas:
Coming back to the bug issue it seems to be related with EcLoading. If you remove it everything works as expected.

src/components/ec-multiple-values-selection/ec-multiple-values-selection.vue
```diff
@@ -0,0 +1,275 @@
...
+      required: true,
+      default: () => ([]),
+    },
+    value: {
+      type: Array,
+      required: false,
+      default: () => ([]),
+    },
+    loading: {
+      type: Boolean,
```
stavros-tomas, line 146: default:false for now? :show prop is required I think that what creates the issue

stavros-tomas: The bug we have happens irrelevant of browser height as a I mentioned earlier.

The issue is that loading component overrides what popover tries to do. 
More in particular that's how popover hides itself . With `visibility: hidden` and `opacity: 0`. 

Screenshot from 2020-11-12 16-19-37

We must always set to something the `show` . If not the default visibility of the loading component will be this
image

As a result the loading will override the visibility of the popover. 
Why we don't see it? Because popover sets the opacity to 0 as well but this is still clickable.

so maybe for now default true with a comment?
`default: false // We will revert to false when we complete functionality`

stavros-tomas:
Kind of relevant with this ticket. When then badge kicks in, the whole element increases its height by 2px. So if we toogle one filter you have this "jumb" effect. Maybe...If we plan to go live before the badge component, is worth fixing this? If not all good we can fix later.

stavros-tomas (approved):

MaruaEl:
> Kind of relevant with this ticket. When then badge kicks in, the whole element increases its height by 2px. So if we toogle one filter you have this "jumb" effect. Maybe...If we plan to go live before the badge component, is worth fixing this? If not all good we can fix later.
> 
> Other than that if you remove the ec-loading for now I think you are good to go. Good Job @MaruaEl

Ye - thank you Stavros. Also @matallo pointed it out - Ive added in the icon badge ticket a note to make sure when we implement the badge component that this reflow is fixed as well

src/components/ec-multiple-values-selection/ec-multiple-values-selection.vue
MaruaEl: For now I've keep the loading and just unwrapped the content

jmesao (approved):

PR #530, opened 2020-11-12
[ONL-5059] Adding Metroline Cards component

jafcalvente:


jmesao (approved):

src/components/ec-metroline-card/ec-metroline-card.story.js
```diff
@@ -15,6 +15,9 @@ export const basic = (args, { argTypes }) => ({
       <ec-metroline-card
         class="tw-mb-24"
         :is-collapsed="isCollapsed"
+        :is-first="isFirst"
+        :is-last="isLast"
+        :is-stand-alone="isStandAlone"
```
matallo, line 20: I'm a bit confused about the `isStandAlone` part, wouldn't it be redundant with first and last true?

Also, it is not being tested in the storybook (all false).

stavros-tomas:
Memory I got confused somewhere, but didn't we say we will make those components on EBO, since are not reusable :/ ?

src/components/ec-metroline-cards/ec-metroline-cards.story.js
```diff
@@ -0,0 +1,95 @@
+import EcMetrolineCards from './ec-metroline-cards.vue';
+import EcIcon from '../ec-icon';
+import EcMetrolineCard from '../ec-metroline-card';
+
+export default {
+  title: 'Metroline Cards',
+  component: EcMetrolineCards,
+};
+
+export const basic = (args, { argTypes }) => ({
```
stavros-tomas, line 10: I think the idea with the new stories is not to do bind the template to a specific scenario so we can reuse it if we want
Then again, here we only have one scenario so as you prefer.
(https://storybook.js.org/docs/react/writing-stories/introduction#using-args )

`const Template = (args, { argTypes }) => ({`

Then at the bottom of the file you can do 

```
export const basic = Template.bind({});
export const anotherStory = Template.bind({});

basic.args = {
  isCollapsible: true,
};

anotherStory.args = {
  isCollapsible: false,
};
```

If we change this here maybe worth changing at the metroline-card as well.

jafcalvente:
You are TOTALLY right @stavros-tomas. We said we do this in EBO, not in Chameleon. I made a mistake. I'm going to move this to EBO and use this PR to remove the EcMetrolineCard component, which was included in a previous story. Thanks for pointing it out.

src/components/ec-metroline-card/ec-metroline-card.story.js
jafcalvente: you're right, the `isStandAlone` prop is actually not necessary for first and last cases, but since we still need it for `Metroline Cards` with more than two items, for the rest of items, I preferred to leave it for homogeneity. Nevertheless, I will improve it so that, if we are in the case of `first` or `last` item, there is no need to also check `isStandAlone`.

The `isStandAlone` prop is used by `Metroline Card` to know if it is being used inside a `Metroline Cards`. Its value by default is `true` (and it is like is displayed in its section in storybook). If it is used within a `Metroline Cards` component the `isStandAlone` prop must be set to `false` (and it is like is displayed in the `Metroline Cards` section in storybook).

PR #532, opened 2020-11-16
ONL 5054 - timer component

stavros-tomas:
Timer component

Has one boolean prop to start it "isTimerRunning" and one number prop to pass the seconds to countdown. Emits an event "time-expired" when complete.

If you compare with mockup you will find few differences. Those changes are approved by design. Moves clockwise not anti-clock wise, has a background circle with gray-7, stroke width is 4px not 3px.

Added also in this pr the "@sinonjs/fake-timers"


src/components/ec-timer/ec-timer.spec.js
```diff
@@ -0,0 +1,65 @@
...
+  it('should render as expected', () => {
+    const wrapper = mountTimer({ seconds: 20, isTimerRunning: true });
+
+    expect(wrapper.element).toMatchSnapshot();
+  });
+
+  it('should emit an event called "time-expired" after the countdown completes', () => {
+    const wrapper = mountTimer({ seconds: 20, isTimerRunning: true });
+
+    jest.advanceTimersByTime(20000);
```
stavros-tomas, line 61: Before I merge I need to figure out why I am getting the following error when I use advanceTimerByTime
image

Is like it doesn't have the time to validate?

matallo:
First comment before reading the code, isn't the "timer" in `isTimerRunning` redundant? We can go straight with `isRunning` right?

jafcalvente:
I'm not sure if from the design perspective they want the movement to be as it is now, taking a little jump after every second, or in a more fluid way, that the movement flows continuously. Could we consult with UX?

jafcalvente:
I closed it by mistake. Open again.

src/components/ec-timer/ec-timer.spec.js
jafcalvente: Maybe it is because invoking `jest.useFakeTimers();` at the root level makes that it applies for all tests, but actually it is needed only for the last one.
It seems that if you remove this line, and include it inside the test case, it works:
image

stavros-tomas:
@jafcalvente I feel on a similar way but to fast track this I would try to keep it MVP until a comment is made from their side.

Similar to the above thinking I didn't make radius and strokeWidth as props and kept them as data properties even though is an easy change and I already bind everything to those properties.

src/components/ec-timer/ec-timer.spec.js
stavros-tomas: Thank you! Works pushiiiing it!

matallo (approved):

src/components/ec-timer/ec-timer.vue
```diff
@@ -0,0 +1,131 @@
...
+      />
+
+      <text
+        x="50%"
+        y="50%"
+        class="ec-timer__text"
+        dominant-baseline="central"
+        text-anchor="middle"
+      >
+        {{ secondsLeft }}s
```
mcibique, line 39: I think this text should be translated. Not sure how this would be displayed in Chinese, but I'm pretty sure, Russian language doesn't have `20s` but `20c`.

src/components/ec-timer/ec-timer.vue
```diff
@@ -0,0 +1,131 @@
...
+  methods: {
+    countdown() {
+      if (this.isRunning && this.secondsLeft > 0) {
+        setTimeout(this.reduceSecondsLeft, 1000);
+      } else {
+        this.$emit('time-expired');
+      }
+    },
+    reduceSecondsLeft() {
+      this.secondsLeft--;
```
mcibique, line 106: I think this approach is wrong, bear in mind that setTimeout is paused when window loses the focus or it's hidden. Especially on mobile device, the setTimeout callback can be executed only once per 10s or 20s, on desktop, it's 5-60s. 

Thus relying that setTimeout(..., 1000) will execute after 1s will not work in every ocassion. You should get `Date.now()` when starting the countdown and remember that value. Then you should get Date.now() again in setTimeout/setInterval callback, subtract it and calculate how many seconds are left.

src/components/ec-timer/ec-timer.vue
```diff
@@ -0,0 +1,131 @@
...
+        } else {
+          this.secondsLeft = this.seconds;
+        }
+      },
+    },
+  },
+  methods: {
+    countdown() {
+      if (this.isRunning && this.secondsLeft > 0) {
+        setTimeout(this.reduceSecondsLeft, 1000);
```
mcibique, line 100: Why not using setInterval instead of setTimeout?

mcibique, line 100: You need to clear the timeout/interval in beforeDestroy event otherwise you will create a memory leak. This component will be ticking till 0 even if it has been already destroyed and removed from DOM.

src/components/ec-timer/ec-timer.vue
```diff
@@ -0,0 +1,131 @@
...
+      this.secondsLeft--;
+      this.countdown();
+    },
+  },
+};
+</script>
+
+<style>
+.ec-timer {
+  &__background {
```
mcibique, line 115: Do not use CSS terminology in class names. Use descriptive name instead, e.g. `__elapsed`.

src/components/ec-timer/ec-timer.vue
```diff
@@ -0,0 +1,131 @@
...
+    stroke: hsl(var(--ec-gray-color-level-7));
+  }
+
+  &__remaining {
+    stroke: hsl(var(--ec-key-color-level-4));
```
mcibique, line 120: `@apply tw-stroke-key-4;`

src/components/ec-timer/ec-timer.vue
```diff
@@ -0,0 +1,131 @@
...
+        fill="transparent"
+        :stroke-width="strokeWidth"
+        :stroke-dasharray="circumference"
+        :stroke-dashoffset="offset"
+      />
+
+      <text
+        x="50%"
+        y="50%"
+        class="ec-timer__text"
```
mcibique, line 35: Can we add data-test for this text, so we can read its value in E2E.

src/components/ec-timer/ec-timer.spec.js
```diff
@@ -0,0 +1,65 @@
...
+  });
+
+  it('should emit an event called "time-expired" after the countdown completes', () => {
+    jest.useFakeTimers();
+
+    const wrapper = mountTimer({ seconds: 20, isRunning: true });
+
+    jest.advanceTimersByTime(20000);
+
+    expect(wrapper.emitted('time-expired').length).toBe(1);
```
mcibique, line 63: You should also test that the text value has changed from 20s to 0s.

src/components/ec-timer/ec-timer.vue
```diff
@@ -0,0 +1,131 @@
...
+    },
+  },
+  watch: {
+    isRunning: {
+      immediate: true,
+      handler(value) {
+        if (value) {
+          this.countdown();
+        } else {
+          this.secondsLeft = this.seconds;
```
jafcalvente, line 92: @stavros-tomas , if when the timer is counting down we set to `false` the `isRunning` prop, the `secondsLeft` variable is reset (as we expect), but the timeout still keeps working, so the event is emitted at the end anyway. In this scenario, we should remove the timeout also in order to avoid emitting the `time-expired` event.

src/components/ec-timer/ec-timer.spec.js
```diff
@@ -0,0 +1,65 @@
...
+  it('should throw an error if "seconds" prop is not an integer', () => {
+    withMockedConsole((errorSpy) => {
+      mountTimer({ seconds: 20.1, isRunning: true });
+
+      expect(errorSpy).toHaveBeenCalledTimes(1);
+      expect(errorSpy.mock.calls[0][0]).toContain('Invalid prop: custom validator check failed for prop "seconds"');
+    });
+  });
+
+  it('should throw an error if "seconds" prop is negative ', () => {
```
jafcalvente, line 32: empty space after `negative` word

src/components/ec-timer/ec-timer.spec.js
```diff
@@ -0,0 +1,65 @@
...
+
+  it('should emit an event called "time-expired" after the countdown completes', () => {
+    jest.useFakeTimers();
+
+    const wrapper = mountTimer({ seconds: 20, isRunning: true });
+
+    jest.advanceTimersByTime(20000);
+
+    expect(wrapper.emitted('time-expired').length).toBe(1);
+  });
```
jafcalvente, line 64: Maybe a last test with snapshot to check that after more than 20 seconds the text still display just `0s`

jaimeperezpedra (approved):

src/components/ec-timer/ec-timer.vue
stavros-tomas: Ha, thought was universal the "s". prop then I guess

stavros-tomas: After talk with the team, will be slot

mgascam (approved):
Please ignore my comment about defining radius and stroke as props, I read it was already commented.

Just a comment about a possible improvement is to use requestAnimationFrame for the animation, I guess you already considered it

src/components/ec-timer/ec-timer.vue
```diff
@@ -0,0 +1,144 @@
...
+      validator(value) {
+        return Number.isInteger(value) && value > 0;
+      },
+    },
+    isRunning: {
+      type: Boolean,
+      required: true,
+    },
+  },
+  data() {
```
jafcalvente, line 62: `timerInterval` should be defined in `data`

src/components/ec-timer/ec-timer.vue
```diff
@@ -0,0 +1,144 @@
...
+    reduceSecondsLeft() {
+      this.currentTime = Math.ceil(Date.now() / 1000);
+      if (this.timeDifference < this.seconds) {
+        this.secondsLeft = this.seconds - this.timeDifference;
+      } else {
+        this.secondsLeft = this.seconds - this.timeDifference;
+        clearInterval(this.timerInterval);
+        this.$emit('time-expired');
+      }
+    },
```
jafcalvente, line 121: Since we are repeating the same line in both if-blocks, could we move it out? Something like:
```suggestion
    reduceSecondsLeft() {
      this.currentTime = Math.ceil(Date.now() / 1000);
      this.secondsLeft = this.seconds - this.timeDifference;
      if (this.timeDifference >= this.seconds) {
        clearInterval(this.timerInterval);
        this.$emit('time-expired');
      }
    },
```

jafcalvente (changes requested):

src/components/ec-timer/ec-timer.vue
stavros-tomas: You are absolutely right

src/components/ec-timer/ec-timer.vue
stavros-tomas: I am not sure about this.

jafcalvente (approved):

PR #537, opened 2020-11-17
[ONL-5055] Expand Currency Input component

jafcalvente:
An additional note field has been included on the left-bottom side of the component.
Also, the story file associated has been updated to the new Docs approach.


src/components/ec-currency-input/ec-currency-input.spec.js
```diff
@@ -45,6 +51,47 @@ describe('EcCurrencyInput', () => {
...
+          const wrapper = mountCurrencyInput({
+            bottomNote: 'Bottom note message',
+            isWarning: true,
+            warningMessage,
+          });
+          expect(wrapper.findByDataTest('ec-currency-input__bottom-note').element).toMatchSnapshot();
+          expect(wrapper.findByDataTest('ec-currency-input__warning-tooltip').attributes('mocked-tooltip-content')).toBe(warningMessage);
+        });
+      });
+    });
```
matallo, line 81: maybe also cover the case when there are both, the error prevails?

src/components/ec-currency-input/ec-currency-input.spec.js
```diff
@@ -45,6 +51,47 @@ describe('EcCurrencyInput', () => {
     expect(wrapper.element).toMatchSnapshot();
   });
 
+  describe('when the bottom note is defined', () => {
```
matallo, line 54: capitalize first part of the spec statement

matallo (approved):

src/components/ec-currency-input/ec-currency-input.spec.js
jafcalvente: Hey Carlos, I'm not totally sure about this, since it seems that that first capitalized word is something we only do (so far) in integration tests, not in unit ones (neither in Chameleon nor in EBO). If we want to start capitalizing it is fine with me, but I just wanted to double-check it.

src/components/ec-currency-input/ec-currency-input.spec.js
jafcalvente: yes, the error always prevails over the bottom note and the warning icon with the tooltip. I added another test case to check this scenario also. Thanks.

src/components/ec-currency-input/ec-currency-input.spec.js
```diff
@@ -45,6 +51,61 @@ describe('EcCurrencyInput', () => {
...
+          expect(wrapper.findByDataTest('ec-currency-input__warning-tooltip').attributes('mocked-tooltip-content')).toBe(warningMessage);
+        });
+
+        describe('and an error message is being displayed', () => {
+          it('should not display neither the bottom note nor the warning icon', () => {
+            const wrapper = mountCurrencyInput({
+              bottomNote: 'Bottom note message',
+              errorMessage: 'Random message',
+              isWarning: true,
+              warningMessage: 'Warning message',
```
jmesao, line 87: @jafcalvente Should we see or not the warningMessage?

src/components/ec-currency-input/ec-currency-input.vue
```diff
@@ -106,6 +128,16 @@ export default {
     note: {
       type: String,
     },
+    bottomNote: {
+      type: String,
+    },
+    warningMessage: {
```
jmesao, line 134: @jafcalvente If this prop is for the tooltip, the name is not too clear to me, maybe it could be better to call it warningTooltipMessage?

src/components/ec-currency-input/ec-currency-input.spec.js
matallo: you're correct, in that case

src/components/ec-currency-input/ec-currency-input.spec.js
jafcalvente: we shouldn't. The error message always has priority and when it is defined neither the bottom note nor the warning icon (which contains the tooltip) should be displayed.

jmesao (approved):

MaruaEl (approved):

stavros-tomas (approved):

src/components/ec-currency-input/ec-currency-input.story.js
```diff
@@ -1,60 +1,45 @@
...
+const bottomNote = 'On balance: EUR 300.00';
 const currencies = ['GBP', 'EUR', 'USD', 'CAD', 'JPY', 'AUD', 'NZD', 'CHF', 'SEK', 'NOK', 'AED', 'BBD', 'BGN', 'BHD', 'BSD', 'BWP', 'BZD', 'CNY', 'CZK', 'DKK', 'GHS', 'HKD', 'HRK', 'HUF', 'ILS', 'INR', 'JMD', 'KES', 'LKR', 'MUR', 'MXN', 'NGN', 'PHP', 'PKR', 'PLN', 'QAR', 'RON', 'RSD', 'RUB', 'SAR', 'SGD', 'SZL', 'THB', 'TND', 'TRY', 'TTD', 'UGX', 'XCD', 'ZAR', 'ZMW'];
+const label = 'Currency input';
+const note = 'Select currency and set amount';
+const warningTooltipMessage = 'Currently there is not enough EUR balance to cover the transaction amount, ensure you add funds to your balance after finishing.';
 
-const stories = storiesOf('Currency Input', module);
+export default {
+  title: 'Currency Input',
+  component: EcCurrencyInput,
```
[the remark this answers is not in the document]
jafcalvente: great contribution @stavros-tomas! Thank you

PR #543, opened 2020-11-19
[WB2-370] Blog Post Component: Creation of the blog post component

thomas-dodds-ebury:
Hi team, 

A request came in to integrate blog posts from the ebury marketing sites to EBO. 

This is my very first attempt at Vue.js and the start of building out the component all be it static. 

I would appreciate any comments or feedback as early on as possible so I grow my knowledge of vue and learn how you as a team like to code. 

BE GENTLE!


src/components/ec-blog-item/ec-blog-item.vue
```diff
@@ -0,0 +1,26 @@
+<template>
+  <div>
+    <p>{{ title }}</p>
```
mgascam, line 3: I would use a <h*> tag for the title but not sure about what are your requirements

src/components/ec-blog-item/ec-blog-item.vue
```diff
@@ -0,0 +1,26 @@
...
+    <p>{{ shortDescription }}</p>
+    <p>{{ author }}</p>
+    <p>{{ category }}</p> hello
```
mgascam, line 6: Is 'hello' intended or a left over?

src/components/ec-blog-item/ec-blog-item.vue
```diff
@@ -0,0 +1,26 @@
...
+  props: {
+    title: { type: String },
+    shortDescription: { type: String },
+    author: { type: String },
+    category: { type: String },
+  },
+};
+</script>
+<style>
+p {
```
mgascam, line 22: I wouldn't style the <p> directly. I'd rather use css classes instead

src/components/ec-blog-item/ec-blog-item.story.js
```diff
@@ -0,0 +1,41 @@
+import { storiesOf } from '@storybook/vue';
+import EcBlogItem from './ec-blog-item.vue';
+
+const stories = storiesOf('Blog Item', module);
```
jmesao, line 4: We don't use storiesOf since the v6 it's active. Take a look [here](https://fxsolutions.atlassian.net/browse/ONL-4998)

src/components/ec-blog-item/ec-blog-item.story.js
```diff
@@ -0,0 +1,41 @@
...
+
+stories.add('all', () => ({
+  components: { EcBlogItem },
+  data() {
+    return {
+          id: 1,
```
jmesao, line 10: too much tabs here?

src/components/ec-blog-item/ec-blog-item.story.js
```diff
@@ -0,0 +1,41 @@
...
+          title: 'Post 1',
+          shortDescription: 'Short 1',
+          author: 'Author 1',
+          category: 'Category 1',
+    };
+  },
+  props: {
+    title: {
+      default: text('title', 'A new title'),
```
jmesao, line 19: When I try to see the storybook link I'm getting that text is not defined
image

src/components/ec-blog-item/ec-blog-item.story.js
mcibique, line 10: The `id` is not used.

src/components/ec-blog-item/ec-blog-item.vue
mcibique, line 22: Setting the `p` style in this component will affect paragraphs in other components too. You should either:
1. use [scoped styles](https://vue-loader.vuejs.org/guide/scoped-css.html#deep-selectors) (not recommended in this project but can be used in other vue projects)
2. Use CSS classes and BEM - that's what all other components do in this project

```html
<div class="ec-blog-item">
  <p class="ec-blog-item__title">...</p>
  <p class="ec-blog-item__short-description">...</p>
  ...
```
```css
.ec-blog-item {
  &__title,
  &__short-description {
    ... your css here.
  }
}
```

src/components/ec-blog-item/ec-blog-item.vue
```diff
@@ -0,0 +1,26 @@
...
+  font-size: 2em;
```
mcibique, line 23: We have predefined typography sizes in EBO design, we should be re-using them: https://chameleon.ebury.now.sh/?path=/docs/css-typography--basic.

What is the target font-size? e.g. if it's 16px and 24px line-height, we should use `body-text` typography. You can apply that using Tailwind by doing:
```
  @apply tw-body-text;
```

The full list of possible font-size/line-height combinations can be found here:
https://github.com/Ebury/chameleon/blob/master/tailwind.config.js#L610

src/components/ec-blog-item/ec-blog-item.vue
```diff
@@ -0,0 +1,26 @@
...
+  text-align: center;
```
mcibique, line 24: We use tailwind, so you can just replace this with `@apply tw-text-center;` You can see the full list of all Tailwind utility classes here: https://chameleon.ebury.now.sh/?path=/docs/css-tailwind--all.

src/components/ec-blog-item/ec-blog-item.spec.js
```diff
@@ -0,0 +1,28 @@
+import { mount } from '@vue/test-utils';
+import EcBlogItem from './ec-blog-item.vue';
+
+const post =
+  {
+    title: 'title',
+    shortDescription: 'shortDescription',
+    author: 'author',
+    category: 'category',
+  };
```
mcibique, line 10: Can we make these texts more random. And by that I mean set them to values that are not the same as prop names. If I accidentally change `<p>{{ author }}</p>` to `<p>author</p>` in your component, this test will still be green. Make it fail in that situation. You can just set `title` to be `My Blog Post`, `shortDescription` to `Lorem ipsum ...`, `author` to be `Random Author`, etc.

src/components/ec-blog-item/ec-blog-item.spec.js
```diff
@@ -0,0 +1,28 @@
...
+    ...mountOpts,
+  });
+}
+
+describe('EcBlogItem', () => {
+
+  describe('posts prop', () => {
+
+    it('should render post', () => {
+      const wrapper = mountBlogItem({ title: post.title, shortDescription: post.shortDescription, author: post.author, category: post.category, });
```
mcibique, line 24: CAN BE MERGED:

You can also use destructuring and just do:
```js
const wrapper = mountBlogItem({ ...post });
```

src/components/ec-blog-item/ec-blog-item.story.js
```diff
@@ -0,0 +1,41 @@
...
+    },
+    category: {
+      default: text('category', 'A new category'),
+    },
+  },
+
+  render() {
+    return (
+      <div>
+        <div>
```
mcibique, line 35: Do you need the component to be wrapped in two divs?

src/components/ec-blog-item/index.js
```diff
@@ -0,0 +1 @@
+export { default } from './ec-blog-item.vue';
```
mcibique, line 1: Don't forget to re-export it also in `main.js` so in EBO, we can easily import it in EBO:
```
import { EcBlogItem } from `@ebury/chameleon-components`;
```

src/components/ec-blog-item/ec-blog-item.story.js
matallo: Do you need the component still to be wrapped in a single div?

src/components/ec-blog-item/ec-blog-item.story.js
```diff
@@ -29,14 +40,23 @@ stories.add('all', () => ({
...
+      });
+    },
+  },
 
   render() {
     return (
-      <div>
         <div>
-          <EcBlogItem title={this.title} author={this.author} category={this.category} featuredImage={this.featuredImage} />
+          <EcBlogItem posts={this.posts} />
```
matallo, line 58: This looks like a `EcBlogList` component

src/components/ec-blog-item/ec-blog-item.vue
```diff
@@ -1,28 +1,34 @@
 <template>
-  <div class="ec-main-container ec-blog-item">
-    <div class="tw-grid tw-border-b tw-border-solid tw-border-gray-6">
+  <transition-group
```
matallo, line 2: leave `transition-group` out for now and just focus on the basic functionality

src/components/ec-blog-item/ec-blog-item.vue
```diff
@@ -1,28 +1,34 @@
...
+    class="ec-main-container ec-blog-item"
```
matallo, line 3: not sure about using these two at the same time, they are not utility classes but component classes based on utility classes, you're mixing responsibilities and if you change one here you might have unintended consequences

src/components/ec-blog-item/ec-blog-item.vue
```diff
@@ -1,28 +1,34 @@
...
+    tag="div"
+  >
+    <div
+      v-for="post of posts"
+      :key="post.id"
+      class="tw-grid tw-border-b tw-border-solid tw-border-gray-6"
+      data-test="ec-blogItem__post"
```
matallo, line 10: we use some special notation, take a look at some examples in this repo https://github.com/Ebury/chameleon/blob/master/src/components/ec-input-field/ec-input-field.vue

src/components/ec-blog-item/ec-blog-item.vue
```diff
@@ -1,28 +1,34 @@
...
+    >
       <div class="tw-col-2">
-        <img class="tw-w-96 tw-h-96 tw-rounded" :src="featuredImage">
+        <img
+          class="tw-w-96 tw-h-96 tw-rounded"
+          :src="post.featuredImage"
```
matallo, line 15: don't read from post, pass and access only `featuredImage`

https://en.wikipedia.org/wiki/Law_of_Demeter

src/components/ec-blog-item/ec-blog-item.vue
```diff
@@ -1,28 +1,34 @@
...
+        >
       </div>
-      <div class="tw-col-10" >
-        <p class="tw-mt-0 ec-blog-item__title" >
-          {{ title }}
-        </p>
-        <p>{{ author }}</p>
-        <p>{{ category }}</p>
+      <div class="tw-col-10">
+        <p class="tw-mt-0 ec-blog-item__title"> {{ post.title }} </p>
```
matallo, line 19: empty blank spaces

matallo, line 19: Do all lines need `tw-mt-0`?

matallo (changes requested):

src/components/ec-blog-item/ec-blog-item.story.js
```diff
@@ -0,0 +1,32 @@
...
+  component: EcBlogItem,
+};
+
+const Template = (args, { argTypes }) => ({
+  props: Object.keys(argTypes),
+  components: {
+    EcBlogItem,
+  },
+  template: `
+      <div class=" ec-blog-item tw-grid tw-border-b tw-border-solid tw-border-gray-6" >
```
MaruaEl, line 14: there is an extra space before ec-blog-item `class=" ec-blog-item`

src/components/ec-blog-item/ec-blog-item.vue
```diff
@@ -0,0 +1,51 @@
+<template>
+  <div
+    class="ec-blog-item tw-grid tw-border-b tw-border-solid tw-border-gray-6"
```
MaruaEl, line 3: I think you should move these tw classes in your `ec-blog-item` class. Unless I'm wrong I remember we agreed to keep all the tw stuff in a bem class. Correct me if Im wrong

src/components/ec-blog-item/ec-blog-item.story.js
```diff
@@ -0,0 +1,32 @@
...
+        <div class="tw-col-2" >
+          <img class="tw-w-96 tw-h-96 tw-rounded" :src="featuredImage" >
+        </div>
+        <div class="tw-col-10">
+          <p class="ec-blog-item__title" > {{ title }} </p>
```
jmesao, line 19: Is there any reason why we have space at first and at the end of the p tag?

src/components/ec-blog-item/ec-blog-item.story.js
matallo: Probably the utility classes could go under the BEM class below

src/components/ec-blog-item/ec-blog-item.story.js
```diff
@@ -0,0 +1,32 @@
...
+
+export default {
+  title: 'Blog Item',
+  component: EcBlogItem,
+};
+
+const Template = (args, { argTypes }) => ({
+  props: Object.keys(argTypes),
+  components: {
+    EcBlogItem,
```
matallo, line 11: Are you using the component anywhere in this file, or just replicating the markup?

src/components/ec-blog-item/ec-blog-item.vue
matallo: Same as above, actually you're replicating the markup, you should do it here and use the `<ec-blog-item>` tag above

matallo (changes requested):

src/components/ec-blog-item/ec-blog-item.spec.js
```diff
@@ -0,0 +1,18 @@
...
+  function mountEcBlogItem(props, mountOpts) {
+    return mount(EcBlogItem, {
+      propsData: {
+        ...props,
+      },
+      ...mountOpts,
+    });
+  }
+
+  it('should render post', () => {
```
matallo, line 14: we can do better in this description

src/components/ec-blog-item/ec-blog-item.vue
```diff
@@ -0,0 +1,61 @@
+<template>
+  <div
+    class="ec-blog-item"
+  >
+    <div
+      class="tw-grid tw-border-b tw-border-solid tw-border-gray-6"
```
matallo, line 6: could we have these as part of `.ec-blog-item` with `@apply`?

matallo (approved):

src/components/ec-blog-item/ec-blog-item.spec.js
```diff
@@ -0,0 +1,18 @@
...
+    const wrapper = mountEcBlogItem();
+    expect(wrapper.element).toMatchSnapshot();
```
jaimeperezpedra, line 16: We should pass the props to check in the snapshots if we are rendering those texts

src/components/ec-blog-item/ec-blog-item.vue
jaimeperezpedra: As Carlos say you can remove this `div` and put all these classes on the `.ec-blog-item` with the `@apply`

src/components/ec-blog-item/ec-blog-item.vue
```diff
@@ -0,0 +1,61 @@
...
+    >
+      <div
+        class="tw-col-2"
+      >
+        <img
+          class="tw-w-96 tw-h-96 tw-rounded"
```
jaimeperezpedra, line 12: I'll prefer to have here another class like that: `.ec-blog-item__image`

src/components/ec-blog-item/ec-blog-item.vue
```diff
@@ -0,0 +1,61 @@
...
+    </div>
+  </div>
+</template>
+
+<script>
+export default {
+  name: 'EcBlogItem',
+  props: {
+    title: {
+      type: Text,
```
jaimeperezpedra, line 40: We generally use `String` type instead of `Text`

src/components/ec-blog-item/ec-blog-item.vue
```diff
@@ -0,0 +1,61 @@
...
+        </p>
+      </div>
+    </div>
+  </div>
+</template>
+
+<script>
+export default {
+  name: 'EcBlogItem',
+  props: {
```
jaimeperezpedra, line 38: I don't know witch prop must be required but I guess at least the title should be right? I didn't found it on the ticket.

A prop required should be declared like that:

`title: { 
  type: String,
  required: true,
}`

jaimeperezpedra (approved):

src/components/ec-blog-item/ec-blog-item.spec.js
```diff
@@ -5,14 +5,33 @@ describe('EcBlogItem', () => {
   function mountEcBlogItem(props, mountOpts) {
     return mount(EcBlogItem, {
       propsData: {
+        title: 'title',
```
jaimeperezpedra, line 8: If you pass the props here, then you don't need to pass the props on the tests, and you can keep only one to test all the props instead of one by property.

Just keep one tests without passing any prop when you are calling the function. Because you are already passing those on the function itself.

jaimeperezpedra (changes requested):

jaimeperezpedra (approved):

jaimeperezpedra (approved):

mcibique:
I'm going to close this PR because it's been open for a very long time. The priorities has changed and this functionality is no longer required. We'll keep the branch in case we need come back to this component in the future. It will need a refresh after we migrated the repo to Vue 3 too.

PR #544, opened 2020-11-20
ONL-4906 add new table filter component

MaruaEl:
Container holding all the filters  and emitting to parent the selected/unselected filters 

image

Ellipsis and default tooltip:
image


matallo:
When I manually deselect all filters the "Clear filters" link still appears

src/components/ec-filter-popover/ec-filter-popover.vue
```diff
@@ -164,6 +165,10 @@ export default {
     @apply tw-border-none;
     @apply tw-text-gray-3;
     @apply tw-bg-transparent;
+    @apply tw-inline-block;
+    @apply tw-w-72;
```
MaruaEl, line 169: I'm not really sure about this... we need a fixed width to show the ellipsis but I'm puzzled - I'll speak with Itzy to ask what she thinks about it. In the main time let me know your thoughts plz

src/components/ec-table-filter/ec-table-filter.spec.js
```diff
@@ -0,0 +1,125 @@
...
+      expect(errorSpy.mock.calls[0][0]).toContain('Missing required prop: "filters"');
+    });
+  });
+
+  it('should render with pre-selected filters if the value prop is passed', () => {
+    const wrapper = mountEcTableFilter({ value, filters });
+    expect(wrapper.element).toMatchSnapshot();
+  });
+
+  it('should hide the clear filters button if there isn\'t any preselected filter', () => {
```
MaruaEl, line 80: I need to add one more test to check if the clear filters button is hidden if we manually de select the options

MaruaEl:
> When I manually deselect all filters the "Clear filters" link still appears

Great spot - thank you Carlos, Ive fixed that but I'll also add a unit test for this scenario

src/components/ec-table-filter/ec-table-filter.vue
```diff
@@ -0,0 +1,167 @@
...
+    },
+    filters: {
+      type: Array,
+      required: true,
+      default: () => ([]),
+    },
+    lessFiltersButtonText: {
+      type: Array,
+      required: false,
+      default: () => (['Less filters', 'More filters']),
```
MaruaEl, line 68: this is to be able to translate the text and pass different text based on the button status

matallo (approved):

src/components/ec-table-filter/ec-table-filter.spec.js
```diff
@@ -0,0 +1,125 @@
...
+  });
+
+  it('should emit an empty object when clear filters button is clicked', () => {
+    const wrapper = mountEcTableFilter({ value, filters });
+
+    wrapper.findByDataTest('ec-table-filter__clear-filters-button').trigger('click');
+    expect(wrapper.emitted().change).toEqual([[{}]]);
+  });
+
+  it('should emit a change event with an empty object when the usert deselect the filter', () => {
```
jmesao, line 109: user instead of usert

src/components/ec-table-filter/ec-table-filter.spec.js
```diff
@@ -0,0 +1,125 @@
...
+    const wrapper = mountEcTableFilter({ value, filters, popoverOptions: { open: true } });
+
+    wrapper.findByDataTest('ec-table-filter__filter-item-1')
+      .findByDataTest('ec-multiple-values-selection__checkbox-deselect')
+      .findByDataTest('ec-checkbox__label').trigger('click');
+
+    expect(wrapper.emitted().change).toEqual([[{}]]);
+  });
+
+  it('should emit a change event with the selected filters object when the usert select a filter', () => {
```
jmesao, line 119: user instead of usert

src/components/ec-filter-popover/ec-filter-popover.vue
jmesao: Could we set a max-width? thoughts?

src/components/ec-date-range-filter/ec-date-range-filter.vue
```diff
@@ -90,15 +90,15 @@ export default {
   computed: {
     fromValueDate: {
       get() {
-        return this.value.from;
+        return this.value?.from;
```
mcibique, line 93: Do we have a test for this?

src/components/ec-table-filter/ec-table-filter.vue
```diff
@@ -0,0 +1,167 @@
...
+  <section
+    class="ec-table-filter"
+    data-test="ec-table-filter"
+  >
+    <component
+      :is="filter.component"
+      v-for="(filter, index) in filters"
+      :key="index"
+      :value="value[filter.name]"
+      v-bind="filter"
```
mcibique, line 11: we need to remove the component from the v-bind, otherwise it renders:
```
<div data-test="ec-table-filter__filter-item-1" class="ec-sync-multiple-values-filter ec-table-filter__filter-item" name="feeType" component="[object Object]">
```

do instead
```
v-bind="{ ...filter, component: null }"
```

BTW, I noticed that the data-test from this component overrides data-test from the filter component, we should then add 
```
:data-test="$attrs['data-test'] ? `${$attrs['data-test']} ec-sync-multiple-values-filter` : 'ec-sync-multiple-values-filter'"
```
to `ec-sync-multiple-values-filter` and fix it in ec-date-range-filter too.

src/components/ec-table-filter/ec-table-filter.vue
```diff
@@ -0,0 +1,167 @@
...
+      :label="filter.label"
```
mcibique, line 12: I think label binding will be handled by `v-bind="{ ...filter }"` too. It's not necessary.

src/components/ec-table-filter/ec-table-filter.vue
```diff
@@ -0,0 +1,167 @@
...
+      :data-test="`ec-table-filter__filter-item-${index}`"
+      class="ec-table-filter__filter-item"
+      @change="onChange(filter.name, $event)"
+    />
+    <!-- TODO ONL-4893
+    <button
+      type="button"
+      data-test="ec-table-filter__less-filters-button"
+      class="ec-table-filter__less-filters-button"
+      @click="handleLessFilterClick()"
```
mcibique, line 22: `onToggleMoreFilters`

src/components/ec-table-filter/ec-table-filter.vue
```diff
@@ -0,0 +1,167 @@
...
+    >
+      {{ setLessFiltersButtonText }}
```
mcibique, line 24: `toggleMoreFiltersText`

src/components/ec-table-filter/ec-table-filter.vue
```diff
@@ -0,0 +1,167 @@
...
+      <ec-icon
+        class="ec-table-filter__less-filters-button--icon"
+        :size="16"
+        :name="setLessFiltersButtonIcon"
```
mcibique, line 28: `toggleMoreFiltersButtonIcon`. `set...` means "something will be changed to a given value". Getting a name should not be setting anything.

src/components/ec-table-filter/ec-table-filter.vue
mcibique, line 68: Can we rather have two props? It's not very intuitive for a dev to pass an array of translated texts.

src/components/ec-table-filter/ec-table-filter.vue
```diff
@@ -0,0 +1,167 @@
...
+  computed: {
+    hasFilters() {
+      return !!Object.keys(this.value).length;
+    },
+    // TODO ONL-4893
+    // setLessFiltersButtonIcon() {
+    //   return this.isLessFilterActive ? 'simple-chevron-down' : 'simple-chevron-up';
+    // },
+    // setLessFiltersButtonText() {
+    //   return this.isLessFilterActive ? this.lessFiltersButtonText[1] : this.lessFiltersButtonText[0];
```
mcibique, line 93: Same here, `[1]` and `[0]` are not readable when you compare it to `? this.lessFiltersButtonText : this.moreFiltersButtonText`

src/components/ec-table-filter/ec-table-filter.vue
```diff
@@ -0,0 +1,167 @@
...
+};
+
+</script>
+
+<style>
+.ec-table-filter {
+  @apply tw-block;
+  @apply tw-bg-gray-7;
+
+  @media screen and (min-width: theme('screens.sm')) {
```
mcibique, line 123: With tailwind, we do mobile first. so everything should be styled for smallest screen and then we should be using `@screen md {...}` for overriding it.

src/components/ec-table-filter/ec-table-filter.vue
```diff
@@ -0,0 +1,167 @@
...
+    &--icon {
+      @apply tw-flex-shrink-0;
+      @apply tw-ml-4;
+    }
+  }
+
+  &__clear-filters-button {
+    @apply tw-self-start;
+
+    @media screen and (min-width: theme('screens.lg')) {
```
mcibique, line 144: With tailwind, we do mobile first. so everything should be styled for smallest screen and then we should be using `@screen lg {...}` for overriding it.

src/components/ec-table-filter/ec-table-filter.vue
```diff
@@ -0,0 +1,167 @@
...
+      default: () => ([]),
+    },
+    lessFiltersButtonText: {
+      type: Array,
+      required: false,
+      default: () => (['Less filters', 'More filters']),
+    },
+    clearFiltersButtonText: {
+      type: String,
+      required: false,
```
mcibique, line 72: Required is false by default, so you can remove that.

src/components/ec-table-filter/ec-table-filter.vue
```diff
@@ -0,0 +1,167 @@
+<template>
+  <section
+    class="ec-table-filter"
+    data-test="ec-table-filter"
+  >
+    <component
+      :is="filter.component"
+      v-for="(filter, index) in filters"
+      :key="index"
```
mcibique, line 9: `name` of filter is always unique so I would you that as a key instead of index.

src/components/ec-table-filter/ec-table-filter.vue
```diff
@@ -0,0 +1,167 @@
...
+      :value="value[filter.name]"
+      v-bind="filter"
+      :label="filter.label"
+      :data-test="`ec-table-filter__filter-item-${index}`"
```
mcibique, line 13: Could you please add `ec-table-filter__filter-item` into this? 
```
:data-test="`ec-table-filter__filter-item ec-table-filter__filter-item-${index}`"
```

So then I can select all of them `findAllByDataTest('ec-table-filter__filter-item')` or only an individual one `findByDataTest('ec-table-filter__filter-item-1')`.

src/components/ec-date-range-filter/ec-date-range-filter.vue
MaruaEl: I didn't add one - will do now

src/components/ec-table-filter/ec-table-filter.vue
MaruaEl: 👍

mcibique:
Styling issue 1 (screen above lg):

Screenshot from 2020-11-23 12-01-01

Payment status is 36px height, with margin-bottom 4px;
Fee type is 36px height, with margin-bottom 4px;
Due date is 34px height, with margin-bottom 4px;
Clear filters is 36px height but centered

They are not aligned on the same baseline. We need to remove 4px margin when items are on the same line, that would solve 1st, 2nd and 4th item will be alligned. You don't need the self align: center for lg screens for the clear filters button after that. and we need to find why Due date is missing 2px.

mcibique:
Styling issue 2:

Screenshot from 2020-11-23 12-06-09

If 1st item is open and 2nd one is hovered over, there is no space between backgrounds. Design says there is 24px gap between the 1st filter chevron and 2nd filter label. 8px is reserved for bg paddings on both, that makes 24px - 2*8px gap between both filters that is missing.

src/components/ec-table-filter/ec-table-filter.vue
MaruaEl: Here I'm targetting the xs screens first. sm up should have white bg

src/components/ec-table-filter/ec-table-filter.story.js
```diff
@@ -0,0 +1,62 @@
+import EcTableFilter from './ec-table-filter.vue';
+import EcSyncMultipleValuesFilter from '../ec-sync-multiple-values-filter';
+import EcDateRangeFilter from '../ec-date-range-filter';
+
+export default {
+  title: 'Filters/Table Filter',
+  component: EcTableFilter,
+};
+
+export const basic = (args, { argTypes }) => ({
```
stavros-tomas, line 10: I am not strong opinionated about it and it perfectly works as is, but maybe we can do this,

`const Template = (args, { argTypes }) => ({`

and then at the bottom do 

```
export const basic = Template.bind({});

basic.args = {
  value: { paymentStatus: [{ text: 'Paid', value: 'paid' }, { text: 'Cancelled', value: 'canceled' }], feeType: [{ 
  ...
```

One of the benefits of writing stories the new way is reusable templates. Some stories might not need one but maybe they will do in the future and some others they will need one. To keep it consistent maybe is worth having all in the following way
https://storybook.js.org/docs/react/writing-stories/introduction#using-args

MaruaEl: I thought it was redundant to use if I wasn't not going to re-use it that's why Im not creating the template ur if you guy think we should always then I'll change it

src/components/ec-filter-popover/ec-filter-popover.vue
MaruaEl: spoken with Itzi - we'll set the max width to 120px. Thank you for the recommendation

src/components/ec-table-filter/ec-table-filter.story.js
MaruaEl: changed it to help with consistency and for future stories we'd even need

stavros-tomas:
Great work @MaruaEl :clap:

stavros-tomas (approved):

jmesao (approved):

src/components/ec-date-range-filter/ec-date-range-filter.spec.js
```diff
@@ -99,4 +100,18 @@ describe('ecDateRangeFilter', () => {
...
+    const wrapper = mountecDateRangeFilterAsTemplate(
+      '<ec-date-range-filter :label="label" v-model="value"/>',
+      {},
+      {
+        data() {
+          return { value: valueEmpty, label };
+        },
+      },
+    );
+    expect(wrapper.vm.fromValueDate).toEqual(undefined);
```
mcibique, line 114: `wrapper.vm` is the vm of the Component that uses the template:
```

  const Component = localVue.extend({
    components: { ecDateRangeFilter },
    template,
    ...wrapperComponentOpts,
  });	
```
The `data` of the component only has `value` and `label`, there are no `fromValueDate`, `toValueDate` on that component, so they are always undefined regardless of what EcDataRangeFilter does. You can verify that by setting the value to something that has a value, e.g.
```
data() {
          return { value: { from: '2020-11-11' }, label };
        },
```

and the test is still green instead of failing.

Also passing `valueEmpty`, which is `{ from: null, to: null }` is not testing the new code you added: `this.value?.from;` 
`?.` means that we should test what if `this.value` is undefined. But `valueEmpty` is an object which has both properties, from and to defined.

I think all we have to test is to pass null value and check the snapshot. We don't even need the v-model for it, so we can just use
```
const wrapper = mountecDateRangeFilter({ label, value: null });
```

PR #559, opened 2020-12-02
 [ONL-5378] EcWithAbortableFetch: fetchError in the state is not reset

mgascam:
We need to reset the fetchError in the state of the component when the fetch method is called.  Currently, if a request fails, the subsequent requests are considered to fail because the state is not reset. This is an issue now that we are implementing the filters and there can be more requests if the user selects a new filter

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@ebury/chameleon-components",
-  "version": "1.2.3",
+  "version": "1.2.4",
   "lockfileVersion": 1,
   "requires": true,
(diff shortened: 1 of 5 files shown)


jaimeperezpedra (approved):

stavros-tomas (approved):

matallo (approved):
Well spotted! Shall we cover this in the spec too?

matallo (approved):

stavros-tomas (approved):

MaruaEl (approved):

PR #569, opened 2020-12-04
[ONL-5336] Currency Input - Placeholder and tooltip

jafcalvente:
To meet the requirements of the story [ONL-5336](https://fxsolutions.atlassian.net/browse/ONL-5336) related to the EBO Convert & Pay functionality, it is necessary to add small features to the Currency Input component:

- Placeholder for the amount input.
- Optional tooltip when the currency selector is disabled.
- Optional tooltip when we are displaying an error message (in order to expand the information given to users).


matallo (approved):

src/components/ec-currency-input/ec-currency-input.vue
```diff
@@ -148,6 +150,9 @@ export default {
       type: Boolean,
       default: false,
     },
+    currenciesTooltipInfo: {
```
mcibique, line 153: We should avoid meaningless words in object names like data, detail, info, etc. These are code smell when naming is hard. I see that you have also currenciesTooltip as a computed prop, that's why this one is an `...Info`. It's a prop, a public interface of the component so it should have the best comprehensive name for somebody who's going to use the component, so name this one `currenciesTooltip`.

The computed prop might have name like `currenciesTooltipOptions` or `currenciesTooltipConfig`, because it's just internal computed prop and it's easy to somebody to have look in the same file what's the meaning.

2nd problem with naming here:
This tooltip is only visible if currency is disabled. I think better name would be `currenciesDisabledReason` or `disabledCurrenciesTooltip`. When I saw `currenciesTooltip` for the first time, I thought it's displayed all the time.

src/components/ec-currency-input/ec-currency-input.vue
```diff
@@ -200,6 +208,16 @@ export default {
     currenciesItems() {
       return this.currencies.map(currency => ({ text: currency, value: currency, id: currency }));
     },
+    currenciesTooltip() {
```
mcibique, line 211: Passing an empty object into the prop will show an empty tooltip, we should be rather checking that the content is present:

```
currenciesTooltipOptions() {
  let { content, placement = 'top' } = this.currenciesTooltip || {};
  if (!content) {
    return null;
  } else {
    return { content, placement };
  }
},
```

src/components/ec-currency-input/ec-currency-input.spec.js
```diff
@@ -106,6 +112,40 @@ describe('EcCurrencyInput', () => {
...
+  describe('when the disabled currency selector has a tooltip associated', () => {
+    it('should render properly', () => {
+      const tooltipMessage = 'Tooltip message';
+      const wrapper = mountCurrencyInput({
+        currenciesTooltipInfo: { content: tooltipMessage },
+        isCurrenciesDisabled: true,
+      });
+      expect(wrapper.findByDataTest('ec-currency-input__currencies').element).toMatchSnapshot();
+      expect(wrapper.findByDataTest('ec-currency-input__currencies').attributes('mocked-tooltip-content')).toBe(tooltipMessage);
+      expect(wrapper.findByDataTest('ec-currency-input__currencies').attributes('mocked-tooltip-placement')).toBe('top');
```
mcibique, line 131: You have tested both of those in `.toMatchSnapshot();`;

src/components/ec-currency-input/ec-currency-input.spec.js
```diff
@@ -106,6 +112,40 @@ describe('EcCurrencyInput', () => {
...
+      expect(wrapper.findByDataTest('ec-currency-input__amount').element).toMatchSnapshot();
+    });
+  });
+
+  describe('when the disabled currency selector has a tooltip associated', () => {
+    it('should render properly', () => {
+      const tooltipMessage = 'Tooltip message';
+      const wrapper = mountCurrencyInput({
+        currenciesTooltipInfo: { content: tooltipMessage },
+        isCurrenciesDisabled: true,
```
mcibique, line 127: We need the test the scenario when tooltip is given but the currency dropdown is not disabled.

jafcalvente: Added.

src/components/ec-currency-input/ec-currency-input.vue
jafcalvente: Much better. Updated. Thanks!

src/components/ec-currency-input/ec-currency-input.spec.js
jafcalvente: Duplicated tests removed.

stavros-tomas (approved):

PR #572, opened 2020-12-08
ONL-5278 dropdown checkbox

jaimeperezpedra:
image


src/components/ec-checkbox/ec-checkbox.vue
```diff
@@ -89,6 +98,14 @@ export default {
       default: false,
       type: Boolean,
     },
+    insideDropdown: {
```
mcibique, line 101: The prop is not tested.

MaruaEl (commented):
I think you should also update the checkbox story to add this new "Single line" label

MaruaEl (approved):

matallo (approved):

PR #593, opened 2020-12-23
add blur event in date range filed component

MaruaEl:
This pr will add a blur event on the date range component to let us use validations on blur for ebo


matallo (changes requested):
Does it come without tests?

MaruaEl:
> Does it come without tests?

Hi Carlos, I guess I'll have to add one because test coverage is failing. In the beginning thought that maybe it was more an integration test type of thing

matallo (approved):

jmesao (approved):

PR #597, opened 2021-01-06
ONL-5145 examples of new way to document stories

jaimeperezpedra:


package.json
```diff
@@ -45,16 +45,16 @@
     "@fullhuman/postcss-purgecss": "3.0.0",
     "@sinonjs/fake-timers": "6.0.1",
     "@storybook/addon-a11y": "6.1.10",
-    "@storybook/addon-actions": "6.0.28",
+    "@storybook/addon-actions": "6.1.11",
```
mcibique, line 48: Could you please update all `@storybook/*` to 6.1.11. I'm not sure why they are all out of sync. We should have only one version.

matallo (approved):

matallo:
Some small suggestions regarding the text, but it is clear enough for @Ebury/ux-design to pick up from here

Good work @jaimeperezpedra !

src/components/ec-timer/ec-timer.story.mdx
```diff
@@ -0,0 +1,22 @@
+import { Meta, Story, Canvas } from '@storybook/addon-docs/blocks';
+
+import EcTimer from './ec-timer';
+
+<Meta title="Timer" component={EcTimer} />
+
+# EcTimer
+
+This is our `EcTimer` component:
```
mcibique, line 9: We should avoid pronouns in technical documentation like our, you, we. 

Instead of we need more descriptive sentence, Saying that EcTimer is our EcTimer component is not helping anybody and sentence like that can be removed. What this component does?

"Simple countdown component with a visual indicator"

src/components/ec-timer/ec-timer.story.mdx
```diff
@@ -0,0 +1,22 @@
...
+# EcTimer
+
+This is our `EcTimer` component: 
+
+<Story name="basic">
+  <EcTimer />
+</Story>
+
+
+This is our `EcTimer` component without stopped:
```
mcibique, line 16: Component can be initialized as stopped.

src/components/ec-timer/ec-timer.story.mdx
```diff
@@ -0,0 +1,22 @@
+import { Meta, Story, Canvas } from '@storybook/addon-docs/blocks';
```
mcibique, line 1: We should definitely use ArgsTable and CustomSource in these stories. Otherwise the mdx format has less features than writing stories in plain JS using just args.

What is actually a benefit of using mdx in comparison to plain js?

src/components/ec-timer/ec-timer.story.mdx
jaimeperezpedra: You are right, but this was a simple example of how to do it, the Ux team will be the people in charge of doing this doc pages, that's why was so simple, I'll fix it anyway

src/components/ec-timer/ec-timer.story.mdx
jaimeperezpedra: I added some ArgsTable to the example.

But tbh I don't understand the CustomSource.

The people from UX will be ones to write this doc pages I think will be easier for them to write it on mdx.

package.json
jaimeperezpedra: Done

stavros-tomas (approved):
Some minor tweaks and you are there, nice cool feature @jaimeperezpedra :)

stavros-tomas (approved):

jmesao (approved):

PR #641, opened 2021-01-24
ONL-5431 fix style of the error/warning message on currency input

jaimeperezpedra:
image
image

There is a big difference between these two images, and it's for the type of the message, but maybe on the error one, the icon it's too much close to the input.

I also fix the shrink of the icon, was making smaller once the text was getting bigger.


jafcalvente (approved):
just one word: perfect!
You will probably have to update the snapshots of the visual regression tests because of the change in margin and colour.

matallo:
Bump snapshots!

matallo (approved):

jaimeperezpedra:
Yes, @jafcalvente also the font-size! Good point!

PR #656, opened 2021-01-27
[ONL-5394] Update the line-height and margin of error messages

jafcalvente:


src/components/ec-checkbox/ec-checkbox.vue
```diff
@@ -217,6 +217,9 @@ export default {
 
   &__error-text {
     @apply tw-flags-text tw-text-error;
+    @apply tw-mt-4;
+
+    line-height: 16px;
```
matallo, line 222: We should use Tailwind classes https://tailwindcss.com/docs/line-height

matallo (changes requested):

src/components/ec-checkbox/ec-checkbox.vue
jafcalvente: Yes, I was going to take that approach initially using `tw-leading-4`, but I discovered that we have explicitly disabled the `lineHeight` related classes in the tailwind configuration (`tailwind.config.js`) because in almost all cases the necessary line-height definition is defined by our typographies. This is a specific case that does not correspond to any of the ones we have, and as other places in the project are also defining the line-height with native css, I finally decided to make it explicit here as well

matallo: In this case I'd move forward, but seems like something we'll need to tackle soon

matallo (approved):

jaimeperezpedra (approved):

mgascam (approved):

PR #713, opened 2021-02-23
Onl 5574 countdown

jaimeperezpedra:


matallo (commented):
You're creating a new countdown instance every time you start the timer component, wouldn't that lead to leaks?

jaimeperezpedra:
> You're creating a new countdown instance every time you start the timer component, wouldn't that lead to leaks?

I see good caught! I'll fix it

src/utils/countdown.js
```diff
@@ -0,0 +1,30 @@
+const EventEmitter = require('events');
```
stavros-tomas, line 1: Not sure, can you use import instead or require?
For example import { EventEmitter } from 'events';

jaimeperezpedra: Yeah! fixed

mcibique: Please investigate how this behave in the browser.It's nodejs native module. How much it costs. How big is the file. Wouldn't solutions mentioned in Vue3 migration guide a better option: https://v3.vuejs.org/guide/migration/events-api.html#migration-strategy.

src/utils/countdown.js
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  get timeDifference() {
+    return this.currentTime - this.startTime;
+  }
+
+  reduceSecondsLeft() {
+    this.currentTime = Math.ceil(Date.now() / 1000);
+    this.secondsLeft = this.seconds - this.timeDifference;
+    this.emit('time-updated');
```
mcibique, line 24: I'd emit the number of seconds left in this event.

src/utils/countdown.spec.js
```diff
@@ -0,0 +1,44 @@
...
+    if (clock) {
+      clock.uninstall();
+    }
+  });
+  it('should finish the countdown', () => {
+    const countdown = new Countdown();
+    countdown.start(20);
+    clock.tick(20000);
+
+    expect(countdown.secondsLeft).toBe(0);
```
jmesao, line 20: Could we check also that the should clear interval?

jaimeperezpedra: I tested what the stop function is called, I was trying to test the clear interval but I couldn't.

Tell me is that's enough.

src/utils/countdown.spec.js
```diff
@@ -0,0 +1,46 @@
...
+    expect(countdown.secondsLeft).toBe(10);
+    clock.tick(10000);
+    expect(countdown.secondsLeft).toBe(0);
+  });
+
+  it('should emit the events', () => {
+    const timeExpiredMock = jest.fn();
+    const timeUpdatedMock = jest.fn();
+    const countdown = new Countdown();
+    countdown.start(20);
```
mcibique, line 38: I would attach the mocks before starting the countdown, just in case.

src/utils/countdown.spec.js
```diff
@@ -0,0 +1,46 @@
...
+    countdown.emitter.on('time-expired', timeExpiredMock);
+    countdown.emitter.on('time-updated', timeUpdatedMock);
+    clock.tick(20000);
+
+    expect(timeExpiredMock).toBeCalledTimes(1);
+    expect(timeUpdatedMock).toBeCalledTimes(20);
```
mcibique, line 44: Also check here that there are no timers left to be executed.
```
expect(clock.countTimers()).toBe(0);
```

src/utils/countdown.spec.js
mcibique: You can check the number of timers left to be executed:

```
expect(clock.countTimers()).toBe(0);
```

src/components/ec-timer/ec-timer.vue
```diff
@@ -93,43 +96,38 @@ export default {
...
-      this.timerInterval = setInterval(this.reduceSecondsLeft, 1000);
-    },
-    reduceSecondsLeft() {
-      this.currentTime = Math.ceil(Date.now() / 1000);
-      this.secondsLeft = this.seconds - this.timeDifference;
-      if (this.timeDifference >= this.seconds) {
-        clearInterval(this.timerInterval);
+    startCountdown() {
+      this.countdown.start(this.seconds);
+      this.countdown.emitter.on('time-updated', (secondsLeft) => {
```
mcibique, line 120: These events are never being deregistered, so if I call `this.startCountdown();` twice, all events are fired two times.

I think you had it right when you was initialising countdown in this function, it was preventing this behaviour to happen. If you are afraid of memory leaks, then stop previous countdown first when starting the new one:
```
startCountdown() {
  if (this.countdown) {
    this.countdown.stop();
  }
  this.countdown = new Countdown();
  this.countdown.start(this.seconds);
}
```

src/components/ec-timer/ec-timer.vue
```diff
@@ -93,43 +96,38 @@ export default {
...
+        this.secondsLeft = secondsLeft;
+      });
+      this.countdown.emitter.on('time-expired', () => {
```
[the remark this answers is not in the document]
jaimeperezpedra: I was thinking about this, but I didn't know how to do it! thanks!!

src/components/ec-timer/ec-timer.vue
jaimeperezpedra: Cool! done!

src/components/ec-timer/ec-timer.vue
```diff
@@ -93,43 +96,44 @@ export default {
...
     isRunning: {
       immediate: true,
       handler(value) {
         if (value) {
-          this.countdown();
+          this.startCountdown();
         } else {
-          clearInterval(this.timerInterval);
+          if (this.countdown) {
+            this.countdown.stop();
```
jafcalvente, line 108: Including `this.countdown = null;` also here we avoid to invoke the `this.countdown.stop();` again the next time `isRunning` is set to `true`

jafcalvente (approved):

src/components/ec-timer/ec-timer.vue
jaimeperezpedra: Done!

jaimeperezpedra:
I a question about the test.

In the next image you will see that we are getting some errors from the test because this props are missing, I tested and those tests that are failing are the ones that should test and we have the method `withMockedConsole` to prevent those errors on the console.

Any idea why this is happening? 

image

stavros-tomas (approved):

matallo (approved):

PR #804, opened 2021-04-26
ONL-5785:fix:do not destroy metroline item if not visible

stavros-tomas:
Do not destroy metroline item when not visible.


jaimeperezpedra (approved):

mcibique:
This change doesn't make sense TBH. You have 12 v-ifs in the metroline-item and you decided to change one of them to v-show. Can the problem you are trying to solve happen in the footer or header as well? or in other places? I don't like that the main is somehow different. It will make the code confusing.

PR #826, opened 2021-06-02
ONL-5889:chore:change alignment of tooltip when placed on the right

stavros-tomas:
just changed the text alignment of tooltip when placed on the right to be left


jeger-at (approved):

jafcalvente (commented):
Is there any snapshot that we should update?

mcibique (approved):

PR #838, opened 2021-06-15
[ONL-5858] Adding new icons and new statuses to EcCard component

jafcalvente:
Adding new icons and adding disabled and has-narrow-padding statuses to the `EcCard` component.


src/styles/components/ec-card/ec-card.css
```diff
@@ -33,7 +33,17 @@
     }
   }
 
-  @screen sm {
+  &--disabled {
```
stavros-tomas, line 36: Hey Jose, 
Couple of questions:
- Do you think is worth changing to "--is-disabled" to make it consistent with the above "--is-interactive"?
- Can the interactive version become disabled? If yes do you think we should stop the card from "jumping/transitioning" when disabled?
- Shall we reflect the disabled status for each card. For example in one column the enabled version for basic/interactive/card with narrow padding and in another next to it the disabled cards?

Not strong opinionated about any of the above. Feel free to merge if you feel they are not important.

stavros-tomas (approved):

stavros-tomas (approved):

src/styles/components/ec-card/ec-card.css
jafcalvente: At first sight, we are using both in the library, `--disabled` and `--is-disabled`, but since we have here `--is-interactive` as you say, it makes sense rename it to `--is-disabled`, yes.
About reflect the disabled status for each card I think it's not necessary, since it is not actually a real Vue component. It's only a bunch of classes and depends on us using them as we need.
This also applies to your second point, about what happens if an interactive card is disabled, and the point is that this scenario can be handle by making both classes exclusives:

```
      <a
        href="#"
        class="ec-card ec-card--has-narrow-padding"
        :class="{
          'ec-card--is-disabled': isDisabled,
          'ec-card--is-interactive': !isDisabled,
        }"
        @click.prevent.stop="$emit('click')"
      >
```

stavros-tomas: Great, thanks Jose

stavros-tomas (approved):

PR #874, opened 2021-08-04
ONL-6085: feat: Added warning bottom note to input field.

mcibique:
Bottom note for input-field using the same approach as in currency-input.


stavros-tomas (approved):
Looks good!

jeger-at (approved):

PR #879, opened 2021-08-18
ONL-5954: npm 7 update and dependabot fixes

mcibique:
**NPM7:**

- Updated package-lock to support NPM 7. However, NPM 7 completely reworked the peer dependencies and it now installs them automatically and also throws an error if peer dependency cannot be installed. This won't be resolved easily because we have to wait until all maintainers update everything. Because of that, I have disabled the new behaviour and set `legacy-peer-deps=true` so the NPM 7 will use algorithm from NPM 6. Some of the examples which prevents us to use peer dependency algorithm from v7:
1. [VUE CLI eslint service](https://github.com/vuejs/vue-cli/blob/v4.5.13/packages/@vue/cli-plugin-eslint/package.json#L35) has a peer dependency on eslint@6 while chameleon uses 7. Fixed in next beta.
2. [cypress-image-snapshot](https://github.com/jaredpalmer/cypress-image-snapshot/blob/v4.0.1/package.json#L49) has a peer dependency on cypress@4, chameleon needs 5. No fix so far.
3. [storybook-vue-router](https://github.com/gvaldambrini/storybook-router/pull/66) has a peer dependency on Storybook 5, we need 6. PR to fix that opened in April.
4. [stylelint-config-standard-scss](https://github.com/Moeriki/stylelint-config-standard-scss/pull/2/) has a peer dependency on stylelint 10, while we need 13. PR opened in June.
5. Indirect dependencies in [@ascendancyy/vue-cli-plugin-stylelint](https://github.com/ascendancyy/vue-cli-plugin-stylelint/blob/master/package.json#L34-L36) requires stylelint 9 and 8.

^^ These also create a concern if this algorithm will not be blocking us from updating dependencies in the future, so it may be possible that we will stay on legacy (v6) forever. E.g. if eslint releases version 8, we need to wait until every dependency in the tree will release version that supports version 8 as a peer dependency, even if most of them are not affected by changes in 8.

**DEPENDABOT:**

- Reviewed dependabot settings that was generated when migrating from preview to final dependabot and resolved lots of issues. Prevented dependabot from updating further:
1. cypress (we can do that in separate ticket)
2. svgo related packages (they were updated for webpack 5) - we will be removing svg sprites anyway
3. postcss7 - we have another ticket to upgrade chameleon to postcss 8
4. clipboard-copy - version 4 removed IE11 support and the code has been rewritten from scratch. We need to update and retest everything. I'll create a ticket for it.

**UPDATES:**

- Updated prod dependencies:
1. mitt - breaking change doesn't affect us
2. focus-trap

- Updated storybook dependencies and resolved warnings in storybook about `loose` mode of babel polyfills. Removed --no-dll flag because it was removed in 6.1. Added @storybook/addon-postcss because of the deprecation in postcss-loader.

- Resolved breaking change in `toBeVisible()` function from `@testing-library/jest-dom`.

- Fixed git hooks linting entire folder instead of staged files.


src/components/ec-menu/ec-menu.spec.js
```diff
@@ -79,7 +79,7 @@ describe('EcMenu', () => {
     expect(wrapper.element).toMatchSnapshot();
 
     wrapper.findAllByDataTest('ec-navigation-link__text').wrappers.forEach((textWrapper) => {
-      expect(textWrapper.element).toBeVisible();
+      expect(textWrapper.isVisible()).toBe(true);
```
stavros-tomas, line 82: Does this change affect EBO?

stavros-tomas (approved):

src/components/ec-menu/ec-menu.spec.js
mcibique: yes, I fixed it there too.

MaruaEl (approved):

PR #880, opened 2021-08-19
ONL-6095 textarea component

stavros-tomas:
Textarea component, follows similar logic with EcInput

*non resizable
*can amend number of rows with prop
*you cannot amend columns (full width always, to be confirmed)


src/components/ec-textarea/ec-textarea.spec.js
```diff
@@ -0,0 +1,158 @@
...
+    );
+
+    expect(wrapper.findByDataTest('ec-textarea__textarea').element.value).toBe('');
+    await wrapper.setData({ text: 'Value Lorem ipsum' });
+    expect(wrapper.findByDataTest('ec-textarea__textarea').element.value).toBe('Value Lorem ipsum');
+  });
+
+  it('renders with a placeholder', () => {
+    const wrapper = mountTextarea({}, { attrs: { placeholder: 'dfdffdfdfdf' } });
+    expect(defaultWrapper.element).toMatchDiffSnapshot(wrapper.element);
```
mcibique, line 55: Please use regular `toMatchSnapshot()` everywhere. `toMatchDiffSnapshot` should be used when you initialize a component in the test with one state and then you change the state during the test. Comparing a rendered component to a global template seems odd and if the default behaviour of the component change, all of these diffs will start throwing diff errors. Also I find reading the errors thrown by toMatchDiffSnapshot very hard.

I mean reading something like this means I need to know what does the default state renders, so I can apply the diff on top of the default state in my mind:
```
exports['EcTextarea :props should render with a bottom note 1'] = `
"Snapshot Diff:
- First value
+ Second value

@@ -9,7 +9,14 @@
      data-test=\\"ec-textarea__textarea\\"
      id=\\"ec-textarea-1\\"
      rows=\\"4\\"
    />

-   <!---->
+   <div
+     class=\\"ec-textarea__bottom-note\\"
+     data-test=\\"ec-textarea__bottom-note\\"
+   >
+     
+     Bottom note lorem ipsum
+   
+   </div>
  </div>"
`;
```
I'm lacking lots of context when reading it.

stavros-tomas: Hey @mcibique I see your point, will amend to what we do normally. It was an attempt to make snapshots cleaner to see if you like it :)

mcibique (approved):

PR #893, opened 2021-09-02
ONL-5876:feat:add completeMetroline method

stavros-tomas:
Added a completeMetroline method.

We can rename to goToLast if you wish.


src/components/ec-metroline/ec-metroline.vue
```diff
@@ -54,6 +54,10 @@ export default {
           this.metroline.isCompleted = false;
           this.$emit('change', this.metroline.activeItemId);
         },
+        completeMetroline: () => {
+          const lastIndex = this.metrolineItemIds.length;
+          this.metroline.goToNext(lastIndex);
```
mcibique, line 59: You are mixing apples and oranges here. `goToNext` expects an ID of the metroline item, you are passing an index of a metrolineItemId which doesn't exist. This will introduce bugs if you use v-ifs on metroline-items. Consider this:

```
<ec-metroline-item :id="1" v-if="isVisible" />
<ec-metroline-item :id="2" />
<ec-metroline-item :id="3" />
<ec-metroline-item :id="4" />
```
At the beginning, 4 items will register itself, so the `this.metrolineItemIds.length` is 4, the ids are [1, 2, 3, 4]. Then isVisible is set to false. Now `this.metrolineItemIds.length` is 3, the ids are [2, 3, 4] and you tell the metroline to goToNext(3), which is the 2nd item. This will not complete the metroline.

You should be calling in this function what's already in goToNext code:
```
this.metroline.isCompleted = true;
this.$emit('complete');
```

and you should call this function in the goToNext():
```
else {
            this.metroline.complete();
          }
```

I'm not sure about the value in `activeItemId`. Should we set it to null when completing the metroline? Can metroline have activeItemId and also be complete? Does it make any difference?

mcibique (changes requested):

src/components/ec-metroline/ec-metroline.vue
stavros-tomas: Badge wise, we have hardcode a checked icon 
```<span
          v-if="!isCompleted && id"
          data-test="ec-metroline-item__index"
        >
          {{ badgeText || id.toString() }}
        </span>

        <ec-icon
          v-else
          data-test="ec-metroline-item__completed-icon"
          name="simple-check"
          :size="14"
        />
        
        then metroline wise, once we complete it we cannot really do something else...so I guess, doesn't matter?

mcibique (approved):

PR #895, opened 2021-09-03
ONL-6126:feat:header cta when item not complete

stavros-tomas:
Removed completely the logic that was controlling the visibility of `header-cta` `header-cta-complete` and `sub-heading` from now on this logic should be done on EBO side.


mcibique (commented):
IMHO, these changes only worsen something that doesn't make sense. Let me summarise the state of heading in metroline-item I see:

1. metroline-item has a heading slot which is visible all time, cool.
2. metroline-item has a sub-heading slot which is only visible when item is complete. why? why heading is always there but sub-heading visibility is limited? whose requirement is that? EBO's or chameleon? I guess it's EBO and a logic like that has no place here. We should have no constrains and display the sub-heading when the slot is passed.
3. metroline-item has a header-cta visible when the item is completed and on top of that there are 2 slots in it that are used when the item isReadonly or not, one of them called header-cta-complete! I'm sorry, what? so header-cta slot is used when item isCompleted and not isReadOnly and headr-cta-complete slot when item isCompleted and isReadonly? are we going to create more slots for every flag in the future? this needs to stop, it doesn't make sense when you read the code.

Instead of adding more flags a make the logic even more convoluted allow the CTA all the time and let the application to decide whether to display it.

Join header-cta and header-cta-complete together, remove constraints isCompleted/isReadonly, and let the application to decide whether the CTA should be there or not. We're passing all information about the state of the item via `v-bind`, use that in the app.

jmesao:
I suggested to Stavros to add _hasHeaderCTAVWhenNotComplete_ but reviewing the _ec-metroline-item_ in deep is true we are adding a lot of flags and new props doing unmaintainable and complex this component. Aout the isReadOnly or not, the slots are the same but different name, I also don't understand well that part, so I agree.

stavros-tomas:
@mcibique @jmesao in retrospective, indeed, we have had this logic in EBO and on chameleon just have the logic of "if <slot> then render it".  Will try now to experiment on that and see how much the flow on EBO will break....

mcibique (approved):

PR #972, opened 2021-11-08
ONL-6216: docs: Better documentation about git tags and how to not loose them in every release

mcibique:
This adds clarification when to create a tag.

Normally, `npm version` does that for us but we then loose the tag after squashing the commits when merging to master. There is a option in `npm version` called `no-git-tag-version` which should prevent creating a tag when bumping up the version, unfortunately, it also prevents the commit to be created -> bug.


jmesao (approved):
Thanks for managing how to proceed with it, all clear now!

PR #1062, opened 2022-02-21
ONL-6337:datepicker

stavros-tomas:
Datepicker Component

image

- We use the input component + Flatpickr
- Has most of the input component props plus some extra like `areWeekendsDisabled` and `disabledDates`
- We can further configure the component via the `options` prop

```
  options: {
    minDate: '2022-02-10',
    maxDate: '2022-03-10',
  },
```

As a sidenote In this PR 

- I had to destroy many times so `enableAutoDestory` was necessary. At the end @mcibique proposed this to be placed in the globalscope which affected some other files that we were using this as well. The changes can be found in this commit [[ONL-6337:refactor:make autodestroy global](https://github.com/Ebury/chameleon/pull/1062/commits/7f64b7409220cb8f064b419e28a450f31f9f329e)]
- added temporary an exception to don't purge the css of flatpicker until we start work on our own css


jafcalvente:
I'm clicking on the input field of the Preview link, but the datepicker is not being shown (at least, not to me)

stavros-tomas:
@jafcalvente neither to me, on my local it does though, will investigate

jafcalvente:
Great. Now the preview is working fine with me. Thanks. A couple of things:

- The datepicker selector is displayed if I click on the input field area, except on the bit where the icon is. If I click there nothing happens.
- The "Disable 2022-02-22" button is not disabling that specific day, but the next day (23).
- With the "Disable weekends" button we are toggling the weekends, whereas with "Disable 2022-02-22" we are disabling a date and it can not be enabled again. It's a minor thing. Not really need to be changed, but I only wanted to highlight this different behavior, even when both buttons have a similar text.

src/components/ec-input-field/ec-input-field.vue
```diff
@@ -206,6 +206,9 @@ export default {
     focus() {
       this.$refs.input.focus();
     },
+    getInputRef() {
```
stavros-tomas, line 209: This is not a leftover, we use it above in the datepicker to mount in a unique reference the flatpickr

src/components/ec-datepicker/ec-datepicker.vue
```diff
@@ -0,0 +1,200 @@
...
+        }
+      },
+    },
+    areWeekendsDisabled() {
+      /* istanbul ignore next */
+      if (this.flatpickrInstance) {
+        this.flatpickrInstance.redraw();
+      }
+    },
+    value(newValue) {
```
jafcalvente, line 112: we could include this into the model setter and save this watcher, right?
```
computed: {
  datepickerModel: {
    ...
    set(value) {
      this.flatpickrInstance.setDate(value);
      this.$emit('value-change', value);
    },
  },
},
```

src/components/ec-input-field/ec-input-field.vue
stavros-tomas: @jafcalvente Here we use this unique ref. We can solve the issue of not opening the calendar when clicking on the icon by instead targetting the input to target the input wrapper. This will make it open when we click on the icon as well.

src/components/ec-datepicker/ec-datepicker.vue
```diff
@@ -0,0 +1,200 @@
...
+        if (newValue) {
+          this.disabledDatesMap = new Map(Object.entries(newValue));
+        }
+        if (this.flatpickrInstance) {
+          this.flatpickrInstance.redraw();
+        }
+      },
+    },
+    areWeekendsDisabled() {
+      /* istanbul ignore next */
```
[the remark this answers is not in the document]
stavros-tomas: Sure, in this commit I added a few ignores for the same reason always https://github.com/Ebury/chameleon/pull/1062/commits/56438ace18f00ec8411bdead2601826a1b1bd160 

Jest was complaining about coverage. Although It is possible that the component might get destroyed before it even gets mounted this is rare and we cannot simulate on tests, hence the "instanbul ignore". I checked with @mcibique and he seems to be ok with it.

src/main.js
```diff
@@ -16,6 +16,7 @@ export { default as EcAmountInput } from './components/ec-amount-input';
 export { default as EcBtn } from './components/ec-btn';
 export { default as EcBtnDropdown } from './components/ec-btn-dropdown';
 export { default as EcButtonGroup } from './components/ec-button-group';
+export { default as EcDatepicker } from './components/ec-datepicker';
```
jafcalvente, line 19: we should move this line a little further down to respect the alphabetical order

jmesao:
> Great. Now the preview is working fine with me. Thanks. A couple of things:
> 
> * The datepicker selector is displayed if I click on the input field area, except on the bit where the icon is. If I click there nothing happens.
> * The "Disable 2022-02-22" button is not disabling that specific day, but the next day (23).
> * With the "Disable weekends" button we are toggling the weekends, whereas with "Disable 2022-02-22" we are disabling a date and it can not be enabled again. It's a minor thing. Not really need to be changed, but I only wanted to highlight this different behavior, even when both buttons have a similar text.

jmesao:
@stavros-tomas I'm testing the component and by default the date selected when rendering is 2022-02-06 and when we open the datepicker and close doing blur, the default value date disappear, it looks like for any reason, the value date is not being persistent...

src/components/ec-datepicker/ec-datepicker.story.js
```diff
@@ -0,0 +1,67 @@
...
+
+export const basic = (args, { argTypes }) => ({
+  components: { EcDatepicker },
+  props: Object.keys(argTypes),
+  data() {
+    return {
+      model: '2022-02-06',
+      disableWeekends: false,
+      notAvailableDates: {
+        '2022-02-21': 'Bank holiday',
```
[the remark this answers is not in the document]
stavros-tomas: @jafcalvente @jmesao It seems there is some difference between the timezones that's why I see it correctly and you don't

London time --> dayElement.dateObj--> Sat Mar 12 2022 00:00:00 GMT+0000 (Greenwich Mean Time) isoDateTime--> 2022-03-12T00:00:00.000Z

Spanish time --> dayElement.dateObj--> Sat Mar 12 2022 00:00:00 GMT+0100 (Central European Standard Time) isoDateTime--> 2022-03-11T23:00:00.000Z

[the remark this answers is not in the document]
stavros-tomas: The conversation is done correctly. 
We pick up midnight times. What is 11 of march in this case at 23:00 is midnight the 12th for UK. Not sure if at this stage we sould set time to noon or something.

jmesao (approved):

jeger-at (approved):

src/components/ec-datepicker/ec-datepicker.spec.js
```diff
@@ -0,0 +1,387 @@
...
+
+    it('should have disabled a specific date', () => {
+      const { calendarWrapper } = mountDatepicker({
+        disabledDates: {
+          '2022-02-22': 'Bank holiday',
+          '2022-02-23': '',
+        },
+      });
+
+      expect(calendarWrapper.findByDataTest('ec-datepicker__calendar-day--2022-02-22').classes('flatpickr-disabled')).toBeTruthy();
```
mcibique, line 127: `classes()` return boolean so we can safely test `.toBe(true)` instead.

src/components/ec-datepicker/ec-datepicker.spec.js
```diff
@@ -0,0 +1,387 @@
...
+
+      await inputWrapper.setProps({
+        options: {
+          minDate: '2022-02-22',
+        },
+      });
+
+      inputWrapper.vm.$nextTick();
+
+      expect(calendarWrapper.findByDataTest('ec-datepicker__calendar-day--2022-02-21').classes('flatpickr-disabled')).toBeTruthy();
```
mcibique, line 205: can we also test the other boundary: 2022-02-22 should stay enabled.

src/components/ec-datepicker/ec-datepicker.spec.js
```diff
@@ -0,0 +1,387 @@
...
+
+    it('should not update with custom hooks after initialization', async () => {
+      const { inputWrapper, calendarWrapper } = mountDatepicker({
+        options: {},
+      });
+
+      await inputWrapper.setProps({
+        options: {
+          onChange: () => {
+            inputWrapper.vm.$emit('changed');
```
mcibique, line 216: We shouldn't be using vm for this kind of tests. You can set `let onChangeSpy = jest.fn()`, pass it to `onChange: onChangeSpy` and then assert that the spy didn't get called.

src/components/ec-datepicker/ec-datepicker.spec.js
```diff
@@ -0,0 +1,387 @@
...
+      expect(calendarWrapper.findByDataTest('ec-datepicker__calendar-day--2022-02-22').classes('flatpickr-disabled')).toBeTruthy();
+      expect(calendarWrapper.findByDataTest('ec-datepicker__calendar-day--2022-02-22').attributes('title')).toBe('Bank holiday');
+
+      await inputWrapper.setProps({
+        disabledDates: {
+          '2022-02-23': 'Bank holiday',
+        },
+      });
+
+      inputWrapper.vm.$nextTick();
```
mcibique, line 264: I was going to comment why nextTick is needed in all tests? Especially because the $nextTick call must be awaited, e.g.
```
await inputWrapper.vm.$nextTick();
```
so I have concerns if it actually does anything.

src/components/ec-datepicker/ec-datepicker.spec.js
```diff
@@ -0,0 +1,387 @@
...
+
+      expect(calendarWrapper.findByDataTest('ec-datepicker__calendar-day--2022-02-22').classes('flatpickr-disabled')).toBeFalsy();
```
mcibique, line 266: `toBe(false)`.

src/components/ec-datepicker/ec-datepicker.spec.js
```diff
@@ -0,0 +1,387 @@
...
+        {
+          data() {
+            return {
+              model: '',
+            };
+          },
+        },
+      );
+
+      inputWrapper.findByDataTest('ec-input-field__input').setValue('2022-02-23');
```
mcibique, line 304: `setValue()` should be awaited.

src/components/ec-datepicker/ec-datepicker.spec.js
```diff
@@ -0,0 +1,387 @@
...
+        {
+          data() {
+            return {
+              model: '',
+            };
+          },
+        },
+      );
+
+      calendarWrapper
```
mcibique, line 322: click() should be awaited.

src/components/ec-datepicker/ec-datepicker.vue
```diff
@@ -0,0 +1,205 @@
...
+  },
+  beforeDestroy() {
+    /* istanbul ignore next */
+    if (this.flatpickrInstance) {
+      this.flatpickrInstance.destroy();
+    }
+  },
+  methods: {
+    openCalendar() {
+      this.flatpickrInstance.open();
```
jafcalvente, line 140: It still has different behavior depending on which part of the component we click on to open the calendar. When we click on the input field part and the calendar is already open, it simply remains open. However, when we click on the icon part, the calendar will re-run the opening animation each time.
I understand that the problem is not solved by checking whether the calendar is being displayed, since the calendar is closed when we click on the icon part.
Anyway, I can't really think of a simple way to solve it (without having to touch the input field component), so I guess I could accept this behavior if there is no simple solution for this.

stavros-tomas: Hi Jose,  I am aware of this behaviour but I felt this solution is good enough. I wouldn't like to overcomplicate things since it does what we want now, it opens the datepicker when you click on the icon. If you are not strong opiniated on this I would like to keep it as is.

src/components/ec-datepicker/ec-datepicker.story.js
```diff
@@ -0,0 +1,68 @@
...
+  `,
+});
+
+basic.args = {
+  label: 'Datepicker',
+  placeholder: 'Choose a date...',
+  options: {
+    minDate: '2022-02-10',
+    maxDate: '2022-03-10',
+    defaultDate: '2022-02-24',
```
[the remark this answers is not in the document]
stavros-tomas: Hi @mcibique it is used. If you set the default value via v-model it does not really set it. Initialization must be done via options `defaultDate`

src/components/ec-datepicker/ec-datepicker.spec.js
stavros-tomas: Sure will amend. @mcibique Just out of curiousity when we should choose one over the other?

src/components/ec-datepicker/ec-datepicker.spec.js
stavros-tomas: Hey thank you for feedback, can you please explain why is bad to use vm for this tests for future reference? Thank you in advance :)

src/components/ec-datepicker/ec-datepicker.spec.js
stavros-tomas: As a general rule when we click to await or only now because potentially flatpickr can take some time to calculate? Shall we do the same for example when we check the click action that toggles the visibility of a <div> in one of our own components?

jafcalvente (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@ebury/chameleon-components",
-  "version": "1.9.4",
+  "version": "1.9.6",
```
stavros-tomas, line 3: This change is correct. Npm had version 1.9.5 whilst here we had to 1.9.4. I had to jump 2 version up to catch up with npm

PR #1082, opened 2022-03-03
ONL-6339:feat: Add i18n to datepicker and use Date instance

jmesao:
Add internationalization and localization to the date picker chameleon component and use Date instance instead of strings.


src/components/ec-datepicker/ec-datepicker.story.js
```diff
@@ -53,6 +77,8 @@ export const basic = (args, { argTypes }) => ({
 
       <button @click="toggleWeekendAvailability">Toggle weekend availability</button>
       <button @click="changeDisabledDates('2022-02-22')">Disable 2022-02-22 UTC</button>
+      <button @click="model = new Date('2022-02-25')">Select 2022-02-25 date</button>
+      <button @click="dateFormatBasedOnLocale = 'd/m/Y'">Change date format to 'd/m/Y'</button>
```
mcibique, line 81: We could just create a dropdown control with multiple choises like `null`, `'d/m/Y'`, `'d.m.Y'` and `'d-m-Y'`.

src/components/ec-datepicker/ec-datepicker.story.js
```diff
@@ -53,6 +77,8 @@ export const basic = (args, { argTypes }) => ({
 
       <button @click="toggleWeekendAvailability">Toggle weekend availability</button>
       <button @click="changeDisabledDates('2022-02-22')">Disable 2022-02-22 UTC</button>
+      <button @click="model = new Date('2022-02-25')">Select 2022-02-25 date</button>
```
mcibique, line 80: The flatpickr uses user's locale so the new model should be using the locale: `model = new Date(2022, 1, 25)`

src/components/ec-datepicker/ec-datepicker.vue
```diff
@@ -147,19 +166,40 @@ export default {
     }
   },
   methods: {
+    getListeners() {
+      const listeners = { ...this.$listeners };
+      delete listeners['value-change'];
+      delete listeners.open;
```
jafcalvente, line 172: Out of curiosity, why did we need to remove all these listeners?

jmesao: $listeners property can forward all event listeners on the component to a specific child element and they were causing conflicts in our ec-datepicker because we already have these events coming from the flatpickr instance, so the events we have defined in the options (mergeWithDefaultOptions method) which are ready, change, open and close we removed it. If for example, we do blur in our datepicker that event will be forwarded to our ec-input-field, we did not remove it, but all the events we have defined in our datepicker options should be removed from $listeners, the onChange was causing issues because first, we do this.$emit('value-change', d); passing directly a Date instance but we also have $listeners and the value-change will be executed again with the value of the input which is formattedValue (string type) setting the v-model to a string instead of having a Date instance, so we don't want to listen the onChange event coming from the ec-input-field component.

src/components/ec-datepicker/ec-datepicker.story.js
jmesao: Done, thanks!

jafcalvente (approved):

stavros-tomas (approved):

jeger-at (approved):

PR #1101, opened 2022-03-11
ONL-5035 Adding of indeterminate to the checkbox

derrick-g:
Added indeterminate to the checkbox, where the checkbox will have a horizontal line when one of the child check-boxes is selected, this will then become a check mark only if all child items are selected.


src/components/ec-checkbox/ec-checkbox.vue
```diff
@@ -129,6 +139,22 @@ export default {
       },
     },
   },
+  watch: {
+    indeterminate(newValue) {
+      this.updateIndeterminate(newValue);
+    },
+  },
+  mounted() {
+    this.updateIndeterminate(this.indeterminate);
```
jafcalvente, line 148: This is being called here in order to check if we have to mark as indeterminate the checkbox as soon as the component is mounted, right? Then maybe we could try to use the `immediate` option in the watch function instead:
```
watch: {
  indeterminate: {
    immediate: true,
    handler(newValue) {
      this.updateIndeterminate(newValue);
    },
  },
},
```

jafcalvente (approved):

src/components/ec-checkbox/ec-checkbox.vue
```diff
@@ -30,14 +30,20 @@
           'ec-checkbox__check-icon-wrapper--focused': inputIsFocused,
           'ec-checkbox__check-icon-wrapper--checked': inputModel,
           'ec-checkbox__check-icon-wrapper--checked-and-focused': inputModel && inputIsFocused,
+          'ec-checkbox__check-icon-wrapper--indeterminate': indeterminate,
+          'ec-checkbox__check-icon-wrapper--indeterminate-and-focused': indeterminate && inputIsFocused,
```
jeger-at, line 34: We also need a indeterminate-and-disabled style which should grey out the indeterminate state when disabled

src/components/ec-checkbox/ec-checkbox.vue
```diff
@@ -30,14 +30,20 @@
...
           'ec-checkbox__check-icon-wrapper--error': isInvalid && !inputModel,
           'ec-checkbox__check-icon-wrapper--disabled': disabled,
           'ec-checkbox__check-icon-wrapper--checked-and-disabled': disabled && inputModel,
         }"
         @click="$refs.checkboxInput.click()"
       >
+        <span
+          v-if="indeterminate"
+          class="ec-checkbox__indeterminate-icon"
```
stavros-tomas, line 43: I have my thoughts if this should be another icon and swap between icons instead. If no one else mention this you can go ahead and ignore :)

src/components/ec-checkbox/ec-checkbox.vue
mcibique: `updateIndeterminate` does use `$refs` which are only available after component has been mounted. `immediate` watchers are triggered before mounted, so the input's ref is still null and it won't get set.

src/components/ec-checkbox/ec-checkbox.vue
derrick-g: Good thinking, I will include those as well.

src/components/ec-checkbox/ec-checkbox.vue
mcibique: The icon for indeterminate is just a white line 10px by 2px. I think it is quite waste of resources trying to extract it into an icon. I don't see the icon to be reusable so it's okay to have it hard coded like this.

stavros-tomas (approved):

jeger-at (approved):

PR #1103, opened 2022-03-14
[ONL-6392] Adding EcFileDropzone component

jafcalvente:


src/components/ec-file-dropzone/ec-file-dropzone.vue
```diff
@@ -0,0 +1,100 @@
...
+    document.removeEventListener('dragleave', this.onDragleave);
+    document.removeEventListener('drop', this.disableDragging);
+  },
+  methods: {
+    onFileInputChange() {
+      this.fileList = [...this.$refs.fileInput.files];
+    },
+    onDrop(dragEvent) {
+      this.disableDragging();
+      this.fileList = [...dragEvent.dataTransfer.files];
```
jeger-at, line 70: Can we filter out folders here? Right now we also see the folder name, when it's dragged & dropped.

jafcalvente: Filtered!

jafcalvente:
I have modified the original approach to drag and drop event handling, away from the original approach based on the coordinates of the pointer on the screen. With the previous way, I was experiencing sporadic problems testing with different browsers. I tried now with Chrome and Firefox on Ubuntu, and Chrome and Safari on MacOS and it worked fine in all tests.

jmesao (approved):

stavros-tomas (approved):

stavros-tomas:
Great stuff @jafcalvente

PR #1119, opened 2022-03-22
ONL-6349: chore: Better data-test attributes for currency-input in case there are 2 on the same page.

mcibique:


src/components/ec-currency-input/ec-currency-input.vue
```diff
@@ -8,13 +8,13 @@
     >
       <span
         v-if="label"
-        data-test="ec-currency-input__label-text"
+        :data-test="prefixedDataTest('label-text')"
```
stavros-tomas, line 11: Hey @mcibique in the root template div would it help if we were adding a dynamic data-test instead?

mcibique: I'm not sure I understand.

stavros-tomas: I put the comment here because I couldn't comment on a line you didn't change.

This component doesn't have a main data-test. If at line 2 we add 
`:data-test="$attrs['data-test'] ? `${$attrs['data-test']} ec-currency-input`: 'ec-currency-input'"`

Would it help what you are trying to achieve?

mcibique: Only partially because my major problem is that the dropdown inside of this component doesn't have `list-data-test` set. I needed to do this prefixed data-test only because of that and once I had that one I decided to do it on all other parts of this component.

PR #1123, opened 2022-03-23
ONL-6465: Disable typing disabled dates

stavros-tomas:
At the EcDatepicker component we have 3 types of disabled dates
- Days before min date limit and days after max date limit (we don't allow to type or select)
- Bank holidays/specific days (we don't allow to select)
- Weekends (we don't allow to select)

All 3 types should have a unified behavior and we should not be able to select them or directly type a disabled date.
When we type a disabled date it should default to an empty input.

This PR is for expanding the behavior that we currently have for min/max dates where we are not allow to type a disabled date to also cover the bank holidays and weekends.

### DIFF
--- a/src/components/ec-datepicker/ec-datepicker.vue
+++ b/src/components/ec-datepicker/ec-datepicker.vue
@@ -218,7 +218,7 @@ export default {
           if (d) {
             const isoDate = this.isoDate(d);
 
-            if (this.disabledDatesMap.has(isoDate)) {
+            if (this.disabledDatesMap?.has(isoDate)) {
               this.clearInput();
               return;
             }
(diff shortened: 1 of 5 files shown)


src/components/ec-datepicker/ec-datepicker.vue
```diff
@@ -240,9 +255,24 @@ export default {
...
+      const day = dateObj.getDay();
+
+      const isSaturday = day === 6;
+      const isSunday = day === 0;
+
+      return isSaturday || isSunday;
+    },
     disableWeekends(dayElement) {
       if (this.areWeekendsDisabled) {
         const day = dayElement.dateObj.getDay();
```
jmesao, line 274: We could reuse the isWeekendDay method here :D

jmesao (approved):

jeger-at (approved):

PR #1130, opened 2022-03-25
ONL-6393 Adding of the multi-file listing component

derrick-g:
Adding of the multi file listing component that will show the files recently added as well as an option to delete an item.


src/components/ec-file-list/ec-file-list.vue
```diff
@@ -0,0 +1,119 @@
...
+  >
+    <div
+      v-for="(item, index) in items"
+      :key="item.name"
+      ref="items"
+      class="ec-file-list__item"
+      :data-test="`ec-file-list__item ec-file-list__item--${index}`"
+    >
+
+      <ec-icon
```
mcibique, line 14: I think we need to be able to display loading indicator here as well. I'm not sure if that's part of the ticket.

src/components/ec-file-list/ec-file-list.vue
```diff
@@ -0,0 +1,119 @@
+<template>
+  <div
+    data-test="ec-file-list"
+    class="ec-file-list"
+  >
+    <div
+      v-for="(item, index) in items"
+      :key="item.name"
+      ref="items"
```
mcibique, line 9: This ref doesn't seem to be used and can be removed.

src/components/ec-file-list/ec-file-list.vue
```diff
@@ -0,0 +1,119 @@
...
+      default: () => [],
+    },
+    isDeleteDisabled: {
+      type: Boolean,
+      default: false,
+    },
+  },
+  data() {
+    return {
+      name: 'test',
```
mcibique, line 64: We can remove this leftover.

src/components/ec-file-list/ec-file-list.vue
```diff
@@ -0,0 +1,119 @@
...
+    };
+  },
+  methods: {
+    onDelete(item) {
+      /* istanbul ignore else */
```
mcibique, line 69: We can test both of the cases. You can mount the component with `isDeleteDisabled` set to true/false, click on the element and then check what was emitted from the component using this: https://v1.test-utils.vuejs.org/api/wrapper/emitted.html#emitted. For disabled state, the emitted() should return an empty array.

EDIT: I actually found tests like that. Why it needs to be ignored then?

src/components/ec-file-list/ec-file-list.vue
```diff
@@ -0,0 +1,119 @@
...
+  },
+};
+</script>
+
+<style>
+  .ec-file-list {
+    &__item {
+      @apply tw-flex tw-flex-row;
+      @apply tw-pb-8;
+      @apply tw-items-center;
```
mcibique, line 81: We can join the flex stuff together with `@apply tw-flex tw-flex-row;`.

src/components/ec-file-list/ec-file-list.vue
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    &__name {
+      @apply tw-w-5/5;
```
mcibique, line 85: You can use `tw-w-full`. I'd use tw-w-5/5 when you also used other fractions somewhere in the code, e.g. tw-w-2/5.

src/components/ec-file-list/ec-file-list.vue
```diff
@@ -0,0 +1,119 @@
...
+      @apply tw-px-4;
+      @apply tw-text-gray-3;
+      @apply tw-overflow-hidden tw-overflow-ellipsis;
+      @apply tw-whitespace-nowrap;
```
mcibique, line 89: There's `tw-truncate` that does all 3 at the same time:
```
overflow: hidden;
text-overflow: ellipsis;
white-space: nowrap;
```

src/components/ec-file-list/ec-file-list.story.js
```diff
@@ -0,0 +1,44 @@
...
+  props: Object.keys(argTypes),
+  methods: {
+    onDelete(item) {
+      action('delete');
+      console.log('delete', item.name);
+    },
+  },
+  template: `
+    <div class="tw-p-20">
+      <div style="width: 100%;">
```
mcibique, line 28: `class="tw-w-full"`

src/components/ec-file-list/ec-file-list.vue
```diff
@@ -0,0 +1,119 @@
...
+        :title="item.name"
+        :data-test="`ec-file-list__name ec-file-list__name--${index}`"
+      >
+        {{ item.name }}
+      </span>
+
+      <ec-icon
+        class="ec-file-list__delete-icon"
+        :class=" { 'ec-file-list__delete-icon--is-disabled': isDeleteDisabled }"
+        :data-test="`ec-file-list__delete-icon ec-file-list__delete-icon----${index}`"
```
mcibique, line 33: `----${index}` should be just s dashes: `--${index}`.

src/components/ec-file-list/ec-file-list.spec.js
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  const items = [
+    { name: 'My invoice.pdf' },
+    { name: 'This is a very long file name that is for testing the ellipses.pdf' },
+    { name: 'DOC_123123_2423490802348_12312323.pdf' },
+    { name: 'number_rates_trades.xscl' },
+    { name: 'untitled.pdf' },
+  ];
+  describe('EcFileList', () => {
```
mcibique, line 22: Please add an empty line before `describe()` block and between each `it()`. It's just nicer to look at. There is no eslint to enforce that unfortunately, so we have to do that manually.

mcibique:
We also need to update visual snapshots for icon stories because 2 new icons have been added.

src/components/ec-file-list/ec-file-list.vue
derrick-g: Good remembering, still needs to be discussed so not part of this ticket.

src/components/ec-file-list/ec-file-list.vue
```diff
@@ -0,0 +1,119 @@
...
+
+      <span
+        class="ec-file-list__name"
+        :title="item.name"
+        :data-test="`ec-file-list__name ec-file-list__name--${index}`"
+      >
+        {{ item.name }}
+      </span>
+
+      <ec-icon
```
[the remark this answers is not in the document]
derrick-g: I like this idea, especially with the focusing and the structure for screen readers.

src/components/ec-file-list/ec-file-list.vue
derrick-g: Thanks, good spotting.

src/components/ec-file-list/ec-file-list.vue
derrick-g: We do test both cases (disabled and non disabled emit checks), with the disabled, I check if the emit sends back undefined, the code coverage just didn't want to highlight it? Not sure if I missed something?

src/components/ec-file-list/ec-file-list.vue
```diff
@@ -0,0 +1,119 @@
+<template>
+  <div
+    data-test="ec-file-list"
+    class="ec-file-list"
+  >
+    <div
+      v-for="(item, index) in items"
+      :key="item.name"
```
jafcalvente, line 8: If we use as key the name instead of the index we have the risk of triggering a Vue warning in case two files have the same name

src/components/ec-file-list/ec-file-list.vue
stavros-tomas: indeed we said to do the loading in another story

src/components/ec-file-list/ec-file-list.vue
stavros-tomas: Do we need to emit the index to delete?

derrick-g: I did it this way because the component lists a few items, when we delete we need to know which one the smart component need to remove or have to deal with.

stavros-tomas (approved):

jmesao (approved):

src/components/ec-file-list/ec-file-list.vue
```diff
@@ -0,0 +1,111 @@
...
+      @apply tw-w-24 tw-h-24;
+      @apply tw-fill-gray-4;
+    }
+
+    &__delete-btn {
+      @apply tw-border-none;
+      @apply tw-bg-transparent tw-fill-gray-4;
+      @apply tw-pr-0;
+
+      @media screen and (max-width: 480px) {
```
mcibique, line 94: We try to develop mobile first, so we should have code for smaller screens without `@media` and write `media` queries for bigger screens. So this could be written as 
```
@apply tw-px-0;

@media screen and (min-width: 480px) {
  @apply tw-px-....;
}
```

Ideally, we should first try to re-use one of the predefined breakpoints: https://github.com/Ebury/chameleon/blob/master/tailwind.config.js#L17. If those doesn't work, we can use a custom breakpoint like `480px`.

Tailwind has a nice helper called `@screen` to make it even easier to re-use existing breakpoints, e.g. from ec-donut.vue
```
    @apply tw-mr-0;
    @apply tw-flex tw-justify-center;

    @screen sm {
      @apply tw-w-auto;
      @apply tw-mr-24;
      @apply tw-block;
      @apply tw-justify-start;
    }
```

will generate `@media (min-width: 576px)` in the final CSS

mcibique: We're basically following this guide: https://tailwindcss.com/docs/responsive-design

PR #1134, opened 2022-04-01
[ONL-6449] - multi file upload

stavros-tomas:
Multi file upload component

* Wrapper of file-dropzone and file-list
* Disabled state
* Simple props like note and label
* Removes an element from the list upon request
* Exposes the files via v-model


src/components/ec-file-dropzone/ec-file-dropzone.spec.js
```diff
@@ -148,5 +156,21 @@ describe('EcFileDropzone', () => {
...
     });
+
+    it('@change - should not be emitted when is disabled', async () => {
+      const wrapper = mountFileDropzone({ isDisabled: true });
+
+      await wrapper.trigger('drop', { dataTransfer: { files: filesAndFolders } });
+      expect(wrapper.emitted('change')).toBeUndefined();
+    });
+
+    it('@change - should be emitted  when not disabled', async () => {
```
jmesao, line 167: two spaces after emitted and should not we say @change - should be emitted when is not disabled? Thanks!

jmesao (approved):

src/components/ec-file-dropzone/ec-file-dropzone.vue
```diff
@@ -1,59 +1,69 @@
...
+    <div
+      data-test="ec-file-dropzone__title"
+      class="ec-file-dropzone__title"
+      :class="{ 'ec-file-dropzone__title--is-disabled': isDisabled }"
+    >
       <slot name="title" />
-    </h3>
+    </div>
 
-    <label
```
jafcalvente, line 30: Why did we have to remove the label? Shouldn't each input have a label for accessibility reasons?

src/components/ec-file-upload/ec-file-upload.vue
```diff
@@ -0,0 +1,135 @@
...
+    note: {
+      type: String,
+    },
+    isDisabled: {
+      type: Boolean,
+    },
+  },
+  data() {
+    return {
+      fileList: [],
```
mcibique, line 84: Why is this needed? There is a full list available in `value` prop. You can use the `value` and you can remove the watcher.

src/components/ec-file-upload/ec-file-upload.vue
```diff
@@ -0,0 +1,135 @@
...
+    value: {
+      immediate: true,
+      handler(newValue) {
+        this.fileList = newValue;
+      },
+    },
+  },
+  methods: {
+    onChange(files) {
+      this.fileList = [...files];
```
jafcalvente, line 97: Shouldn't this component be responsible for maintaining all the files in the list? Right now, instead of adding the files received with the event to the list, we are replacing the existing list. I thought that this would be the component that would add files to the list it manages and that the files would only be deleted if the user manually deletes them via the button

src/components/ec-file-upload/ec-file-upload.vue
```diff
@@ -0,0 +1,135 @@
...
+      this.$emit('change', this.fileList);
```
mcibique, line 98: Emit `files` directly, do not store them in local state.

src/components/ec-file-upload/ec-file-upload.vue
```diff
@@ -0,0 +1,135 @@
...
+    },
+    onDelete(fileToDelete) {
+      this.fileList = this.fileList.filter(fileItem => fileItem !== fileToDelete);
```
mcibique, line 101: Filter `this.value` directly. do not store the filtered list in the local state.

src/components/ec-file-upload/ec-file-upload.story.js
```diff
@@ -0,0 +1,56 @@
...
+          href="#"
+          @click.prevent
+          :class="{
+            'tw-cursor-not-allowed' : isFileUploaderDisabled,
+            'tw-text-gray-6' : isFileUploaderDisabled,
+          }">browse</a>.</template>
+        <template #help-text>Attach documentation related to this transaction (ie. invoice(s), customer order, packing list, bill of lading, etc.)</template>
+      </ec-file-upload>
+
+      <button
```
mcibique, line 40: We should control this via Controls. Storybook automatically creates a control for each prop including the `is-disabled`. There is no need to have another button in the story for that.

src/components/ec-file-upload/ec-file-upload.story.js
```diff
@@ -0,0 +1,56 @@
...
+        class="tw-mt-48"
+        @click="toggleDisable"
+      >
+        Toggle disable
+      </button>
+
+      <p class="tw-mt-48">Value:</p>
+      <div v-for="item in model"><br>{{ item.name }}</div>
```
mcibique, line 48: Do not use `<br>`. Use TW margins if needed.

src/components/ec-file-upload/ec-file-upload.story.js
```diff
@@ -0,0 +1,56 @@
...
+  },
+  template: `
+    <div class="tw-p-24">
+      <ec-file-upload
+        v-bind="$props"
+        v-model="model"
+        :is-disabled="isFileUploaderDisabled"
+      >
+        <template #title>Drag and drop</template>
+        <template #subtitle>files here or <a
```
mcibique, line 30: An anchor without `href` should be a `<button>`.

src/components/ec-file-upload/ec-file-upload.story.js
```diff
@@ -0,0 +1,56 @@
...
+  component: EcFileUpload,
+};
+
+export const basic = (args, { argTypes }) => ({
+  components: { EcFileUpload },
+  props: Object.keys(argTypes),
+  data() {
+    return {
+      isFileUploaderDisabled: false,
+      model: [{ name: 'test', type: '.pdf' }],
```
mcibique, line 14: I can't find where is this `type` being used.

src/components/ec-file-dropzone/ec-file-dropzone.vue
```diff
@@ -119,14 +132,36 @@ export default {
...
-    @apply tw-small-text tw-cursor-pointer;
+    @apply tw-small-text;
+
+    &--is-disabled {
+      @apply tw-text-gray-6;
+    }
   }
 
   &--dragging {
     @apply tw-border-key-4;
```
jafcalvente, line 159: The blue border is being displayed on dragging even when the component is disabled. I think that we should not apply this effect when we are in disabled status.

src/components/ec-file-dropzone/ec-file-dropzone.vue
stavros-tomas: Yes, is related to that, then we had a catch up with Milos and we said to make it a div or something.

src/components/ec-file-upload/ec-file-upload.vue
stavros-tomas: We have 3 components
file-Dropzone: Allow the user to upload the files.
file list: Renders which files we have and if a user clicks to remove a file, it emits the item clicked to be delete.
file-upload: Is the core component that wraps the other two. The state is hold there. We upload there the data from dropzone and file list gets the data to render the list. File list acts as the medium of communication for the other two components. It also exposes a v-model with the data-hold.

src/components/ec-file-dropzone/ec-file-dropzone.vue
```diff
@@ -119,14 +132,36 @@ export default {
...
   &__img {
     @apply tw-fill-key-4 tw-mb-8;
+
+    &--is-disabled {
+      @apply tw-fill-gray-6;
+    }
+  }
+
+  &__title {
+    @apply tw-h3;
```
jafcalvente, line 142: Out of curiosity, why did you have to change the HTLM tag from `h3` to `div` and then apply the `tw-h3` class instead of leave it as a `h3`?

stavros-tomas: We were discussing it with Milos and Derrick. Can look like an h3 but semantically this is not an h3 so we changed to plain div.

src/components/ec-file-upload/ec-file-upload.story.js
stavros-tomas: In many cases like we have seen previously you do need to have because storybook unfortunately re-renders stuff every time you change a control. In this particular case I wanted to test the scenario

--> Load files
--> Disable
--> Check you cannot delete

But storybook was rerendering and you were loosing the files the moment you were swapping the isDisabled control. 
That said, I introduced in the story a  model where we prepopulate the value so we can test similar scenario that way.

src/components/ec-file-dropzone/ec-file-dropzone.vue
jafcalvente: Oh, I see. It's interesting. Until now I had only seen it from the perspective of the visual aspect, and it makes sense to look at it from that other point of view. Thanks for the explanation!

src/components/ec-file-upload/ec-file-upload.vue
jafcalvente: Yes, I agree with all that. Sorry if I didn't express myself well, but my comment was more in the sense that the `EcFileUpload` component is not handling the files received. It doesn't add them to the list of existing ones, but replaces the list. When an event with files is received from the `EcFileDropzone` component, the `EcFileUpload` component simply discards the existing files in the data `fileList` and replaces them with the new ones. I thought that when the component received an event with files from `EcFileDropzone` it would add it to the existing `fileList` data.

stavros-tomas: Great, thank you for pointing it out @jafcalvente , I added the functionality. Pending some test for it which for now give me a bit of hard time :)

jafcalvente:
Hey Stavros, I have realised one thing. When you click the delete button on a file, the file is deleted correctly but, if the file was not the last one in the list, the icon of the delete button of the item that occupies its place remains blue

stavros-tomas:
@jafcalvente That was very funny bug :D 
Solution and a good reason why we should not use index as key here --> https://github.com/Ebury/chameleon/pull/1134/commits/b7a69e6a295b098ecb5be5542acda00db97c113c

My only concern is if file name has any weird characters. @mcibique WDYT?

jafcalvente:
> 

At the same time, using the name as the _key_ does not allow us to add files with the same name even if they are really different files, and if I don't remember bad, we commented that we would allow this possibility.

src/components/ec-file-upload/ec-file-upload.vue
```diff
@@ -0,0 +1,131 @@
...
+      <slot
+        name="subtitle"
+        slot="subtitle"
+      />
+    </ec-file-dropzone>
+
+    <div
+      data-test="ec-file-upload__help-text"
+      class="ec-file-upload__help-text"
+      :class="{ 'ec-file-upload__help-text--is-disabled': isDisabled }"
```
jafcalvente, line 39: We have a misalignment here. The class used is `ec-file-upload__help-text--is-disabled`, however, within the `<style>` block the defined is `&--disabled` (without the `is-` prefix). I actually think you can remove both pieces of code: this line, and the `&--disabled` style block, since it is not being applied right now and it looks good. In fact, if you apply the defined style the text background is displayed in grey.

src/components/ec-file-upload/ec-file-upload.vue
```diff
@@ -0,0 +1,125 @@
...
+    },
+    isDisabled: {
+      type: Boolean,
+    },
+  },
+  methods: {
+    onChange(newFiles) {
+      let updatedFileList = [...this.value];
+
+      for (const newFile of newFiles) {
```
jafcalvente, line 85: I think could be enough only with having inside this `for` loop the content of the `else` block, since there you already are removing any possible duplicated file, and push the new one after. I mean, just having:
```
for (const newFile of newFiles) {
  updatedFileList = updatedFileList.filter(prevFile => prevFile.name !== newFile.name);
  updatedFileList.push(newFile);
}
```

But going one step further, perhaps you can avoid having nested loops by doing something like this:
```
onChange(newFiles) {
  const newFilesNames = newFiles.map(file => file.name);
  const updatedFileList = this.value.filter(file => !newFilesNames.includes(file.name));
  updatedFileList.push(...newFiles);
  this.$emit('change', updatedFileList);
},
```

stavros-tomas: Yes, this is super elegant, brilliant @jafcalvente!

jafcalvente (approved):

src/components/ec-file-upload/ec-file-upload.story.js
mcibique: I'd rather create another story for that particular scenario.

src/components/ec-file-upload/ec-file-upload.vue
mcibique: I disagree with this. We shouldn't filter users input without telling anything to the user - that's a bad UX. We should emit all files and let the parent to do the validation and display an error message about some files have been removed.

If you decide to keep it then we need to change the code to use Set() because `.include` inside of the `.filter` is O(n^2).
```js
const existingNames = new Set(this.value.map(file => file.name));
newFiles = newFiles.filter(file => !existingNames.has(file.name));
this.$emit('change', [this.value, newFiles]);
```

jeger-at: For file uploads, it's not necessarily bad UX. I'd argue, that most file uploads behave that way, so you could expect such behaviour. In theory, we could emit a replace event and show an information toast message, but personally, I'd not do that except UX wants to have this in place.

derrick-g: Also to consider is that whilst the filenames may be the same, the file may be different, so checking against file size and name may be beneficial if you would like to avoid the same file being added. I'd say the parent should do the validation as we may want to show a custom message of the sort such as 'You have already added XX file, do you want to continue?"

mcibique (approved):

jeger-at (approved):

derrick-g (approved):

PR #1184, opened 2022-05-31
Onl 6597   adding of refresh icon and exposing loading icon

derrick-g:
Adding of refresh icon and exposing the ec-loading-icon in main.js so it can be used as a separate component.


derrick-g:
> 

When I limit the cypress to just icons, it fails for flags for some reason, and needs to be updated.

stavros-tomas (approved):

PR #1208, opened 2022-08-03
ONL-6703-Checkbox-migration to Vue 3

derrick-g:


jafcalvente (approved):

jeger-at (approved):

src/components/ec-checkbox/ec-checkbox.vue
```diff
@@ -1,21 +1,26 @@
...
       v-model="inputModel"
-      :aria-describedby="errorId"
-      type="checkbox"
-      class="ec-checkbox__input"
-      :disabled="disabled"
-      data-test="ec-checkbox__input"
       @focus="inputIsFocused = true"
       @blur="inputIsFocused = false"
-      v-on="{ ...$listeners, 'update:modelValue': null }"
+      v-on="{ 'update:modelValue': null }"
```
mcibique, line 23: we shouldn't need this one after enabling MODE: 3. The problem with Vue 2 was that `<ec-checkbox v-model="blah">` made the `update:modelValue` event appearing in `$listeners` so I had to remove it from there so it didn't get attached to the child input. After removing $listeners, it's safe to remove this line, because we're now binding `$attrs` only and they do not contain the `update:modelValue` event because is declared in `defineEmits` array.

src/components/ec-checkbox/ec-checkbox.vue
```diff
@@ -114,48 +119,52 @@ export default {
...
-      },
-    },
-  },
-  watch: {
-    indeterminate(newValue) {
-      this.updateIndeterminate(newValue);
-    },
+);
+
+const uid = ref(getUid());
```
mcibique, line 124: I'd make this one static (no `ref`). We don't expect this one to change at runtime. The id is given when the component is created and that's it. It's less work for vue's reactivity and we don't have to use `uid.value` everywhere.

src/components/ec-checkbox/ec-checkbox.vue
```diff
@@ -114,48 +119,52 @@ export default {
...
+const slots = useSlots();
+
+const id = computed(() => (`ec-checkbox-${uid.value}`));
```
mcibique, line 127: Similar case for this one. This doesn't have to be a computed prop. We can just make it static too:
```js
const id = `ec-checkbox-${uid}`;
```

src/components/ec-checkbox/ec-checkbox.spec.js
```diff
@@ -204,6 +204,26 @@ describe('EcCheckbox', () => {
       );
       expect(wrapper.element).toMatchSnapshot();
     });
+    it('should render with custom attributes', () => {
```
mcibique, line 207: Add an empty line between `it()` calls.

src/components/ec-checkbox/ec-checkbox.spec.js
```diff
@@ -204,6 +204,26 @@ describe('EcCheckbox', () => {
...
+      const wrapper = mountCheckbox({}, {
+        attrs: {
+          'data-test': 'my-data-test',
+          class: 'my-class',
+          id: 'test-id',
```
mcibique, line 212: This id is overridden and not visible in the final snapshot. I'd keep it because it's a nice test that the user can't override it but I'd like to add another attribute that doesn't get overridden, e.g. `role="checkbox"` so we can see in the snapshot where custom attributes go in the rendered HTML.

mcibique (approved):

PR #1261, opened 2022-09-06
[ONL-6712] - migrate ec-timer to vue3

DisegnAle:
migrated time componet to vue3 
Paired with @derrick-g

### DIFF
--- a/src/components/ec-timer/ec-timer.vue
+++ b/src/components/ec-timer/ec-timer.vue
@@ -83,15 +83,15 @@ const offset = computed(() => circumference.value + steps.value * secondsLeft.va
 const radius = ref(24);
 const strokeWidth = ref(4);
 const secondsLeft = ref(props.seconds);
-const countdown = ref(null);
+let countdown = null;
 
 function startCountdown() {
-  countdown.value = new Countdown();
-  countdown.value.start(props.seconds);
-  countdown.value.on('time-updated', (newValue) => {
+  countdown = new Countdown();
+  countdown.start(props.seconds);
+  countdown.on('time-updated', (newValue) => {
     secondsLeft.value = newValue;
   });
-  countdown.value.on('time-expired', () => {
+  countdown.on('time-expired', () => {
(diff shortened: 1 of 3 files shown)


src/components/ec-timer/ec-timer.vue
```diff
@@ -46,96 +46,87 @@
...
+      return Number.isInteger(value) && value > 0;
     },
   },
-  emits: ['time-expired'],
-  data() {
-    return {
-      radius: 24,
-      strokeWidth: 4,
-      secondsLeft: this.seconds,
-      startTime: null,
```
DisegnAle, line 79: this variable as well as currentTime were not used. We removed them

src/components/ec-timer/ec-timer.vue
```diff
@@ -46,96 +46,87 @@
...
+const diameter = computed(() => radius.value * 2 + strokeWidth.value);
+const viewbox = computed(() => `0 0 ${diameter.value} ${diameter.value}`);
+const circumference = computed(() => 2 * Math.PI * radius.value);
+const steps = computed(() => circumference.value / props.seconds);
+const offset = computed(() => circumference.value + steps.value * secondsLeft.value);
+
+const radius = ref(24);
+const strokeWidth = ref(4);
+const secondsLeft = ref(props.seconds);
+const countdown = ref(null);
```
mcibique, line 86: Do we need the countdown to be reactive? is it used anywhere in the template or computed props? We can save some work for Vue's reactivity and just do this without `ref`. We don't need to then `countdown.value` anywhere but just `countdown`.

derrick-g: Agreed, it can just be a basic const. As its not in the template nor a computed function.

DisegnAle: changed it to be a let

mcibique (approved):

jeger-at (approved):

PR #1264, opened 2022-09-07
[ONL-6724] - migrate button dropdown component

DisegnAle:
migrated component to vue3


src/components/ec-btn-dropdown/ec-btn-dropdown.vue
```diff
@@ -5,31 +5,31 @@
     :data-test="$attrs['data-test'] ? `${$attrs['data-test']} ec-btn-dropdown` : 'ec-btn-dropdown'"
   >
     <ec-btn
-      :is-disabled="isDisabled"
+      :is-disabled="props.isDisabled"
```
mcibique, line 8: Props are automatically available in the template, so you don't have to update their references in the template. You can remove all occurrences of `props.*` from the template.

src/components/ec-btn-dropdown/ec-btn-dropdown.vue
```diff
@@ -44,55 +44,57 @@
...
-  data() {
+  listDataTest: {
+    type: String,
+  },
+});
+
+const emit = defineEmits(['click', 'change']);
+
+const isOpen = ref(false);
+const popoverOptions = ref({
```
mcibique, line 74: I noticed that this object is static and can't change in the future. We can optimize that either by removing the `ref()` so the reactivity doesn't have to keep track of it, or we can just inline the object into the template `:popover-options="{ autoSize: 'min', placement: 'bottom-end', }"`

src/components/ec-btn-dropdown/ec-btn-dropdown.vue
DisegnAle: corrected

src/components/ec-btn-dropdown/ec-btn-dropdown.vue
DisegnAle: corrected

mcibique (approved):

PR #1283, opened 2022-09-16
[ONL-6888] Disable @vue/compat

mcibique:
The entire library is now using only syntax that is fully compatible with Vue 3 and the compat mode with Vue 2 is no longer needed. Moving to Vue 3 without the compat mode will enable us to move to Vite and Vitest in next iteration.

Changes:
- Updated latest dependencies (SB, vue, vueuse)
- Removed `configureCompat` when running SB and Jest
- Removed `compatConfig`s from all components including patches for `floating-vue` package
- Re-enabled eslint rules that were disabled while refactoring old syntax
- Updated warning spies in tests as they are no longer reporting migration suggestions
- Migrate HOC helpers to Vue 3, they no longer have to rely on `$scopedSlots` - this was a workaround for a bug in `@vue/compat`
- Tried to remove the `v-model` hack in ec-input-field but it didn't work. The workaround is still needed in Vue 3 without compat mode, it needs more investigation

Please review commit by commit to get better understanding of context.


src/components/ec-input-field/ec-input-field.vue
```diff
@@ -204,9 +204,8 @@ function focus() {
 }
 
 watchEffect(() => {
-  // Hack: since we enabled @vue/compat we occasionally run into issues that the inputModel and value of the input are getting out of sync
+  // Hack: since Vue 3 we occasionally run into issues that the inputModel and value of the input are getting out of sync.
```
wolandec, line 207: Really strange behavior, is there any task for investigate it?

mcibique: Nope, I'll create one. It happens in our components only so I believe it has something to do with our code. The ec-dropdown uses ec-input-field in readonly mode to display the selected value. The problems start when we render the ec-dropdown with initial `modelValue` Without this hack, the input gets rendered empty instead of having the value in it. When checking the components in devtools, we see that inputModel in ec-input-field and modelValue in ec-dropdown are set and have the same value, but the input is empty ... I couldn't simulate it in https://sfc.vuejs.org/ so I believe there must be something going on with ec-dropdown and ec-input-field.

wolandec (approved):

src/components/ec-input-field/ec-input-field.vue
wolandec: That is the problem that VTU can't provide the native browser experience. I thought recently that maybe it is worth to use some headless solution for the unit tests too.

stavros-tomas (approved):

derrick-g (approved):

src/components/ec-input-field/ec-input-field.vue
mcibique: Yeap, it may be worth trying, both, Storybook and Cypress, have introduced component testing that you can execute in the browser.

PR #1289, opened 2022-09-26
[ONL-6702] - migrate navigation component to vue3

DisegnAle:
migrated component to vue3


src/components/ec-navigation/ec-navigation.vue
```diff
@@ -53,30 +53,27 @@
...
-    showBrandingLogo: {
-      type: Boolean,
-      default: true,
-    },
+<script setup>
+defineProps({
+  isCollapsed: {
+    type: Boolean,
+    required: true,
+    default: false,
```
wolandec, line 61: "default" or "required" field should be removed, there is no reason to keep both simultaniously.

src/components/ec-navigation/ec-navigation.vue
```diff
@@ -53,30 +53,27 @@
...
   },
-};
+  isCollapsable: {
+    type: Boolean,
+    required: true,
+  },
+  branding: {
+    type: Object,
+    default: () => ({}),
```
wolandec, line 69: Here is the same situation with default and required fields.

wolandec (approved):

mcibique (approved):

PR #1291, opened 2022-09-26
[ONL-6708] - Migrate donut component

corinaper:


src/components/ec-donut/ec-donut.vue
```diff
@@ -73,47 +73,37 @@
...
-      return this.dashArray * (1 - this.percentageUsed / 100);
-    },
+  used: {
+    type: Number,
+    required: true,
   },
-};
+});
+
+const radius = 48;
```
wolandec, line 92: In previous implementation this was a reactive parameter and now it isn't, why? If it is a contant then it should be naming in capslock case(RADIUS), so we could understand at the glance that it is unchangable value,

src/components/ec-donut/ec-donut.vue
```diff
@@ -73,47 +73,37 @@
...
+  if (props.used > props.amount) {
+    return 100;
+  }
+  if (props.used <= 0) {
+    return 0;
+  }
+  return (props.used / props.amount) * 100;
+});
+// Calculate the circumference
+const dashArray = computed(() => 2 * Math.PI * radius);
```
wolandec, line 104: Returned value is not an array. It is better to name it "circumference" and remove the comment above.

corinaper: I think dasharray here is referring to the svg element (stronke-dasharray and dashoffset), not an actual "array". But let me know if you think it's better to rename it.

src/components/ec-donut/ec-donut.vue
corinaper: thank you, I missed it, will change it to reactive param.

wolandec: But it never changes in the code, so maybe it is more constant than reactive variable?

src/components/ec-donut/ec-donut.vue
wolandec: I see :). Maybe it is better to rename both "array" vars and remove their comments, just to make the script code cleaner.

circumference
usedArcLength

src/components/ec-donut/ec-donut.vue
corinaper: true, just realised it doesn't. Will edit the naming

mcibique: In Vue 2 there weren't many options to make variables non-reactive because everything attached to the data (the state of the component) was made reactive automatically. In Vue 3, we can control that and only have reactive things that are going to change in the future. Making the variable static (non-reactive) means that Vue 3 reactivity can skip lots of checks when comparing changes in VDOM a make the updates in DOM a little bit faster.

Agree, it should be renamed to RADIUS, no `ref()` needed.

wolandec (approved):

src/components/ec-donut/ec-donut.spec.js
```diff
@@ -56,11 +56,11 @@ describe('EcDonut', () => {
...
   it('should get the correct computed properties', () => {
     const wrapper = mountDonut();
-    const circumference = 2 * Math.PI * wrapper.vm.radius;
+    const circumference = 2 * Math.PI * wrapper.vm.RADIUS;
     expect(wrapper.vm.percentageUsed).toBe(20);
     const offset = circumference * (1 - wrapper.vm.percentageUsed / 100);
-    expect(wrapper.vm.dashArray).toBe(circumference);
-    expect(wrapper.vm.dashOffset).toBe(offset);
+    expect(wrapper.vm.circumference).toBe(circumference);
+    expect(wrapper.vm.usedArcLength).toBe(offset);
```
mcibique, line 63: A test like these can be removed, we shouldn't test internals of the component like state and computed props. Accessing `wrapper.vm` is a bad thing, it makes tests less maintainable. Checking circumference and usedArcLength should be tested by checking their values in the snapshot and it's already tested on line 29.

More in the docs:

https://github.com/Ebury/eburyonline/blob/master/docs/FRONTEND.md?plain=1#L409

The only acceptable place to use `wrapper.vm` is when testing `v-model`:

https://github.com/Ebury/eburyonline/blob/master/docs/FRONTEND.md?plain=1#L447

mcibique (approved):

PR #1292, opened 2022-09-26
[ONL-6701] - Migrate user-info component to vue3

DisegnAle:
migrated component to vue3


wolandec (approved):

src/components/ec-user-info/ec-user-info.vue
```diff
@@ -29,29 +29,24 @@
...
-};
+  isCollapsable: {
+    type: Boolean,
+  },
+});
+const emit = defineEmits(['toggle']);
+
+function toggle() {
+  emit('toggle');
+}
```
mcibique, line 49: CAN BE MERGED: This function can be even inlined into the template because it's so small.

```
@click="emit('toggle')"
```

It was a good practise to create a method in Vue 2 because they weren't cached and were forcing unnecessary component updates, but now, in Vue 3, compiler handles inline functions and caches them automatically.

derrick-g (approved):

src/components/ec-user-info/ec-user-info.vue
DisegnAle: done

wolandec (approved):

mcibique (approved):

PR #1293, opened 2022-09-26
[ONL-6721] - Migrate phone-number-input to vue3

DisegnAle:
migrated component


src/components/ec-phone-number-input/ec-phone-number-input.vue
```diff
@@ -149,197 +149,186 @@
...
+const uid = getUid();
+const countriesHasFocus = ref(false);
+const popoverOptions = ref({
+  autoSize: 'min',
+});
+const popoverWidthReference = ref(null);
+
+const id = computed(() => `ec-phone-number-input-field-${uid}`);
+const errorId = computed(() => (isInvalid.value ? `ec-phone-number-input-${uid}` : null));
+const countriesItems = computed(() => props.countries.map(country => ({
```
wolandec, line 255: "props.countries" maybe undefined and code will throw an error.

src/components/ec-phone-number-input/ec-phone-number-input.vue
```diff
@@ -149,197 +149,186 @@
...
+  'after-open',
+  'country-change',
+  'phone-number-change',
+]);
+
+const uid = getUid();
+const countriesHasFocus = ref(false);
+const popoverOptions = ref({
+  autoSize: 'min',
+});
```
mcibique, line 250: We don't need a `ref` here. You could inline it into the template as well.

src/components/ec-phone-number-input/ec-phone-number-input.vue
```diff
@@ -149,197 +149,186 @@
...
+const errorId = computed(() => (isInvalid.value ? `ec-phone-number-input-${uid}` : null));
+const countriesItems = computed(() => props.countries.map(country => ({
+  areaCode: country.areaCode,
+  iconPath: getCountryFlagPath(country.countryCode),
+  id: country.countryCode,
+  name: country.text,
+  text: `${country.text} ${country.areaCode}`, // the text is only displayed in the tooltip, it's just to make items searchable by area code and name of the country
+  value: country,
+})));
+const countriesModel = computed({
```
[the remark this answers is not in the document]
mcibique: I would go even further and group stuff together. You can add comments to mark a sections too:

```
// ids
const uid
const id
const errorId

// items
const countriesItems
const selectedCountryAreaCode
const selectedCountryImage 
const selectedCountryName
function getCountryFlagPath(countryCode) {

// focus
const countriesHasFocus
function onFocusCountry() {

// models
const countriesModel = computed({
const phoneNumberModel = computed({
function onCountryChange(evt) {
function onPhoneNumberChange(evt) {

// popover
const popoverOptions = ref({
function getPopoverStyle() {
```

I think this is the benefit of having one setup function instead of the logic of the component scattered across the whole component like in the options API.

BTW, this component can be refactored in the future to support multiple v-models, that would be a breaking change which we don't want to add yet.

src/components/ec-phone-number-input/ec-phone-number-input.vue
```diff
@@ -149,197 +149,186 @@
...
+    emit('update:modelValue', {
+      ...props.modelValue,
+      phoneNumber,
+    });
+  },
+});
+const selectedCountryAreaCode = computed(() => props.modelValue?.country?.areaCode || null);
+const selectedCountryImage = computed(() => getCountryFlagPath(props.modelValue?.country?.countryCode));
+const selectedCountryName = computed(() => props.modelValue?.country?.name);
+const isInvalid = computed(() => !!props.errorMessage);
```
mcibique, line 292: This computed prop doesn't seem to be used anywhere else but one place, we can inline the logic there:
```
const errorId = computed(() => (props.errorMessage ? `ec-phone-number-input-${uid}` : null));
```

DisegnAle: i can see it is used on lines 38,105, 115

src/components/ec-phone-number-input/ec-phone-number-input.vue
DisegnAle: corrected, inserted a default value

src/components/ec-phone-number-input/ec-phone-number-input.vue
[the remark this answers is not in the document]
DisegnAle: ordered the script structure

src/components/ec-phone-number-input/ec-phone-number-input.vue
DisegnAle: removed ref on popover options and put it inline

src/components/ec-phone-number-input/ec-phone-number-input.vue
mcibique: I meant the `isInvalid` is not used anywhere else but in `errorId` computed prop. So we can inline it into the `errorId` computed prop.

DisegnAle: ah ok but also isInvalid is present on lines 116, 122, 252

mcibique (approved):

src/components/ec-phone-number-input/ec-phone-number-input.vue
wolandec: As I mentioned in Corina's PR there is no need to introduce isInvalid here as it is just renaming of errorMessage prop.

DisegnAle: removed isInvalid variable

src/components/ec-phone-number-input/ec-phone-number-input.vue
```diff
@@ -107,19 +107,19 @@
...
         @change="onPhoneNumberChange"
-        @focus="$emit('focus')"
+        @focus="emit('focus')"
       />
     </div>
 
     <div
       :id="errorId"
-      v-if="isInvalid"
+      v-if="!!props.errorMessage"
```
wolandec, line 116:
```suggestion
      v-if="errorMessage"
```

wolandec (approved):

PR #1295, opened 2022-09-27
[ONL-6767] Migrate Currency Input component

corinaper:
migrated currency input


src/components/ec-currency-input/ec-currency-input.vue
```diff
@@ -110,166 +110,151 @@
...
+
+const uid = getUid();
+let currenciesHasFocus = false;
+const popoverOptions = {
+  autoSize: 'min',
 };
+const popoverWidthReference = ref(null);
+const attrs = useAttrs();
+
+const id = computed(() => `ec-currency-input-field-${uid}`);
```
wolandec, line 197: `uid` is not reactive so here we could remove computed

src/components/ec-currency-input/ec-currency-input.vue
```diff
@@ -110,166 +110,151 @@
...
+const errorId = computed(() => (isInvalid.value ? `ec-currency-input-field-${uid}` : null));
+const isInvalid = computed(() => !!props.errorMessage);
```
wolandec, line 199: This is a redundant computed and we could use props.errorMessage instead.

src/components/ec-currency-input/ec-currency-input.vue
```diff
@@ -110,166 +110,151 @@
...
+const currenciesItems = computed(() => props.currencies.map(currency => ({ text: currency, value: currency, id: currency })));
```
wolandec, line 200: props.currencies could be undefined and an error will be thrown.

src/components/ec-currency-input/ec-currency-input.vue
```diff
@@ -110,166 +110,151 @@
...
+const currenciesTooltipOptions = computed(() => {
+  const { content, placement = 'top' } = props.disabledCurrenciesTooltip || {};
```
wolandec, line 202: Even though we do not use TypeScript yet it is worth moving that kind of entities that looks like enums is such kind of structures, to use in code not a string but a property of object.

So here we could use something like this:

`
  const { content, placement = TooltipPlacement.TOP } = props.disabledCurrenciesTooltip || {};
`

src/components/ec-currency-input/ec-currency-input.vue
```diff
@@ -110,166 +110,151 @@
...
+  if (!content) {
+    return null;
+  }
+  return { content, placement };
+});
+const currencyModel = computed({
```
wolandec, line 208:
```suggestion

const currencyModel = computed({
```

Code with blank line delimiters between big snippets or between logical different parts looks more readable.

src/components/ec-currency-input/ec-currency-input.vue
corinaper: I think this just might be for convenience as we also use this constant twice within the template. Happy to change them all to props.errorMessage if you think this makes more sense though.

src/components/ec-currency-input/ec-currency-input.vue
corinaper: true, will add a condition. Thank you

src/components/ec-currency-input/ec-currency-input.vue
wolandec: It could be used in the way the errorTooltipMessage variable is used, and that will be even more semantic because right in the place errorMessage is used, we could see what "is invalid". That approach with computed isInvalid variable is suitable for cases when there is some logic under the hood of the computed callback function. Here there is no one.

src/components/ec-currency-input/ec-currency-input.vue
wolandec: uid will be the same in both cases. Even more, computed reacts and recalls its callback function only if that function has a reactive stuff inside it, but uid is not reactive, so errorId will not be changed because of it.

src/components/ec-currency-input/ec-currency-input.vue
```diff
@@ -27,7 +27,7 @@
         v-if="isCurrenciesDisabled"
         v-ec-tooltip="currenciesTooltipOptions"
         class="ec-currency-input__currencies ec-currency-input__currencies--is-disabled"
-        :class="{ 'ec-currency-input__currencies--is-disabled-and-has-error': isInvalid }"
+        :class="{ 'ec-currency-input__currencies--is-disabled-and-has-error': !!errorMessage }"
```
wolandec, line 30: We can use just errorMessage without double negative !!

src/components/ec-currency-input/ec-currency-input.vue
```diff
@@ -68,19 +68,19 @@
...
         @change="onAmountChange"
-        @focus="$emit('focus')"
+        @focus="emit('focus')"
       />
     </div>
 
     <div
       :id="errorId"
-      v-if="isInvalid"
+      v-if="!!errorMessage"
```
wolandec, line 77: here as well

src/components/ec-currency-input/ec-currency-input.vue
```diff
@@ -68,19 +68,19 @@
...
       :data-test="prefixedDataTest('error-text')"
       class="ec-currency-input__error-text"
     >
       <span>{{ errorMessage }}</span>
       <ec-icon
-        v-if="isInvalid && errorTooltipMessage"
+        v-if="!!errorMessage && errorTooltipMessage"
```
wolandec, line 83: and here

src/components/ec-currency-input/ec-currency-input.vue
```diff
@@ -110,166 +110,153 @@
...
+let currenciesHasFocus = false;
+const popoverOptions = {
+  autoSize: 'min',
 };
+const popoverWidthReference = ref(null);
+const attrs = useAttrs();
+const id = `ec-currency-input-field-${uid}`;
+
+// eslint-disable-next-line no-extra-boolean-cast
+const errorId = computed(() => (!!props.errorMessage ? `ec-currency-input-field-${uid}` : null));
```
wolandec, line 200: and here props.errorMessage is enough, we use !! when want to convert some value to the boolean

For example `const a = !!someVariable` , but in logical operations like here this is not necessary as JS will do it for us.

corinaper: thank you

src/components/ec-currency-input/ec-currency-input.vue
mcibique: We already call them enums e.g.: https://github.com/Ebury/chameleon/blob/master/src/enums/sort-direction.js. If you are going to create one then also don't forget to export it in `main.js`: https://github.com/Ebury/chameleon/blob/master/src/main.js#L7, so it can be re-used in EBO too.

All possible values for the enum can be found here: https://github.com/Akryum/floating-vue/blob/main/packages/floating-vue/src/util/popper.ts#L5

src/components/ec-currency-input/ec-currency-input.vue
```diff
@@ -110,166 +110,152 @@
...
+  },
+  noCurrenciesText: {
+    type: String,
+    default: 'No results found',
+  },
+});
+
+const emit = defineEmits(['update:modelValue', 'change', 'focus', 'amount-change', 'currency-change', 'open', 'after-open']);
+
+const uid = getUid();
```
mcibique, line 190: This component is very similar to the phone-number-input refactored in the other PR. I commented there that we should be grouping chunks of code together, e.g. like this: https://github.com/Ebury/chameleon/pull/1293#discussion_r980251636. 

The options API was grouping state together, computed props into one place, methods into one place, but the functionality related to them was scattered across the whole component. `<script setup>` allows us to move state, computed props and methods that are related together. We can use the same grouping as in ec-phone-number-input.

corinaper: sure, done

stavros-tomas (approved):

wolandec (approved):

PR #1299, opened 2022-09-29
ONL-6717[chore] - Migrate metroline components

stavros-tomas:
Migrate metroline components


src/components/ec-metroline-item/ec-metroline-item.vue
```diff
@@ -103,69 +103,62 @@
...
-    isCompleted() {
-      return this.status === MetrolineItemStatus.COMPLETED;
-    },
-  },
-  created() {
-    this.metroline.register(this.id);
-  },
-  beforeUnmount() {
-    this.metroline.unregister(this.id);
+const metroline = inject('metroline');
```
wolandec, line 114: If we use provide/inject pattern it should be only situation where components are tightly coupled. If we decided that `ec-metroline-item` and `ec-metroline` component are such tightly coupled, then I believe they should be placed in one directory named `ec-metroline`, and `ec-metroline-item` should be placed in subdirectory `elements` as it could be used only inside `ec-metroline`

src/components/ec-metroline/ec-metroline.vue
```diff
@@ -7,62 +7,55 @@
...
+  },
+  complete: () => {
+    metroline.isCompleted = true;
+    emit('complete');
+  },
+});
+
+const metrolineItemIds = ref([]);
+
+provide('metroline', metroline);
```
wolandec, line 58: while we providing something it is better to use Symbol as identifier, because strings are not unique.

src/components/ec-metroline-item/ec-metroline-item.vue
```diff
@@ -103,69 +103,62 @@
   </div>
 </template>
 
-<script>
+<script setup>
+import {
+  computed, inject, onBeforeUnmount, onMounted, toRefs,
+} from 'vue';
+
 import * as MetrolineItemStatus from '../../enums/metroline-item-status';
```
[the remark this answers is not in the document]
mcibique: Makes sense having the enum closer to the implementation. This enum AFAIK is not exported and not used in EBO either, so it has more like internal feeling that is used only internally in metroline-item.

src/components/ec-metroline/ec-metroline.vue
mcibique: Agree, let's create a symbol in this file, export it and import it in metroline-item. I don't think we should be moving it to a separate file.

src/components/ec-metroline-item/ec-metroline-item.vue
```diff
@@ -103,69 +103,62 @@
...
-    complete() {
-      this.metroline.complete();
-    },
+  badgeText: {
+    type: String,
   },
-};
+});
+
+const { id } = toRefs(props);
```
mcibique, line 126: You don't have to destructure the props, you can also use `props.id` directly. It shows the origin of the prop and you don't need `.value` then.

src/components/ec-metroline/ec-metroline.vue
stavros-tomas: At the end I moved it externally in a folder called "Symbols". Seems script setup does not like exports.

src/components/ec-metroline/components/ec-metroline-item/ec-metroline-item.vue
```diff
@@ -103,69 +103,66 @@
...
-    this.metroline.register(this.id);
-  },
-  beforeUnmount() {
-    this.metroline.unregister(this.id);
+<script setup>
+import {
+  computed, inject, onBeforeUnmount, onMounted,
+} from 'vue';
+
+import METROLINE_SYMBOL from '../../../../symbols/provide-inject';
```
wolandec, line 111: We should keep all constants that belong to some code close to that code. When we use inject in this situation we will use it in components that are in subdirectory of our parent metroline component that provides data, then it is worth placing its symbols inside its directory.

wolandec: So here we could place file constants.js in `ec-metroline` directory.

mcibique (approved):

PR #1302, opened 2022-09-30
[ONL-6844] - Use disable payment creation in CTA btn

DisegnAle:
- new disabling buttons logic. Accordingly, the 'isBtnDropdownDisabled' prop is only disabling the btn-dropdown. The btn-popover instead is automatically disabled when all the dropdown's items passed are disabled
- renamed isDisabled prop to isBtnDropdownDisabled


src/components/ec-btn-dropdown/ec-btn-dropdown.vue
```diff
@@ -71,17 +71,17 @@
...
 import EcDropdownSearch from '../ec-dropdown-search';
 
-defineProps({
+const props = defineProps({
   items: {
     type: Array,
     default: () => [],
   },
-  isDisabled: {
+  isBtnDropdownDisabled: {
```
wolandec, line 84: Am I right that we could disable button, but have a popover with links available at the same time?

derrick-g (approved):

mcibique (approved):

PR #1303, opened 2022-09-30
[ONL-6719] chore - Migrate inline actions

stavros-tomas:
Migrate inline actions

### DIFF
--- a/src/components/ec-inline-actions/ec-inline-actions.vue
+++ b/src/components/ec-inline-actions/ec-inline-actions.vue
@@ -67,54 +67,53 @@
   </ec-popover>
 </template>
 
-<script>
-import EcClosePopover from '../../directives/ec-close-popover';
-import EcTooltip from '../../directives/ec-tooltip';
+<script setup>
+import VEcClosePopover from '../../directives/ec-close-popover';
+import VEcTooltip from '../../directives/ec-tooltip';
 import EcIcon from '../ec-icon';
 import EcPopover from '../ec-popover';
 
-export default {
-  components: { EcPopover, EcIcon },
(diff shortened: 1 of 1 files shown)


src/components/ec-inline-actions/ec-inline-actions.vue
```diff
@@ -67,54 +67,53 @@
...
+    return item.download;
+  }
+  if (!!item.href && item.download === '') {
+    return '';
+  }
+  return null;
+}
+
+function getHrefAttr(item) {
+  return !!item.href && !item.disabled ? item.href : null;
```
wolandec, line 111:
```suggestion
  return item.href && !item.disabled ? item.href : null;
```

wolandec (approved):

src/components/ec-inline-actions/ec-inline-actions.vue
stavros-tomas: Hey Anton why remove the `!!` ?

wolandec: `item.href` will convert to boolean by js as it is logical `&&` after it.

stavros-tomas: Not sure this is 100% accurate in all cases 
```js
!!'' && !false // Will output "false"
'' && !false // Will output ''
```

wolandec: We have to test in context of that ternary operator because it converts `'' && !false` into boolean too.

`'' && !false ? '' : null` returns `null`

mcibique (approved):

PR #1304, opened 2022-09-30
[ONL-6912] Added complementary colors to chameleon and gradients that are using them

mcibique:
The Figma spec has another colors theme called complementary colors, those are only to pair them with key colors to create gradients. We use a gradient like this in the legacy code (login page) and now there's a requirement for the modern code (Cards).

Complementary colors are brandable but only 7 brands (all based on Ebury's base color) are going to have them set. Rest of the brands will use key-color-4 and key-color-2 as a backup if complementary colors are not specified (that's what legacy does too).

There will be another PR in EBO to populate `--ec-theme-complementary-color-level-1` and `--ec-theme-complementary-color-level-2` from brand settings. The goal is to be able use a new class `tw-bg-gradient-1` to produce the same gradient as in Figma.

Design link: https://www.figma.com/file/E8q7GWDqRi6WG6pmmTURpG/01.-Chameleon-Library?node-id=0%3A1
Checkout the SB: https://chameleon-git-onl-6912-complementary-color-ebury.vercel.app/?path=/story/css-colors--all. You can change the theme via CSS Resources addon to see how it behaves for other themes which don't have these colors set.

### DIFF
--- a/src/styles/settings/colors/colors.story.js
+++ b/src/styles/settings/colors/colors.story.js
@@ -1,10 +1,66 @@
+import { createEventHook } from '@vueuse/core';
 import Color from 'color';
+import {
+  getCurrentInstance, onBeforeMount, onBeforeUnmount,
+} from 'vue';
 
 export default {
   title: 'CSS/Colors',
 };
 
+function useCssResourceAddonSync({ global, document }) {
+  const onChanged = createEventHook();
+
+  let mutationObserver;
+
+  onBeforeMount(() => {
+    if (global.MutationObserver) {
+      mutationObserver = new global.MutationObserver((ev) => {
+        onChanged.trigger(ev);
+      });
+      mutationObserver.observe(document.body, {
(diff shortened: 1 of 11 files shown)


src/styles/settings/colors/colors.story.js
```diff
@@ -1,10 +1,66 @@
...
 export default {
   title: 'CSS/Colors',
 };
 
+function useCssResourceAddonSync({ global, document }) {
+  const onChanged = createEventHook();
+
+  let mutationObserver;
+
+  onBeforeMount(() => {
```
wolandec, line 16: Why in previous implementation it was in mounted hook and now in beforeMount?

mcibique: True, I'll fix that.

stavros-tomas (approved):

derrick-g (approved):

PR #1305, opened 2022-09-30
ONL-6709 - Migration of File Upload

derrick-g:
Migration of the file components


src/components/ec-file-dropzone/ec-file-dropzone.vue
```diff
@@ -42,82 +42,86 @@
...
-      this.dragEnterLeaveEventCounter--;
-    },
-    disableDragging() {
-      this.dragEnterLeaveEventCounter = 0;
-    },
-  },
-};
+});
+
+const dragEnterLeaveEventCounter = ref(0);
```
wolandec, line 64: Why we use here a number and not a boolean? We could even delete this variable and replace it with

const isDragging = ref(false)

src/components/ec-file-dropzone/ec-file-dropzone.vue
```diff
@@ -42,82 +42,86 @@
...
+  // Necessary for the drop listener to work:
+  // https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API#define_a_drop_zone
+  dragEvent.preventDefault();
+}
+
+function onDragleave() {
+  dragEnterLeaveEventCounter.value--;
+}
+
+function disableDragging() {
```
wolandec, line 121: The name of this function is not semantic, it not represents its goal. We can remove it and use isDragging in all the places it is used now.

src/components/ec-file-upload/ec-file-upload.vue
```diff
@@ -42,47 +42,42 @@
...
+  update([...updatedFileList, ...newFiles]);
+}
+
+function onDelete(fileToDelete) {
+  update(props.modelValue.filter(fileItem => fileItem !== fileToDelete));
+}
+
+function update(items) {
+  emit('update:modelValue', items);
+  emit('change', items);
```
wolandec, line 79: Why we emit here two identic events? We could stick with one `change`. I would understand if we emitted here updated items, I mean **new** files in `update:modelValue` event, but we don't.

derrick-g: The `change` emit is filtered upwards into EBO and is used to trigger file validation. This can in the future be refactored into the watch function in those two files, and then removed here.

wolandec: I propose to stay with `change` and remove `update:modelValue`. BTW, I didn't find using of update:modelValue in EBO.

src/components/ec-file-dropzone/ec-file-dropzone.vue
derrick-g: The goal of the component is for when a user starts dragging a file anywhere on the page, the dropzone highlights, directing the user to drop the file on it. Because of this, both the `onDragEnter` and `onDragLeave` functions are constantly firing as the user drags their file around, with a boolean it creates unpredictable results. The counter is always in a positive range when a dragging event is occurring and then setting it to zero or when the user leaves the page goes below 0, removing the highlighting.

src/components/ec-file-dropzone/ec-file-dropzone.vue
derrick-g: I don't think it should be removed, but I agree with the naming, maybe it can be renamed to resetDraggingCounter to accurately represent it.

src/components/ec-file-upload/ec-file-upload.vue
derrick-g: The `update:modelValue` updates the v-model of the component which contains the file(s) from what I can see so maybe refactoring the `change` event in the future could be the better route.

src/components/ec-file-dropzone/ec-file-dropzone.vue
wolandec: Thank you! I see, sorry.

src/components/ec-file-upload/ec-file-upload.vue
wolandec: OMG, native update:modelValue event, I see :).

stavros-tomas (commented):
Can't comment directly on that line and pls correct me if I am wrong, does line 38 on`ec-file-dropzone.vue` needs a small change to get rid of the dollar sign? Seems to be working now but  what is the most correct way to do it. With or without?

```
@click="fileInput.click()"
```

stavros-tomas (approved):

derrick-g:
> Can't comment directly on that line and pls correct me if I am wrong, does line 38 on`ec-file-dropzone.vue` needs a small change to get rid of the dollar sign? Seems to be working now but what is the most correct way to do it. With or without?
> 
> ```
> @click="fileInput.click()"
> ```

Thats a good point @mcibique is it best to remove the $ref in the template, especially in this case where we referring to it for the element click event?

mcibique:
> > Can't comment directly on that line and pls correct me if I am wrong, does line 38 on`ec-file-dropzone.vue` needs a small change to get rid of the dollar sign? Seems to be working now but what is the most correct way to do it. With or without?
> > ```
> > @click="fileInput.click()"
> > ```
> 
> Thats a good point @mcibique is it best to remove the $ref in the template, especially in this case where we referring to it for the element click event?

We no longer needs `$refs` in templates because node refs became a standard `ref()` values. So we should change the `@click="$refs.fileInput.click()"` to just `@click="fileInput.click()"`. I'm not sure if we need the the `fileInput.value.click()` there (we shouldn't).

stavros-tomas (approved):

PR #1308, opened 2022-10-04
[ONL-6575] - Replace focus trap directive

DisegnAle:
Replacing focus trap custom directive with vueuse composable functions


DisegnAle:
@mcibique i added a couple of tests but i am not sure how to tests the @vueuse functionalities, to test whether they were called or not

src/components/ec-modal/ec-modal.spec.js
```diff
@@ -33,6 +42,31 @@ describe('EcModal', () => {
     expect(wrapper.element).toMatchSnapshot();
   });
 
+  it('should be focus trapped', async () => {
```
wolandec, line 45: This test is testing that after a one `tab` you will get focus on a close button. But in the name of test it is saying that it should trap focus. We should be testing in a test a behaviour that it declares, and here it is trapping focus, that means do not let another elements outside component get it while we are "tabing".

src/components/ec-modal/ec-modal.spec.js
```diff
@@ -1,10 +1,19 @@
 import { mount } from '@vue/test-utils';
+import { useFocusTrap } from '@vueuse/integrations/useFocusTrap';
 import { defineComponent } from 'vue';
 
 import { withMockedConsole } from '../../../tests/utils/console';
 import EcModal from './ec-modal.vue';
 
-jest.mock('../../directives/ec-focus-trap');
+jest.mock('@vueuse/integrations/useFocusTrap', () => ({
```
wolandec, line 8: Here we are going to use in our unit test our knowledge about internal implementation that component has. It is worth testing component as a black box with interface, and do not care about code inside it.

src/components/ec-modal/ec-modal.vue
```diff
@@ -156,10 +156,20 @@ function hasPositiveButton() {
 function hasNegativeButton() {
   return !!slots.negative;
 }
+
+// zIndex
```
wolandec, line 160: We know from the variable name that this is a zIndex, I think the comment is not necessary here. We could add some comments to separate and unit some parts of code from time to time, in those cases when we want to declare some blocks of code and it is not obviously without comments.

src/components/ec-modal/ec-modal.spec.js
DisegnAle: so should just i change the assertion?

src/components/ec-modal/ec-modal.spec.js
DisegnAle: i don't get what i am supposed to do

src/components/ec-modal/ec-modal.vue
DisegnAle: yes the comment is also needed just to separate the code's groups

src/components/ec-modal/ec-modal.spec.js
wolandec: I'd try to test the trapping functionality described in the name of the test. For that purpose, we could create an environment for our modal component, as you started on line 46. For example, div.#root with a `button` inside it, mount modal to the div.#root and then trigger the "tab" key, checking every time that `button` doesn't get a focus.

src/components/ec-modal/ec-modal.spec.js
wolandec: We shouldn't test the internal implementation of the component. So mocking the libraries used inside the component and not exposed outside of the component should be considered a red flag.

In this case, we should reconsider our testing approach to the trap functionality of the `ec-modal` component to avoid mocking and testing its internal implementation.

So if https://github.com/Ebury/chameleon/pull/1308/files#r987874713 would work, the mocking of useFocusTrap will not be needed.

src/components/ec-modal/ec-modal.vue
```diff
@@ -170,9 +180,17 @@ function getFocusTrapOptions() {
 
   return options;
 }
-
-// zIndex
-const zIndexStyle = computed(() => (props.zIndex ? { zIndex: props.zIndex } : null));
+const { deactivate, pause, unpause } = useFocusTrap(focusTrapTarget, getFocusTrapOptions(focusTrapOptions));
```
mcibique, line 183: `getFocusTrapOptions()` function doesn't accept any arguments.

src/components/ec-modal/ec-modal.vue
```diff
@@ -156,10 +156,20 @@ function hasPositiveButton() {
...
+const focusTrapOptions = {
+  immediate: true,
+  escapeDeactivates: props.isClosable,
+  clickOutsideDeactivates: props.isClosable,
+};
 function getFocusTrapOptions() {
   const options = {
-    escapeDeactivates: props.isClosable,
-    clickOutsideDeactivates: props.isClosable,
+    ...focusTrapOptions,
```
mcibique, line 172: If `focusTrapOptions` is used only on this place, I'd keep the options inline in this function.

src/components/ec-modal/ec-modal.spec.js
mcibique: `useFocusTrap` is a wrapper around the `focus-trap` library and it can't run in JSDOM because it accesses lots of API that JSDOM doesn't implement. So we decided to not test the trap at all in unit tests and always ignore it. We have an integration test in EBO to check that 2 modals can be displayed at the same time, we don't have any tests checking that the focus is really trapped in a modal. If we want to test the trap, let's do it in EBO using Cypress so we have the browser environment to test everything.

We were even mocking the focus-trap directive globally using the `ec-focus-trap.mock.js` in `tests/mocks` so it is always ignored when we use our modals in unit tests.

tests/mocks/ec-focus-trap.mock.js
```diff
@@ -1,16 +0,0 @@
-jest.mock('../../src/directives/ec-focus-trap', () => {
```
mcibique, line 1: We need to create a new global mock that will work in EBO when using modals in the unit tests.

mcibique (approved):

PR #1310, opened 2022-10-05
[ONL-6720] - Chore - Migrate inline input field components

stavros-tomas:
Migrate inline input field components


src/components/ec-inline-input-field/ec-inline-input-field.vue
```diff
@@ -55,99 +55,86 @@
...
+  },
+  errorMessage: {
+    default: '',
+    type: String,
+  },
+});
+
+const emit = defineEmits(['cancel', 'edit', 'submit']);
+
+const valueForLoading = ref(null);
```
mcibique, line 116: I'd call it `editedValue`. In Vue 2 this valueForLoading has been initialised from the props:
```
  data() {
    return {
      valueForLoading: this.value,
    };
```
So this code should be `const valueForLoading = ref(props.value);`

src/components/ec-inline-input-field/ec-inline-input-field.vue
```diff
@@ -55,99 +55,86 @@
...
+
+const textClasses = computed(() => {
```
mcibique, line 118: I'm not sure there are any benefits of these `textClasses` computed props in all of these components. We could use simpler syntax in template:
```
:class="{
  'ec-inline-input-field__content': true,
  [config.sensitiveClass]: isSensitive,
}"
```

src/components/ec-inline-input-field/ec-inline-input-field.vue
```diff
@@ -55,99 +55,86 @@
...
+
+  return classes;
+});
+
+function cancel() {
+  emit('cancel');
+}
+function edit() {
+  emit('edit');
+}
```
mcibique, line 133: both, cancel and edit, can be inlined into the template too, there are used only on one place and Vue 3 now caches the inline functions in the template, so it's no longer a performance gain.

src/components/ec-inline-input-field/components/edit/edit.vue
```diff
@@ -55,59 +55,67 @@
...
-    },
-    errorMessage: {
-      default: '',
-      type: String,
-    },
-    labelTooltip: {
-      default: '',
-      type: String,
-    },
+const inputId = computed(() => `ec-inline-input-field-edit__input-${getUid()}`);
```
mcibique, line 68: This value is not going to change over time, no need to make it reactive. It's more work for Vue 3 to track all reactive stuff.

```
const inputId = `ec-inline-input-field-edit__input-${getUid()}`;
```

mcibique (approved):

PR #1312, opened 2022-10-05
ONL-6930 Migrate to Vite

wolandec:


src/components/ec-phone-number-input/ec-phone-number-input.vue
```diff
@@ -333,12 +339,11 @@ export default {
       this.$emit('phone-number-change', evt);
     },
     getCountryFlagPath(countryCode) {
-      if (!countryCode) {
+      if (!countryCode || !Object.keys(countries).includes(countryCode)) {
```
mcibique, line 342: This `getCountryFlagPath` function is called in a for loop, I think we can optimise this a bit. The `countries` object is not going to change so we can create a Set outside of the component:
```
import countries from 'svg-country-flags/countries.json';
const supportedCountries = new Set(Object.keys(countries));
```

... and then just call `if (!countryCode || !supportedCountries.has(countryCode)) {`

vite.config.js
```diff
@@ -0,0 +1,14 @@
+import vue from '@vitejs/plugin-vue';
+import { defineConfig } from 'vite';
+
+const path = require('path');
+
+// https://vitejs.dev/config/
+export default defineConfig({
+  plugins: [vue()],
```
mcibique, line 8: Should we add eslint and stylelint plugins? Does storybook displays linter errors in the overlay when using vite builder instead of webpack?

package.json
```diff
@@ -65,9 +68,7 @@
     "@storybook/theming": "6.5.12",
     "@storybook/vue3": "6.5.12",
     "@testing-library/jest-dom": "5.16.5",
-    "@vue/cli-plugin-babel": "5.0.8",
-    "@vue/cli-service": "5.0.8",
-    "@vue/compiler-sfc": "3.2.39",
```
mcibique, line 70: I think we can also remove:
```
    "@storybook/builder-webpack5": "6.5.12",
    "@storybook/manager-webpack5": "6.5.12",
```

vite.config.js
wolandec: It doesn't show. I'm not sure that overlay is helpful for DX. It sometimes is a little bit annoying, and all that overlay says - say, Eslint and Stylelint in a console. Are we sure that we need it?

stavros-tomas: From my perspective is not a blocker to not have it but if I could have it in my environment I wouldn't turn it off.

mcibique (approved):

PR #1314, opened 2022-10-06
ONL-6939,6940 Replace vue-cli with ESLint and Jest

wolandec:


package.json
```diff
@@ -17,12 +17,12 @@
     "icons": "webpack --config ./src/icons/webpack.config.js --mode=production",
     "lint": "npm run lint:js && npm run lint:style",
     "lint:no-fix": "npm run lint:js:no-fix && npm run lint:style:no-fix",
-    "lint:js": "vue-cli-service lint ./",
-    "lint:js:no-fix": "vue-cli-service lint --no-fix ./",
+    "lint:js": "eslint --ext .js,.jsx,.vue --ignore-path .gitignore --ignore-pattern svg-sprites.js --fix src",
+    "lint:js:no-fix": "eslint --ext .js,.jsx,.vue --ignore-path .gitignore --ignore-pattern svg-sprites.js --no-fix src",
```
mcibique, line 21: We need to run eslint over `tests` and `.storybook` folder too. That's why we were executing the lint service with `./`. We can also control what is ignored with .eslintignore files, e.g. the `svg-sprites.js` is already ignored.

package.json
```diff
@@ -49,8 +49,7 @@
     "vue": "3.2.39"
   },
   "devDependencies": {
-    "@babel/core": "7.19.1",
-    "@babel/eslint-parser": "7.19.1",
+    "@babel/eslint-parser": "^7.19.1",
```
mcibique, line 52: All versions in package.json must be fixed because we don't trust npm community. We've seen so many breaking changes shipped as minors that we have everything fixed. I think we should enforce that in the `.npmrc` and add `save-exact=true`.

package.json
```diff
@@ -109,8 +108,9 @@
     "svgo": "2.8.0",
     "svgo-loader": "3.0.1",
     "vue-router": "4.1.5",
+    "vue3-jest": "^27.0.0-alpha.1",
```
mcibique, line 111: Can we check why we need this one? I don't like it's alpha and I also see we have `@vue/vue3-jest` already. I think we should be using that one.

src/components/ec-smart-table/__snapshots__/ec-smart-table.spec.js.snap
```diff
@@ -461,6 +461,22 @@ exports[`EcSmartTable filtering should handle changes in filters and reload the
     Test 2 (N/A)
   </div>
   
+  <!-- TODO ONL-4893
```
mcibique, line 464: There's a this part in the jest.config to configure the compiler:
```
'vue-jest': {
      compilerOptions: {
        whitespace: 'condense',
        isCustomElement: tag => customElements.has(tag),
        comments: false,
      },
    },
```
It disables the comments, so they are not rendered into the HTML. It seems like this option is now not working. We need to check it and revert all these snapshots.

package.json
wolandec: For sure, I will leave `@vue/vue3-jest`, but in peerDependencies there is ` "@vue/vue3-jest": "^27.0.0-alpha.3",`. In master too.

src/components/ec-smart-table/__snapshots__/ec-smart-table.spec.js.snap
wolandec: Yes, It is solved by changing the transformer you've mentioned upper. The minor version 1 -> 3 in the alpha build of @vue/vue3-jest makes sense, I guess.

.eslintrc.js
```diff
@@ -3,9 +3,7 @@ module.exports = {
   reportUnusedDisableDirectives: true,
   env: {
     node: true,
-  },
-  parserOptions: {
-    parser: '@babel/eslint-parser',
```
mcibique, line 8: I think we can remove the `@babel/eslint-parser` from the `package.json` too.

.eslintrc.js
```diff
@@ -3,9 +3,7 @@ module.exports = {
...
+    es2021: true,
```
mcibique, line 6: Nice, this would help Vite performance too. Any concern why we shouldn't go with `es2022`? We like having every modern syntax (at least in Stage 3) available.

package.json
```diff
@@ -17,12 +17,12 @@
...
-    "lint:js": "vue-cli-service lint ./",
-    "lint:js:no-fix": "vue-cli-service lint --no-fix ./",
+    "lint:js": "eslint --fix ./",
+    "lint:js:no-fix": "eslint --no-fix ./",
     "lint:style": "stylelint ./src/**/*.{css,vue} --fix",
     "lint:style:no-fix": "stylelint ./src/**/*.{css,vue}",
-    "test": "vue-cli-service test:unit",
-    "test:watch": "vue-cli-service test:unit --watch",
+    "test": "jest",
+    "test:watch": "jest --watch",
```
mcibique, line 25: We can remove `@vue/cli-plugin-unit-jest` from the package.json too.

package.json
```diff
@@ -17,12 +17,12 @@
     "icons": "webpack --config ./src/icons/webpack.config.js --mode=production",
     "lint": "npm run lint:js && npm run lint:style",
     "lint:no-fix": "npm run lint:js:no-fix && npm run lint:style:no-fix",
-    "lint:js": "vue-cli-service lint ./",
-    "lint:js:no-fix": "vue-cli-service lint --no-fix ./",
+    "lint:js": "eslint --fix ./",
+    "lint:js:no-fix": "eslint --no-fix ./",
```
mcibique, line 21: We can remove `@vue/cli-plugin-eslint` from the package.json too.

mcibique (approved):

PR #1317, opened 2022-10-10
Onl 6948 disable autocomplete for chameleon input components

corinaper:
Added prop to ec-input to autocomplete, default is null. 
Added autocomplete prop to phone number component as well.

set autocomplete to ‘OFF’ for every input that has an ‘amount’


src/components/ec-input-field/ec-input-field.spec.js
```diff
@@ -45,6 +45,11 @@ describe('EcInputField', () => {
     expect(wrapper.element).toMatchSnapshot();
   });
 
+  it('renders properly with the given prop autocomplete', () => {
```
wolandec, line 48: We need to test a positive case too, when we set input autocomplete to true.

src/components/ec-phone-number-input/ec-phone-number-input.vue
```diff
@@ -237,6 +238,10 @@ export default {
         return ['notification', 'modal', 'tooltip', 'level-1', 'level-2', 'level-3'].includes(value);
       },
     },
+    autocomplete: {
```
wolandec, line 241: We should add this controls into the stories, to let Chameleon user see that property right in the Storybook.

src/components/ec-input-field/ec-input-field.spec.js
corinaper: done

src/components/ec-input-field/ec-input-field.vue
```diff
@@ -155,6 +156,10 @@ const props = defineProps({
     type: Boolean,
     default: false,
   },
+  autocomplete: {
+    type: String,
```
wolandec, line 160: In the task the `autocomplete` is boolean.

mcibique: I think we should change that in the task, because autocomplete can have lots of string values: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete. I'd like to support everything the HTML can.

wolandec: Agree.

stavros-tomas:
@corinaper I think is worth having this to the <ec-textarea> as well since is another form of text input. If you add it, please add a small unit test similar to what you did for input element. Thank you.

corinaper:
> @corinaper I think is worth having this to the as well since is another form of text input. If you add it, please add a small unit test similar to what you did for input element. Thank you.

@stavros-tomas  I'm not sure what you're referring to, could you please expand on what is worth having and where

stavros-tomas:
> > @corinaper I think is worth having this to the as well since is another form of text input. If you add it, please add a small unit test similar to what you did for input element. Thank you.
> 
> @stavros-tomas I'm not sure what you're referring to, could you please expand on what is worth having and where

Sorry I was referring to the "ec-textarea". I wrote it with "<" and github hid it :D

wolandec (approved):

mcibique (approved):

stavros-tomas (approved):

PR #1329, opened 2022-10-20
[ONL-6924] - Fix keyboard navigation in btn dropdown

DisegnAle:


src/components/ec-dropdown-search/ec-dropdown-search.vue
```diff
@@ -260,26 +269,48 @@ function show() {
   }
 }
 
+function waitForPopoverFocus() {
+  return new Promise((resolve) => {
+    requestAnimationFrame(() => {
+      requestAnimationFrame(() => {
+        requestAnimationFrame(resolve);
```
wolandec, line 276: I see this code not yours, but why we use here this chain of RAF's?

src/components/ec-dropdown-search/ec-dropdown-search.vue
```diff
@@ -334,24 +365,45 @@ function isItemSelected(item) {
...
 // CTA
+const itemsOverflowContainer = ref(null);
 const isCtaAreaFocused = ref(false);
+const { deactivate, activate } = useFocusTrap(itemsOverflowContainer, getFocusTrapOptions());
+function getFocusTrapOptions() {
+  return {
+    immediate: false,
+    escapeDeactivates: true,
+    clickOutsideDeactivates: true,
+    fallbackFocus: popoverWrapper,
```
wolandec, line 376: This fallbackFocus option do not change too during the runtime. So there is no reason to create function.

`fallbackFocus` should be `HTMLElement` or `SVGElement` or `string` or `Function`, but `popoverWrapper` is a `Ref`. 

I suppose we should pass here `() => popoverWrapper.value` as a fallbackFocus.

src/components/ec-dropdown-search/ec-dropdown-search.vue
```diff
@@ -260,26 +269,48 @@ function show() {
...
   if (props.isSearchEnabled) {
-    await new Promise((resolve) => {
-      requestAnimationFrame(() => {
-        requestAnimationFrame(() => {
-          requestAnimationFrame(resolve);
-        });
-      });
-    });
-
+    await waitForPopoverFocus();
```
wolandec, line 288: We can move all three `waitForPopoverFocus` before the `if` on the line 287 as it is called in every condition.

src/components/ec-dropdown-search/ec-dropdown-search.vue
DisegnAle: because before using all the callbacks after the opening of the dropdown we need to wait for it to be opened. so this is the solution it was found

src/components/ec-dropdown-search/ec-dropdown-search.vue
```diff
@@ -123,6 +124,7 @@
                   ...tooltipOptions,
                   ...item.tooltip,
                 }"
+                :tabindex="!hasLinksInItems ? 0: -1"
```
wolandec, line 127:
```suggestion
                :tabindex="hasLinksInItems ? -1: 0"
```
It is better to operate with positive conditions, our brain understands those much more easier.

src/components/ec-dropdown-search/ec-dropdown-search.vue
```diff
@@ -494,6 +549,8 @@ function loseFocus() {
     ctaAreaElementFocusable.blur();
   }
 }
+
+const hasLinksInItems = computed(() => props.items.some(item => !!item.to || !!item.href));
```
wolandec, line 553:
```suggestion
const hasLinksInItems = computed(() => props.items.some(item => item.to || item.href));
```
Casting to boolean will be proceeding for us by js as these values are used in logical operation ||.

src/components/ec-dropdown-search/ec-dropdown-search.vue
wolandec: Why don't use `nextTick` here?

src/components/ec-dropdown-search/ec-dropdown-search.vue
DisegnAle: infact the first condition is a positive statement, isn't it?

src/components/ec-dropdown-search/ec-dropdown-search.vue
DisegnAle: corrected, thanks

src/components/ec-dropdown-search/ec-dropdown-search.vue
DisegnAle: corrected thanks

src/components/ec-dropdown-search/ec-dropdown-search.vue
DisegnAle: corrected, thanks

src/components/ec-dropdown-search/ec-dropdown-search.vue
```diff
@@ -334,24 +365,44 @@ function isItemSelected(item) {
 const selectedItemIndex = computed(() => filteredItems.value.indexOf(toRaw(props.modelValue)));
 
 // CTA
+const itemsOverflowContainer = ref(null);
 const isCtaAreaFocused = ref(false);
+const { deactivate, activate } = useFocusTrap(itemsOverflowContainer, getFocusTrapOptions());
+function getFocusTrapOptions() {
```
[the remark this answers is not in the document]
DisegnAle: corrected thanks

src/components/ec-dropdown-search/ec-dropdown-search.vue
wolandec: The condition is only one here `"hasLinksInItems` and it is negative because of `!` that means `negative`. For that who read this condition easier to see conditions without `!`. So if you could avoid it - it is better to do.

DisegnAle: ah ok understood. thank you

src/components/ec-dropdown-search/ec-dropdown-search.vue
```diff
@@ -494,6 +548,8 @@ function loseFocus() {
     ctaAreaElementFocusable.blur();
   }
 }
+
+const hasLinksInItems = computed(() => props.items.some(item => item.to || item.href));
```
mcibique, line 552: This makes dropdown-search depending on logic from ec-btn-dropdown. We don't need tabindexes on `<li>`s if they don't contain links? The native select don't do that either.

src/components/ec-dropdown-search/ec-dropdown-search.vue
mcibique: I'm not sure where the comment documenting this hack went, I remember writing one.

Every time we open the ec-popover (which uses v-popover from floating-vue), they try to get a focus on the floating div after it's open. But they don't do it immediately, but after 2 RAFs (they have an util for it: https://github.com/Akryum/floating-vue/blob/main/packages/floating-vue/src/util/frame.ts#L1). This breaks lots of our functionality, because we want to focus on first item in the ec-popover like search or CTA if they are available. If we just focus in afterShow, the focus will be moved to div 2 RAFs after. So the trick is to wait 3 frames and bypass their nonsense. That's why this function does what it does.

src/components/ec-dropdown-search/ec-dropdown-search.spec.js
```diff
@@ -289,6 +290,19 @@ describe('EcDropdownSearch', () => {
...
   });
 
+  it('should initialize the "useFocusTrap composable" with mandatory options', () => {
+    mountDropdownSearch({ }, {
+      slots: {
+        default: '<a href="#" @click.prevent>Open me</a>',
+      },
+    });
+    const options = useFocusTrap.mock.calls[0][1];
+    expect(Boolean(options.clickOutsideDeactivates)).toBe(true);
```
mcibique, line 300: Why do you convert these to `Boolean`? They are all boolean already.

src/components/ec-dropdown-search/ec-dropdown-search.vue
DisegnAle: this means that we cannot focus on any items when opening the dropdown, is that fine?

src/components/ec-dropdown-search/ec-dropdown-search.spec.js
DisegnAle: corrected, thanks

mcibique (approved):

PR #1330, opened 2022-10-21
[ONL-6974] Migrate EcDateRangeFilter component

corinaper:


corinaper:
Let me know if you think it's worth replacing ec-input with datepicker

src/components/ec-date-range-filter/ec-date-range-filter.vue
```diff
@@ -50,101 +50,102 @@
...
+
+// disable clear button
+const numberOfSelectedFilters = computed(() => {
+  let datesSelected = 0;
+  if (fromValueDate.value && toValueDate.value) {
+    datesSelected = 2;
+  } else if (fromValueDate.value || toValueDate.value) {
+    datesSelected = 1;
+  }
+  return datesSelected;
```
wolandec, line 136:
```suggestion
  if (fromValueDate.value && toValueDate.value) {
    return 2;
  } 
  
  if (fromValueDate.value || toValueDate.value) {
    return 1;
  }
  
  return 0;
```

src/components/ec-date-range-filter/ec-date-range-filter.vue
```diff
@@ -50,101 +50,102 @@
...
+});
+const isDisabled = computed(() => !numberOfSelectedFilters.value > 0);
```
wolandec, line 138:
```suggestion
const isButtonDisabled = computed(() => numberOfSelectedFilters.value <= 0);
```
It is better and more natural for our mind to avoid and understand easier a positive not a negative conditions. Name of variable should explain itself. This makes reading of the code more fluent and doesn't force reader to jump from script to template and vice versa to understand what isDisabled for.

wolandec (approved):

mcibique (approved):

PR #1331, opened 2022-10-21
[ONL-6976] Migrate ec multiple values selection component

corinaper:


src/components/ec-multiple-values-selection/ec-multiple-values-selection.vue
```diff
@@ -110,120 +110,111 @@
...
+
+const searchInputRef = ref(null);
+function focus() {
+  if (props.isSearchable) {
+    nextTick(() => {
+      searchInputRef?.value.focus();
+    });
+  }
+}
+defineExpose({ focus });
```
wolandec, line 188: Why do we expose focus here?

wolandec (approved):

src/components/ec-multiple-values-selection/ec-multiple-values-selection.vue
corinaper: It is used in the parent component: EcSyncMultipleValuesFilter, without exposing it, the tests in the parent are failing. But let me know if you think there's a better approach

mcibique (approved):

PR #1339, opened 2022-10-25
[ONL-6725] migrate ec table filter component

corinaper:


src/components/ec-table-filter/ec-table-filter.vue
```diff
@@ -45,78 +29,62 @@
...
-    // toggleMoreFiltersButtonIcon() {
-    //   return this.isLessFilterActive ? 'simple-chevron-down' : 'simple-chevron-up';
-    // },
-    // toggleMoreFiltersText() {
-    //   return this.isLessFilterActive ? this.moreFiltersButtonText : this.lessFiltersButtonText;
-    // },
+  filters: {
+    type: Array,
+    required: true,
+    default: () => ([]),
```
wolandec, line 50: Here should be whether required or default

mcibique (approved):

PR #1347, opened 2022-10-31
[ONL-6986] - Opaque backdrop for ec-modal

DisegnAle:


src/components/ec-full-screen-overlay/ec-full-screen-overlay.vue
```diff
@@ -0,0 +1,105 @@
...
+                <h1
+                  v-if="title"
+                  data-test="ec-full-screen-overlay__title"
+                  class="ec-full-screen-overlay__title"
+                >
+                  {{ title }}
+                </h1>
+              </div>
+              <div class="tw-col-full sm:tw-col-3 tw-text-right">
+                <a
```
mcibique, line 26: We're trying to improve the A11Y everywhere, we should use unstyled button instead of `<a href="#">`. We should also add a text "Close" into the button visible only for screen readers (`tw-sr-only`).

src/components/ec-full-screen-overlay/ec-full-screen-overlay.vue
```diff
@@ -0,0 +1,105 @@
...
+                  href="#"
+                  data-test="ec-full-screen-overlay__close-icon-container"
+                  @click="emit('closeOverlay')"
```
mcibique, line 29: Emits should be in `kebab-case` but there's no need to repeat the name of the component in it, so `close` is enough.

src/components/ec-full-screen-overlay/ec-full-screen-overlay.vue
```diff
@@ -0,0 +1,105 @@
...
+<script setup>
+import { useFocusTrap } from '@vueuse/integrations/useFocusTrap';
+import { onUnmounted, ref } from 'vue';
+
+import EcIcon from '../ec-icon';
+
+defineProps({
+  title: {
+    type: String,
+    default: null,
```
mcibique, line 59: I think the title should be mandatory. The component would look like weird if it wasn't provided.

src/components/ec-full-screen-overlay/ec-full-screen-overlay.vue
```diff
@@ -0,0 +1,105 @@
...
+      <div
+        ref="overlayContainer"
+        class="tw-col-full"
+        data-test="ec-full-screen-overlay__container"
+      >
+        <div class="tw-grid tw-grid-cols-3">
+          <div class="tw-col-full sm:tw-col-3" />
+          <div class="tw-col-full sm:tw-col-6">
+            <div class="tw-grid">
+              <div class="tw-col-full sm:tw-col-9">
```
mcibique, line 16: We can just use a basic flex with 2 columns instead of grid, the one column for title can have flex-grow set to 1, the one for the icon to 0. Then we're sure that the icon only removes space as small as possible from the space needed for title and you you can also remove `tw-text-right` from the icon. Another thing is that having tw-col-9 only `min-width: 580px` breaks the title on small screens:

Screenshot from 2022-11-03 18-07-51

src/components/ec-full-screen-overlay/ec-full-screen-overlay.vue
```diff
@@ -0,0 +1,105 @@
+<template>
```
mcibique, line 1: We need a placeholder for the content we want to put into the component and provide a `<slot>` that we will populate from the EBO.

src/components/ec-full-screen-overlay/ec-full-screen-overlay.story.js
```diff
@@ -0,0 +1,24 @@
...
+  },
+  template: `
+    <ec-full-screen-overlay v-bind="args"></ec-full-screen-overlay>
+  `,
+});
+
+export const basic = Template.bind({});
+basic.args = {
+  title: 'I am a title',
+};
```
mcibique, line 23: Could you also add a story which displays more content in the component? Lots of parahraphs with Lorem ipsum texts is enough, but I'd like to see the overflow on this component working.

src/components/ec-full-screen-overlay/ec-full-screen-overlay.vue
```diff
@@ -0,0 +1,105 @@
...
+</script>
+
+<style>
+  @import '../../styles/tools/transitions.css';
+
+  .ec-full-screen-overlay {
+    @apply tw-flex tw-flex-row tw-items-stretch;
+    @apply tw-min-h-screen;
+    @apply tw-z-level-3;
+    @apply tw-absolute;
```
mcibique, line 85: I think the position should be `fixed`, because `absolute` is affected by the parent with `position: relative`. `ec-modal` does exactly that.

src/components/ec-full-screen-overlay/ec-full-screen-overlay.vue
stavros-tomas: What if we use this in the future for something like a mobile navigation menu? Maybe there the title won't be necessary

src/components/ec-full-screen-overlay/ec-full-screen-overlay.vue
```diff
@@ -0,0 +1,110 @@
+<template>
+  <transition name="ec-full-screen-overlay__fade">
+    <div
+      v-show="show"
```
stavros-tomas, line 4: Even if `v-show` does the job I have the feeling v-if is better for this particular case.
The reason is `v-show` will keep the element always in the DOM and just toggle its visibility, on the other hand `v-if` will completely destroy it so when we do not need it.

src/components/ec-full-screen-overlay/ec-full-screen-overlay.vue
```diff
@@ -0,0 +1,110 @@
...
+                  href="#"
+                  data-test="ec-full-screen-overlay__close-icon-container"
+                  @click="emit('closeOverlay')"
+                >
+                  <ec-icon
+                    class="ec-full-screen-overlay__close-icon"
+                    data-test="ec-full-screen-overlay__close-icon"
+                    name="simple-close"
+                    type="interactive"
+                    :size="15"
```
stavros-tomas, line 37: Please speak to designers. I know the design says 15px but is this is wrong for 2 reasons:
* is not a multiple of 4
* Is not big enough, will be hard for people to notice, click, "touch"

Most likely needs to be 24px. Same like ec-modal

src/components/ec-full-screen-overlay/ec-full-screen-overlay.vue
```diff
@@ -0,0 +1,105 @@
...
+      data-test="ec-full-screen-overlay"
+    >
+      <div
+        ref="overlayContainer"
+        class="tw-col-full"
+        data-test="ec-full-screen-overlay__container"
+      >
+        <div class="tw-grid tw-grid-cols-3">
+          <div class="tw-col-full sm:tw-col-3" />
+          <div class="tw-col-full sm:tw-col-6">
```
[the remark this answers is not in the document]
stavros-tomas: Maybe as Milos say we can try a simpler structure without grid? The grid itself if we need it we can pass it via slot. Something like the following:

```html
<div data-test="ec-full-screen-overlay">
  <div data-test="ec-full-screen-overlay__content">
    <header data-test="ec-full-screen-overlay__header">
      <slot name="header"></slot>
      <a data-test="ec-full-screen-overlay__close"><ec-icon data-test="ec-full-screen-overlay__close"</a>
    >
    <main data-test="ec-full-screen-overlay__main"><slot name="main"></slot></main>
  </div>
<div>
```
Regarding the `ec-full-screen-overlay`
* occupies the whole screen

Regarding the `ec-full-screen-overlay__content`
* Centered in the screen
* Probably needs a max width of 704px. This to be confirmed with designers. I have the feeling though that we cannot have this metroline bigger/smaller than C&P. C&P has a  max-width 704px. 
* Potentially this max-width can be a prop for future re-usability? Let's discuss it with the team
* Lets give it a 24px margin as well

Regarding the `ec-full-screen-overlay__close`
* Make sure it does not move around from mobile to desktop. Maybe needs to be absolute positioned.

src/components/ec-full-screen-overlay/ec-full-screen-overlay.vue
```diff
@@ -0,0 +1,140 @@
...
+    @apply tw-fixed;
+    @apply tw-inset-0;
+    @apply tw-bg-gray-8;
+    @apply tw-flex;
+    @apply tw-justify-center;
+    @apply tw-overflow-y-auto;
+
+    &__content {
+      max-width: 704px;
+      width: 704px;
```
stavros-tomas, line 99: Few notes on this one:
* potentially this could have been 100% so we will not have to change 2 values when we change the width.
* Did designers confirm the 704px that is the correct value?
* Do we want this at the end with adjustable width as a prop or we hard code for now?

src/components/ec-full-screen-overlay/ec-full-screen-overlay.vue
```diff
@@ -0,0 +1,140 @@
...
+      @apply tw-m-0;
+      @apply tw-text-gray-3;
+    }
+
+    &__fade {
+      @mixin ec-fade-transition;
+    }
+
+    @media (max-width: 1024px) {
+      @mixin ec-full-screen-overlay-responsive;
```
stavros-tomas, line 136: I have some doubts if this will be something re-usable. Maybe should live here instead of the tools. Let's see what @mcibique thinks

src/components/ec-full-screen-overlay/ec-full-screen-overlay.vue
```diff
@@ -0,0 +1,140 @@
...
+            data-test="ec-full-screen-overlay__title"
+            class="ec-full-screen-overlay__title"
+          >
+            {{ title }}
+          </h1>
+
+          <button
+            class="ec-full-screen-overlay__close-icon-btn"
+            data-test="ec-full-screen-overlay__close-icon-btn"
+            @click="emit('close-overlay')"
```
stavros-tomas, line 28: Silly thing but can we please make this `@click="emit('close')"`

We tend to have it like that everywhere else. We do not repeat the name of the component in the event.
So we have simple event names like `close`, `change`, `hide` etc.

Thank you

src/components/ec-full-screen-overlay/ec-full-screen-overlay.vue
mcibique: Use a CSS variable so it can we overridden if needed in EBO.

src/components/ec-full-screen-overlay/ec-full-screen-overlay.vue
mcibique: It would be reusable if there was a prop to turn the responsivness on/off like ec-alert does. If we don't plan to turn this on/off we don't need a mixin and it can be declared in this file.

src/components/ec-full-screen-overlay/ec-full-screen-overlay.vue
```diff
@@ -0,0 +1,140 @@
...
+      @apply tw-h1;
+      @apply tw-m-0;
+      @apply tw-text-gray-3;
+    }
+
+    &__fade {
+      @mixin ec-fade-transition;
+    }
+
+    @media (max-width: 1024px) {
```
mcibique, line 135: `max-width` in media queries is prohibited. We should style the component for mobile first, without any media queries and then style the component for larger screens. Check the TW docs in for better explanation: https://expert-umbrella-249394df.pages.github.io/FRONTEND.html#responsive-design

mcibique (approved):

stavros-tomas (approved):

PR #1360, opened 2022-11-16
[ONL-7025]  - SummaryInfo component

DisegnAle:


src/components/ec-summary-info/ec-summary-info.story.js
```diff
@@ -0,0 +1,76 @@
...
+    <ec-summary-info
+      v-bind="args"
+    />
+  `,
+});
+
+export const basic = Template.bind({});
+basic.args = {
+  title: 'This is the title',
+  iconName: 'simple-sell',
```
wolandec, line 26: It is worth having lineItems here, to not fill it manually in storybook

src/components/ec-summary-info/ec-summary-info.story.js
```diff
@@ -0,0 +1,76 @@
...
+  `,
+});
+
+withContentLinesProp.args = {
+  ...basic.args,
+  lineItems: [
+    {
+      text: 'Some text',
+      cssClasses: ['tw-text-gray-4', 'tw-uppercase'],
+      iconCssClasses: ['tw-h-16'],
```
wolandec, line 47: For now I'd not create this cssClasses options in line Items and stick with preset for line only(style property in lineItems in the task). We could name it preset  instead of style to not confuse it with style attribute.

src/components/ec-summary-info/ec-summary-info.story.js
wolandec, line 26: Can't we put it inside return in line 12? It seems to be a complexity with no need here.

src/components/ec-summary-info/ec-summary-info.story.js
```diff
@@ -0,0 +1,76 @@
...
+  },
+  template: `
+    <ec-summary-info
+      v-bind="args"
+    />
+  `,
+});
+
+withContentLinesProp.args = {
+  ...basic.args,
```
wolandec, line 42: and here too, we could put it right in return object on line 32

src/components/ec-summary-info/ec-summary-info.story.js
```diff
@@ -0,0 +1,76 @@
...
+        <div class="tw-flex tw-small-text">
+          <span class="tw-text-key-4">
+            Some text from a slot
+          </span>
+        </div>
+    </ec-summary-info>
+  `,
+});
+
+withSlots.args = {
```
wolandec, line 73: to line 60

src/components/ec-summary-info/ec-summary-info.vue
```diff
@@ -0,0 +1,131 @@
...
+    class="ec-summary-info"
+    data-test="ec-summary-info"
+  >
+    <div
+      v-if="iconName"
+    >
+      <ec-icon
+        class="ec-summary-info__icon"
+        data-test="ec-summary-info__icon"
+        :class="{'tw-mr-12': iconName}"
```
wolandec, line 12: Why we use here iconName as condition as we already check it on line 7?

src/components/ec-summary-info/ec-summary-info.vue
```diff
@@ -0,0 +1,131 @@
...
+
+<script setup>
+
+import { useSlots } from 'vue';
+
+import VEcTooltip from '../../directives/ec-tooltip';
+import EcIcon from '../ec-icon';
+
+defineProps({
+  title: {
```
wolandec, line 75: There is no need to have a title property anymore, this will be a line, with preset style "secondary"(name is an example). Ordinary line will have preset name "default"

src/components/ec-summary-info/ec-summary-info.vue
```diff
@@ -0,0 +1,131 @@
...
+    type: Array,
+    default: () => [],
+  },
+});
+
+const slots = useSlots();
+
+function hasSlot(name) {
+  return name in slots;
+}
```
wolandec, line 93: Why do we have this code here? We just need to wrap our `ec-summary-info__content` div with `<slot></slot>`
https://vuejs.org/guide/components/slots.html#slot-content-and-outlet

src/components/ec-summary-info/ec-summary-info.vue
```diff
@@ -0,0 +1,131 @@
...
+</script>
+
+<style>
+:root {
+  --ec-summary-info--width: 184px;
```
wolandec, line 98: width of SummaryInfo should be set by user(parent component). We shouldn't use this max-width inside component.

src/components/ec-summary-info/ec-summary-info.vue
```diff
@@ -0,0 +1,131 @@
...
+          >
+            {{ item.text }}
+          </span>
+
+          <span
+            v-if="item.iconName"
+            v-ec-tooltip="{
+              content: item.tooltipText
+            }"
+            :class="{'tw-ml-4': item.iconName}"
```
wolandec, line 50: Why we use condition here as it is already on the line 46

src/components/ec-summary-info/ec-summary-info.vue
```diff
@@ -0,0 +1,131 @@
...
+          >
+            <ec-icon
+              class="ec-summary-info__content-line-item-icon"
+              :class="[item.iconCssClasses]"
+              :name="item.iconName"
```
wolandec, line 55: I'm not sure we should support this iconName here for now, maybe later if it will be needed, now I can't find any other icon except `i` icon.

src/components/ec-summary-info/ec-summary-info.vue
```diff
@@ -0,0 +1,131 @@
...
+        <div
+          v-for="item in lineItems"
+          :key="item.text"
+          :class="[item.cssClasses]"
+          class="ec-summary-info__content-line-item"
+        >
+          <span
+            :class="[item.textCssClasses]"
+          >
+            {{ item.text }}
```
wolandec, line 42: If we look at figma design on bene page, we'll see a `...`  in places where text is too long. We should support this behavior, and support it for the scenario with having a tooltip too.

src/components/ec-summary-info/ec-summary-info.vue
```diff
@@ -0,0 +1,131 @@
...
+          </span>
+
+          <span
```
wolandec, line 45: Do we need this span here and couldn't it be enough using only ec-icon?

src/components/ec-summary-info/ec-summary-info.vue
```diff
@@ -0,0 +1,131 @@
...
+      <div
+        v-if="lineItems.length > 0 || hasSlot('default')"
+        data-test="ec-summary-info__content-lines"
+        class="ec-summary-info__content-lines"
+      >
+
+        <div
+          v-for="item in lineItems"
+          :key="item.text"
+          :class="[item.cssClasses]"
```
mcibique, line 36: I think we don't need to wrap the `cssClasses` in the array because it is already an array.

I see the same problem on other lines in this component, could you please fix it everywhere?

src/components/ec-summary-info/ec-summary-info.vue
```diff
@@ -0,0 +1,131 @@
...
+          class="ec-summary-info__content-line-item"
+        >
+          <span
```
wolandec, line 39: https://www.figma.com/file/um1wKnapCkOe0BKkzIgXAY/Add-Benef?node-id=758%3A44993&t=Hr8qrbtfJmXIhUTh-4 regarding this we should provide title attribute here with the `item.text` value

src/components/ec-summary-info/ec-summary-info.vue
```diff
@@ -0,0 +1,131 @@
...
+        v-if="lineItems.length > 0 || hasSlot('default')"
+        data-test="ec-summary-info__content-lines"
+        class="ec-summary-info__content-lines"
+      >
+
+        <div
+          v-for="item in lineItems"
+          :key="item.text"
+          :class="[item.cssClasses]"
+          class="ec-summary-info__content-line-item"
```
mcibique, line 37: This class doesn't exist in the `<style>`, it can be removed.

mcibique:
We need to align the icon and the content, they don't have the same size:
Screenshot from 2022-11-17 11-46-00

Would it be also possible to wrap the stories in a div with a padding and center the summary to the middle? I'd also use an icon in stories that has some height because it's hard to tell if it's centered to the first line or not, e.g. the payment icon

src/components/ec-summary-info/ec-summary-info.story.js
DisegnAle: ok, corrected

src/components/ec-summary-info/ec-summary-info.story.js
DisegnAle: ok, corrected

src/components/ec-summary-info/ec-summary-info.story.js
DisegnAle: ok, corrected

src/components/ec-summary-info/ec-summary-info.vue
DisegnAle: you are right, corrected

src/components/ec-summary-info/ec-summary-info.story.js
DisegnAle: ok, corrected. thanks

src/components/ec-summary-info/ec-summary-info.vue
DisegnAle: ok, corrected. thanks

wolandec:
> We need to align the icon and the content, they don't have the same size: Screenshot from 2022-11-17 11-46-00
> 
> Would it be also possible to wrap the stories in a div with a padding and center the summary to the middle? I'd also use an icon in stories that has some height because it's hard to tell if it's centered to the first line or not, e.g. the payment icon

https://eburydev.slack.com/archives/CMBHDRU0Z/p1668696529098419?thread_ts=1668688619.406719&cid=CMBHDRU0Z this is an intentional shift

src/components/ec-summary-info/ec-summary-info.vue
DisegnAle: corrected, thanks

src/components/ec-summary-info/ec-summary-info.vue
DisegnAle: corrected, thanks

src/components/ec-summary-info/ec-summary-info.vue
DisegnAle: corrected, thanks

src/components/ec-summary-info/ec-summary-info.vue
DisegnAle: done, thanks

src/components/ec-summary-info/ec-summary-info.vue
DisegnAle: removed the cssClasses property

src/components/ec-summary-info/ec-summary-info.vue
DisegnAle: now i am using it, thanks for pointing it out though

DisegnAle: now i am using it, thanks for pointing it out though

src/components/ec-summary-info/ec-summary-info.vue
DisegnAle: if i remove it, it messes up the icon style

src/components/ec-summary-info/ec-summary-info.vue
DisegnAle: ok, changed it

package.json
```diff
@@ -25,7 +25,7 @@
     "test:watch": "jest --watch",
     "test:visual": "cypress open",
     "test:visual:ci": "npm run start & wait-on http://localhost:42475/ && npm run test:visual:all",
-    "test:visual:all": "cypress run --browser chrome --headless --config video=false,videosFolder=tests/integration/videos/chrome/,trashAssetsBeforeRuns=true && cypress run --browser firefox --headless --config video=false,videosFolder=tests/integration/videos/firefox/,trashAssetsBeforeRuns=false",
+    "test:visual:all": "cypress run --browser chrome -e storyIdFilter=full-screen-overlay --headless --config video=false,videosFolder=tests/integration/videos/chrome/,trashAssetsBeforeRuns=true && cypress run --browser firefox -e storyIdFilter=full-screen-overlay --headless --config video=false,videosFolder=tests/integration/videos/firefox/,trashAssetsBeforeRuns=false",
```
wolandec, line 28: Why?

DisegnAle: this changes have been reverted

src/components/ec-summary-info/ec-summary-info.spec.js
```diff
@@ -0,0 +1,191 @@
...
+      expect(wrapper.findByDataTest('ec-summary-info__main-icon').exists()).toBe(false);
+    });
+  });
+
+  describe('lineItems', () => {
+    it('should render with the given "lineItems" prop', () => {
+      const wrapper = mountSummaryInfo({
+        lineItems: [
+          {
+            stylePreset: 'title',
```
wolandec, line 34: Let's do this stylePreset optional, and by default it will be 
`label` for the 0 line, 
`text` for 1 and 
'description' for 2
and 'text' that lines after 2

and it is better to name that stylePresets as here https://www.figma.com/file/qLv2in1OHCrSOEoNivIQA1/Guidelines?node-id=519%3A21849&t=EXGa1bcEESWoW1DH-4

src/components/ec-summary-info/ec-summary-info.spec.js
```diff
@@ -0,0 +1,191 @@
...
+            text: 'The title',
+          },
+          {
+            stylePreset: 'description',
+            text: 'Some text',
+            showIcon: true,
```
wolandec, line 40: Icon shows if tooltipText is exists, no need this property.

src/components/ec-summary-info/ec-summary-info.vue
```diff
@@ -0,0 +1,133 @@
...
+    class="ec-summary-info"
+    data-test="ec-summary-info"
+  >
+    <div
+      v-if="iconName"
+      class="tw-mr-12"
+    >
+      <ec-icon
+        class="ec-summary-info__main-icon"
+        :data-test="`ec-summary-info__main-icon ec-summary-info__main-icon__${iconName}`"
```
wolandec, line 12: we may delete that data-test id with iconName, because in those tests we will test Vue props passing functionality, which we shouldn't that is not our functionality.

src/components/ec-summary-info/ec-summary-info.vue
```diff
@@ -0,0 +1,133 @@
...
+        :name="iconName"
+        :size="12"
+      />
+    </div>
+
+    <slot>
+      <div class="ec-summary-info__content">
+        <div
+          v-if="lineItems.length > 0"
```
wolandec, line 21: what this div and condition for?

src/components/ec-summary-info/ec-summary-info.vue
```diff
@@ -0,0 +1,133 @@
...
+              :title="item.text"
+              class="ec-summary-info__content-line-item-text"
+              :data-test="`ec-summary-info__content-line-item-text ec-summary-info__content-line-item-text-${item.stylePreset}`"
+              :class="`ec-summary-info__content-line-item-text-${item.stylePreset}`"
+            >
+              {{ item.text }}
+            </span>
+
+            <span
+              v-if="item.showIcon"
```
wolandec, line 39:
```suggestion
              v-if="item.tooltipText"
```

src/components/ec-summary-info/ec-summary-info.spec.js
DisegnAle: ok corrected

src/components/ec-summary-info/ec-summary-info.spec.js
DisegnAle: ok, corrected, thanks

src/components/ec-summary-info/ec-summary-info.vue
DisegnAle: done, thanks

src/components/ec-summary-info/ec-summary-info.vue
```diff
@@ -0,0 +1,133 @@
...
+</template>
+
+<script setup>
+import VEcTooltip from '../../directives/ec-tooltip';
+import EcIcon from '../ec-icon';
+
+defineProps({
+  iconName: {
+    type: String,
+    default: null,
```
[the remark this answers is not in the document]
DisegnAle: ok, corrected, thanks

src/components/ec-summary-info/ec-summary-info.spec.js
```diff
@@ -0,0 +1,191 @@
...
+              {
+                stylePreset: 'description',
+                text: 'testing the description',
+              },
+            ],
+          });
+          expect(wrapper.findByDataTest('ec-summary-info__content-line-item-text-description').exists()).toBe(true);
+          expect(wrapper.findByDataTest('ec-summary-info__content-line-item-text-description').text()).toBe('testing the description');
+        });
+      });
```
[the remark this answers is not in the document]
DisegnAle: no we didn't

src/components/ec-summary-info/ec-summary-info.vue
```diff
@@ -0,0 +1,155 @@
...
+  },
+  lineItems: {
+    type: Array,
+    default: () => [],
+  },
+});
+
+function assignStylePresetAccordingToIndex(index) {
+  let stylePreset = 'title'; // index === 0
```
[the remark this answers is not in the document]
DisegnAle: in figma the other ones are also different. should i rename them as well?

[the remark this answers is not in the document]
DisegnAle: in figma the other ones are also different. should i rename them as well?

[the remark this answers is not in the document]
wolandec: You don't agree with code suggestion of with renaming that presets?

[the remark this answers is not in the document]
DisegnAle: yes. because, in my opinion the names we are using now are much more explanatory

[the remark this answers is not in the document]
DisegnAle: infact i agree with you about the suggestion. i am concerned about the names used in the figmas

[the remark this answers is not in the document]
wolandec: Well this is the designer's descition how to name it. Not sure we have something to discuss here, those names in figma are ok, for me.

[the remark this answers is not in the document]
DisegnAle: ok, renamed presets

src/components/ec-summary-info/ec-summary-info.vue
```diff
@@ -0,0 +1,153 @@
+<template>
+  <div
+    class="ec-summary-info"
+    data-test="ec-summary-info"
+  >
+    <div
```
stavros-tomas, line 6: Hi, 
1. I feel this might work without the div. Can you please try it?
2. The margin is it coming from the designers? I can see in their designs sometimes is 9, some times 9.333, sometimes 10. Probably this needs to be 8px the same like convert and pay.
3.  Do we need the data-test of the icon to change according to what icon we pass?

src/components/ec-summary-info/ec-summary-info.vue
```diff
@@ -0,0 +1,153 @@
+<template>
+  <div
+    class="ec-summary-info"
+    data-test="ec-summary-info"
```
stavros-tomas, line 4: Can we please change it to,

:data-test="$attrs['data-test'] ? `${$attrs['data-test']} ec-summary-info` : 'ec-summary-info"

We need to have it like that for every new component we create. It allows extra data-tests to be applied other than the  default one

src/components/ec-summary-info/ec-summary-info.vue
```diff
@@ -0,0 +1,153 @@
...
+      />
+    </div>
+
+    <slot>
+      <div
+        v-if="lineItems.length > 0"
+        class="ec-summary-info__content"
+      >
+        <div
+          data-test="ec-summary-info__content-lines"
```
stavros-tomas, line 24: Hi can we please double check we need this. 

We have a div `ec-summary-info__content` inside a div `ec-summary-info__content-lines`

Does the parent div add any value? Can we merge those 2?

src/components/ec-summary-info/ec-summary-info.vue
```diff
@@ -0,0 +1,153 @@
...
+            <span
+              :title="item.text"
+              class="ec-summary-info__content-line-item-content"
+              :data-test="`ec-summary-info__content-line-item-content ec-summary-info__content-line-item-content-${getStylePreset(item.stylePreset, index)}`"
+              :class="`ec-summary-info__content-line-item-content-${getStylePreset(item.stylePreset, index)}`"
+            >
+              {{ item.text }}
+            </span>
+
+            <span
```
stavros-tomas, line 40: Hi, can we please double check we need this span?

Probably we can go away with just 

```
            <ec-icon
              v-if="item.tooltipText"
              v-ec-tooltip="{
                content: item.tooltipText
              }"
              :class="`ec-summary-info__content-line-item-icon-${getStylePreset(item.stylePreset, index)}`"
              :data-test="`ec-summary-info__content-line-item-icon ec-summary-info__content-line-item-icon-${getStylePreset(item.stylePreset, index)}`"
              class="ec-summary-info__content-line-item-icon tw-ml-4"
              name="simple-info"
              :size="14"
            />
```

src/components/ec-summary-info/ec-summary-info.vue
```diff
@@ -0,0 +1,153 @@
...
+        class="ec-summary-info__main-icon"
+        :data-test="`ec-summary-info__main-icon ec-summary-info__main-icon__${iconName}`"
+        :name="iconName"
+        :size="16"
+      />
+    </div>
+
+    <slot>
+      <div
+        v-if="lineItems.length > 0"
```
stavros-tomas, line 20: Hi, 
I see 2 ways to use this component
* with slots, in this case this div will not render anyway even if we do not have this protection
* with passing lines. In this case this protection I feel is no needed because I cannot think of a scenario that we will not pass any lines. 

If we would like to have some kind of protection maybe we can add a prop validator?
Something like that but haven't tested it.
```  lineItems: {
    type: Array,
    default: () => [],
    validator(value) {
      return value.length > 0;
    },
  },

File: src/components/ec-summary-info/ec-summary-info.vue [hunk 30]
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  &__content-line-item {
+    line-height: 20px;
+    @apply tw-min-w-0;
+    @apply tw-flex;
+  }
+
+  &__content-line-item-content {
+    text-overflow: ellipsis;
```
stavros-tomas, line 120: You can replace all the following classes with `@apply tw-truncate`

src/components/ec-summary-info/ec-summary-info.vue
```diff
@@ -0,0 +1,153 @@
...
+    @apply tw-fill-gray-4;
+  }
+
+  &__content {
+    @apply tw-flex-grow;
+    @apply tw-w-full;
+  }
+
+  &__content-line-item {
+    line-height: 20px;
```
stavros-tomas, line 114: I would replace with the following
```
  &__content-line-item {
    @apply tw-flex tw-items-center;
  }

File: src/components/ec-summary-info/ec-summary-info.vue [hunk 32]
```diff
@@ -0,0 +1,153 @@
...
+    @apply tw-text-gray-3;
+    @apply tw-small-text;
+  }
+
+  &__content-line-item-content-description {
+    @apply tw-help-text;
+    @apply tw-text-gray-5;
+  }
+
+  &__content-line-item-icon {
```
stavros-tomas, line 140: Since we have flex lines, you can align at the parent,
We still though need to cover the scenario that all icons don't get squeezed. (i think now is broken for the helper icon)
so please add
```
  &__content-line-item-icon {
    @apply tw-flex-shrink-0;
  }
```

src/components/ec-summary-info/ec-summary-info.vue
```diff
@@ -0,0 +1,153 @@
...
+
+  &__content-line-item-content {
+    text-overflow: ellipsis;
+    white-space: nowrap;
+    overflow: hidden;
+  }
+
+  &__content-line-item-content-label {
+    @apply tw-text-gray-5;
+    @apply tw-small-text;
```
stavros-tomas, line 127: Possibly this needs to be 
`@apply tw-caption-text;` 
as per C&P

src/components/ec-summary-info/ec-summary-info.vue
DisegnAle: > 2. The margin is it coming from the designers? I can see in their designs sometimes is 9, some times 9.333, sometimes 10. Probably this needs to be 8px the same like convert and pay.

yes, it was on the figmas. I reduced it to 8.

DisegnAle: > 3. Do we need the data-test of the icon to change according to what icon we pass?
you are right, i removed the dynamic part

src/components/ec-summary-info/ec-summary-info.vue
DisegnAle: ok corrected

src/components/ec-summary-info/ec-summary-info.vue
DisegnAle: merged them, thanks

src/components/ec-summary-info/ec-summary-info.vue
DisegnAle: the validation would be nice but it would be triggered even when there's a slot. any idea how to solve this?

src/components/ec-summary-info/ec-summary-info.vue
DisegnAle: corrected, thanks

src/components/ec-summary-info/ec-summary-info.vue
DisegnAle: ok corrected

src/components/ec-summary-info/ec-summary-info.vue
DisegnAle: ok corrected, thanks

src/components/ec-summary-info/ec-summary-info.vue
DisegnAle: > 3. Do we need the data-test of the icon to change according to what icon we pass?
ok, corrected, thanks

src/components/ec-summary-info/ec-summary-info.vue
```diff
@@ -0,0 +1,135 @@
...
+
+function getStylePresetByIndex(index) {
+  let stylePreset = '';
+  if (index === 0) { stylePreset = 'label'; }
+
+  if (index === 1 || index > 2) { stylePreset = 'text'; }
+
+  if (index === 2) { stylePreset = 'description'; }
+  return stylePreset;
+}
```
wolandec, line 72:
```suggestion
function getStylePresetByIndex(index) {
  if (index === 0)
   return 'label'

  if (index === 1 || index > 2)
    return 'description';

  if (index === 2)
    return 'text';
}
```

wolandec (approved):

src/components/ec-summary-info/ec-summary-info.vue
```diff
@@ -0,0 +1,131 @@
...
+  &__main-icon {
+    @apply tw-align-top;
+    @apply tw-flex-grow-0;
+    @apply tw-fill-gray-4;
+    @apply tw-mr-8;
+    @apply tw-flex-shrink-0;
+  }
+
+  &__content {
+    @apply tw-flex-grow;
```
stavros-tomas, line 91: Hey @DisegnAle not sure, do we need those grow and width properties? Maybe is on a scenario I cannot think of.

I think though we do need overflow hidden here because ellipsis doesn't seem to work properly without it

```
  &__content {
    @apply tw-overflow-hidden;
  }

```

src/components/ec-summary-info/ec-summary-info.vue
```diff
@@ -0,0 +1,131 @@
...
+
+    <slot>
+      <div
+        data-test="ec-summary-info__content"
+        class="ec-summary-info__content"
+      >
+        <div
+          v-for="(item, index) in lineItems"
+          :key="item.text"
+          class="ec-summary-info__content-line-item"
```
stavros-tomas, line 22: Can we please add here a unique data-test. In theory we can pass multiple "descriptions", "text" etc per summaryinfo component.

If this happens we will not have a way in the tests to target them. Something like the following maybe
```
:data-test="`ec-summary-info__content-line-item-${index}`"
```

src/components/ec-summary-info/ec-summary-info.story.js
```diff
@@ -0,0 +1,106 @@
...
+          {
+            stylePreset: 'description',
+            text: 'A very very very long description',
+          },
+        ],
+      },
+    };
+  },
+  template: `
+  <div
```
stavros-tomas, line 97: OPTIONAL,
If you think is better, you can amend this template to this one, but you don't have to. Might help better to showcase the truncated text

```
  <div class="tw-mx-0 tw-my-auto tw-flex tw-justify-center tw-items-center tw-h-screen tw-w-screen">
    <div
      class="tw-grid tw-bg-gray-7 tw-flex-1"
      style="max-width: 616px;">
      <ec-summary-info
        class="tw-col-4"
        v-for="(item, index) in 3"
        v-bind="args"
        :key="index"
      />
    </div>
  </div>
```

PR #1371, opened 2022-11-28
[ONL-7052] feat:reposition flatpickr when is open and user scrolls

stavros-tomas:
In this PR:

When the datepicker is placed inside an element that overflows(scrollbar visible) AND when the user scrolls, the datepicker currently appears detached from the input. 

image

We need to:
* listen to the scroll event when datepicker is open
* debounce the event every few hundred ms
* upon event, reposition the datepicker
* remove the event when the datepicker closes


stavros-tomas:
Thank you @mcibique for your thoughts in this, please let me know if this is what you had in mind.

src/components/ec-datepicker/ec-datepicker.vue
```diff
@@ -193,6 +194,20 @@ function onBlur(evt) {
   }
 }
 
+const isTimeoutRunning = false;
+// In case the datepicker is open and the user scrolls we need to reposition the datepicker otherwise will appear detached from the input.
+/* istanbul ignore next */
```
stavros-tomas, line 199: Not sure why this is not respected.

mcibique (approved):

PR #1375, opened 2022-12-02
ONL-7032 Adds popover inject functionality

derrick-g:
This adds provide functionality to change the container for where the popover will be placed within, default is BODY if none stated, otherwise it will use the ref container provided.

### DIFF
--- a/src/components/ec-full-screen-overlay/ec-full-screen-overlay.vue
+++ b/src/components/ec-full-screen-overlay/ec-full-screen-overlay.vue
@@ -80,6 +80,14 @@ const { deactivate } = useFocusTrap(overlayContent, {
   allowOutsideClick: true,
 });
 
+function getFocusTrapContainer() {
+  return overlayContent.value;
+}
+
+defineExpose({
+  getFocusTrapContainer,
+});
+
 onUnmounted(() => {
   deactivate();
 });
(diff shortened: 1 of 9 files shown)


src/components/ec-full-screen-overlay/ec-full-screen-overlay.vue
```diff
@@ -80,6 +80,14 @@ const { deactivate } = useFocusTrap(overlayContent, {
   allowOutsideClick: true,
 });
 
+function getFocusTrapContainer() {
+  return overlayContent.value;
+}
+
+defineExpose({
+  getFocusTrapContainer,
+});
```
wolandec, line 89: Can't find where we use it

wolandec (approved):

src/components/ec-full-screen-overlay/ec-full-screen-overlay.vue
derrick-g: This will be used in EBO, where we will get the container in the `Ebo-full-screen-overlay.vue` file

wolandec: I see you added test, thank you :)

jeger-at (approved):

PR #1387, opened 2022-12-16
[ONL-7096] fix: ellipsis on menu

stavros-tomas:
Small change, specifies width to 100%. This gives context to the ellipsis and allows it to be triggered.

Before change
image

After change
image

ALSO
Small change on the name of the tooltip export

### DIFF
--- a/src/components/ec-menu/ec-menu.vue
+++ b/src/components/ec-menu/ec-menu.vue
@@ -80,12 +80,13 @@ function getLinkDataTest(link) {
 
   &__item {
     @apply tw-block;
-    @apply tw-min-w-full;
+    @apply tw-w-full;
 
     .ec-menu--horizontal &,
     .ec-menu--is-reversed-horizontal & {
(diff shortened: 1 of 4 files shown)


jeger-at (approved):

src/components/ec-menu/ec-menu.vue
```diff
@@ -80,12 +80,13 @@ function getLinkDataTest(link) {
 
   &__item {
```
wolandec, line 81: why don't we use here truncate tailwind class?

wolandec (approved):

src/components/ec-menu/ec-menu.vue
stavros-tomas: Not sure if this would ok, items can also be too elements on the same line(horizontal). Not sure if applying trancate to those would play well.

wolandec (approved):

PR #1390, opened 2022-12-16
[ONL-7097]-Add responsive design for ec-alert

corinaper:


wolandec:
We need to update story.ts to reflect a mobile view in there and update the snapshots.

src/components/ec-alert/ec-alert.story.ts
```diff
@@ -16,7 +17,17 @@ export default {
   },
 };
 
-const Template = args => ({
+const args = {
```
wolandec, line 20:
```suggestion
const args: AlertProps = {
```

src/styles/tools/ec-alert.css
```diff
@@ -15,12 +15,37 @@
     }
   }
 
+  .ec-alert__body {
+
+    @media (min-width: 640px) {
```
wolandec, line 20: why we use here this breakpoint and why it is not a tailwind breakpoint?

corinaper: 640px is the sm tailwind breakpoint, I couldn't use tw directly though as it didn't work, I think we don't have the node-modules for it

wolandec (approved):

jeger-at (approved):

PR #1421, opened 2023-01-24
ONL-7169 Adds iconType to inputField

derrick-g:
Exposes the iconType attribute to inputField allowing us to decide on the colour of the icon used.

### DIFF
--- a/src/components/ec-input-field/ec-input-field.story.ts
+++ b/src/components/ec-input-field/ec-input-field.story.ts
@@ -3,6 +3,7 @@ import { action } from '@storybook/addon-actions';
 import { ref, watchEffect } from 'vue';
 
 import { IconName } from '../ec-icon/iconNames';
+import { IconType } from '../ec-icon/types';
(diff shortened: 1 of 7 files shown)


wolandec:
I noticed that after moving some components to TS we lost the controls to play with component properties in Storybook, could you please add argTypes to the story or we could create a ticket to do it for all our components in TS?

wolandec (approved):

src/components/ec-input-field/ec-input-field.story.ts
```diff
@@ -14,23 +15,32 @@ export default {
...
       control: { type: 'select' },
     },
+    iconType: {
+      options: IconType,
+      control: { type: 'select' },
+    },
+    icon: {
+      options: IconName,
+      control: { type: 'select' },
+    },
```
wolandec, line 25: inputType too

wolandec:
Ok from my side

PR #1442, opened 2023-02-22
ONL-7171: Added the new option card, tests and visual regression tests

MaruaEl:
A new component is added called Option Card. See the guidelines here: https://www.figma.com/file/qLv2in1OHCrSOEoNivIQA1/Guidelines?node-id=10%3A5188&t=bbJwWHDtlJeQFm2p-0

image
It has 4 variants: Icon and Title, Title and caption, Icon and title only, Title only. The only mandatory one is the title. 
It has 3 states: Default, Accent, and Danger. 
The Option card can be disabled and the styles change in base of that also.
The Option card can be a `<button>`, a `<a href>` and a `<router-link>`;
The Option card is dumb in terms of loading states. 
The Option card will emit to the parent a click event or a submit event if the type is button/submit button.
 
option-card

ERI: https://drive.google.com/file/d/1lJj7nNTMSj2VlmLSH_KuNEOOhCU-DeQ8/view?usp=sharing


src/components/ec-option-card/ec-option-card.vue
```diff
@@ -0,0 +1,199 @@
...
+
+const attrs = useAttrs();
+
+interface OptionCardProps {
+  isDisabled?: boolean,
+  isSubmit?: boolean,
+  optionTitle: string,
+  optionCaption?: string,
+  optionCardIconName?: IconName,
+  optionCardType?: OptionCardType,
```
wolandec, line 53: I believe we can remove all prefixies 'option'

src/components/ec-option-card/ec-option-card.vue
```diff
@@ -0,0 +1,199 @@
...
+    }"
+    class="ec-option-card"
+    :class="getOptionCardClasses()"
+    @click="setClickAction()"
+  >
+    <span
+      data-test="ec-option-card__title"
+    >
+      <ec-icon
+        v-if="optionCardIconName && optionTitle"
```
wolandec, line 17: why we check `optionTitle` here?

src/components/ec-option-card/ec-option-card.vue
```diff
@@ -0,0 +1,199 @@
...
+        data-test="ec-option-card__title-img"
+        class="ec-option-card__title-img"
+        :class="getOptionIconClass()"
+        :name="optionCardIconName"
+        :alt="optionTitle"
+        :size="15"
```
wolandec, line 23: Should it be 16?

src/components/ec-option-card/ec-option-card.vue
```diff
@@ -0,0 +1,199 @@
...
+      />{{ optionTitle }}</span>
+    <p
+      v-if="optionCaption"
```
wolandec, line 26: do we have a variant with no caption?

src/components/ec-option-card/ec-option-card.vue
```diff
@@ -0,0 +1,199 @@
...
+    &--danger {
+      @apply tw-fill-error;
+    }
+
+    &--disabled {
+      @apply tw-fill-gray-4;
+    }
+  }
+
+  &__caption {
```
wolandec, line 183: image
We should clarify how to deal with long strings in caption. As well as in title too.

src/components/ec-option-card/ec-option-card.vue
wolandec, line 53:
```suggestion
  title: string,
  caption?: string,
  iconName?: IconName,
  type?: OptionCardType,
```

src/components/ec-option-card/ec-option-card.vue
```diff
@@ -0,0 +1,199 @@
...
+  const newProps = {
+    ...isOptionCardDisabled.value,
+    ...(componentTag.value === 'button' && { type: optionCardButtonType.value }),
+    ...optionCardNavigationType.value,
+  };
+
+  return newProps;
+});
+
+const emit = defineEmits(['click', 'submit']);
```
wolandec, line 96: could you please have a look how emits are defined in ec-input ? We have to define it using types.

src/components/ec-option-card/ec-option-card.vue
```diff
@@ -0,0 +1,199 @@
...
+});
+
+const isOptionCardDisabled = computed(() => ({ disabled: isDisabled?.value }));
+
+const optionCardButtonType = computed(() => (isSubmit?.value ? 'submit' : 'button'));
+
+const optionCardNavigationType = computed(() => (to?.value ? { to: to.value } : { href: href?.value }));
+
+const componentProps = computed(() => {
+  const newProps = {
```
wolandec, line 87: This variable is redundant

src/components/ec-option-card/ec-option-card.vue
```diff
@@ -0,0 +1,199 @@
+<template>
+  <component
```
wolandec, line 2: image
Whad do you think? This way we could type properly our components are used in template. And code should become more structured.

src/components/ec-option-card/ec-option-card.vue
MaruaEl: Because the icon is ever only visible with a title and we can have a title without an icon. This check prevents people from pass an icon but not a title, this would avoid that scenario.

src/components/ec-option-card/ec-option-card.vue
MaruaEl: Why 16?
I put 15 because the design had 14.4px so I went for 14 because 13 was too small. 
image

src/components/ec-option-card/ec-option-card.vue
MaruaEl: Yes, Only title. 
image
https://www.figma.com/file/qLv2in1OHCrSOEoNivIQA1/Guidelines?node-id=893%3A22867&t=bbJwWHDtlJeQFm2p-0

src/components/ec-option-card/ec-option-card.vue
wolandec: I see 15 now, but we have no such size of icon for now. Maybe we could clarify this with Alba?

src/components/ec-option-card/ec-option-card.vue
MaruaEl: Oh yes, didn't think of that.... I'll ask Alba in the slack thread...

src/components/ec-option-card/ec-option-card.vue
MaruaEl: Like this? 
```
const emit = defineEmits<{(e: 'update:modelValue', value: InputFieldEvents[InputFieldEvent.UPDATE_MODEL_VALUE]): void
  (e: 'icon-click', value: InputFieldEvents[InputFieldEvent.ICON_CLICK]): void
}>();
```
so in this case it would be 

```
const emit = defineEmits<{(e: 'click'): void
  (e: 'submit', value: ButtonValueEvent): void
}>();

```
`ButtonValueEvent` I'm not sure what it could be probably a string...

src/components/ec-option-card/ec-option-card.vue
MaruaEl: You mean this? https://vuejs.org/guide/components/provide-inject.html#provide

wolandec: Yep

src/components/ec-option-card/ec-option-card.vue
wolandec: Yes, even more, I think we could remove submit.
```
const emit = defineEmits<{(e: 'click'): void}>();
```

src/components/ec-option-card/ec-option-card.vue
MaruaEl: the only one I'm not sure about is`type?: OptionCardType` as there is also the icon type - and I feel is too confusing to just have a type here. I would prefer to be more specific to avoid confusion with `optionCardType?: OptionCardType`

wolandec: Sorry, I can't see icon type prop. BTW if we want to adjust icon, then we could introduce `iconProps?: IconProps` instead of different props `iconName` and `iconType`

src/components/ec-option-card/ec-option-card.vue
MaruaEl: Spoke with Alba - should be actually 24px... so will set it to that and fix the alignments. Apparently we've been exporting all the icons wrong from Figma... Alba will explain more...

src/components/ec-option-card/ec-option-card.vue
MaruaEl: > _Sorry, I can't see icon type prop._ 

I meant in general and becuase we use an Icon here in the Option card - I didn't want the developer to get confused and think is related to the icon. 

```
// icon.vue

interface IconProps {
  name: IconName,
  size?: number,
  type?: IconType,
}
```

> _we could introduce iconProps?: IconProps instead of different props iconName and iconType_

Oh for this I think is fine because we only pass the icon name - we don't set an icon type here.

wolandec: I think it's fine to have just `type`, because we have no iconType here in this component. Moreover we see a type of that type prop - `OptionCardType`

src/components/ec-option-card/ec-option-card.vue
```diff
@@ -0,0 +1,196 @@
...
+
+const optionCardNavigationType = computed(() => (to?.value ? { to: to.value } : { href: href?.value }));
+
+const componentProps = computed(() => ({
+  ...isOptionCardDisabled.value,
+  ...(componentTag.value === 'button' && { type: componentTag.value }),
+  ...optionCardNavigationType.value,
+}));
+
+function getOptionCardClasses() {
```
wolandec, line 90: it should be `computed` not a function

src/components/ec-option-card/ec-option-card.vue
```diff
@@ -0,0 +1,196 @@
...
+  return {
+    'ec-option-card--disabled': Boolean(isDisabled?.value),
+    'ec-option-card--accent': type?.value === OptionCardType.OPTION_CARD_ACCENT,
+    'ec-option-card--danger': type?.value === OptionCardType.OPTION_CARD_DANGER,
+  };
+}
+
+function getOptionIconClass() {
```
wolandec, line 98: This too

src/components/ec-option-card/ec-option-card.vue
```diff
@@ -0,0 +1,196 @@
...
+
+function getcaptionClass() {
+  return {
+    'ec-option-card__caption--disabled': Boolean(isDisabled?.value),
+    'ec-option-card__caption--accent': type?.value === OptionCardType.OPTION_CARD_ACCENT,
+    'ec-option-card__caption--danger': !isDisabled?.value && (type?.value === OptionCardType.OPTION_CARD_DANGER),
+  };
+}
+
+function setClickAction() {
```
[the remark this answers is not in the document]
wolandec: If you want to create and return function callback here, then name could be `getClickHandlerFunction`

src/components/ec-option-card/ec-option-card.vue
```diff
@@ -0,0 +1,196 @@
...
+  if (componentTag.value === 'button' && !isDisabled?.value) {
+    return emit(OptionCardEvent.OPTION_CARD_CLICK);
+  }
+  return null;
+}
+</script>
+
+<style>
+:root {
+  --ec-option-card-width: 304px;
```
wolandec, line 124: card should be flexible and have no fixed size, please have a look at `ec-summary-info` for example, it should use space provided by upper level html element.

src/components/ec-option-card/ec-option-card.vue
wolandec: How Alba said we deal with this situation?

src/components/ec-option-card/types.ts
```diff
@@ -0,0 +1,29 @@
...
+  caption?: string,
+  iconName?: IconName,
+  type?: OptionCardType,
+  to?: string,
+  href?: string,
+}
+
+export enum OptionCardType {
+  OPTION_CARD_ACCENT = 'accent',
+  OPTION_CARD_DANGER = 'danger',
```
wolandec, line 16:
```suggestion
  ACCENT = 'accent',
  DANGER = 'danger',
```

src/components/ec-option-card/types.ts
```diff
@@ -0,0 +1,29 @@
...
+}
+
+export enum OptionCardEvent {
+  OPTION_CARD_CLICK = 'click'
```
wolandec, line 20:
```suggestion
  CLICK = 'click'
```

src/components/ec-option-card/types.ts
```diff
@@ -0,0 +1,29 @@
+import type { IconName } from '../ec-icon/types';
+
+export interface OptionCardProps {
+  isDisabled?: boolean,
+  isSubmit?: boolean,
```
wolandec, line 5: I can't see this prop in vue file, what it is for? The interfaces should be the same.

src/components/ec-option-card/types.ts
```diff
@@ -0,0 +1,29 @@
...
+  OPTION_CARD_CLICK = 'click'
+}
+
+export interface OptionCardEvents {
+  [OptionCardEvent.OPTION_CARD_CLICK]: undefined
+}
+
+export interface OptionCardTypeEmits {
+  (click: OptionCardEvent.OPTION_CARD_CLICK): void
+}
```
wolandec, line 29:
```suggestion
```

src/components/ec-option-card/types.ts
MaruaEl: sorry forgot to remove it

src/components/ec-option-card/ec-option-card.vue
MaruaEl: Alba said to make the box height grow and it doesn't matter if the Option cards have a different height.

src/components/ec-option-card/ec-option-card.vue
MaruaEl: Hey Anton, sorry can you explain this a bit more?
Why should be flexible if in the design has a fixed width? Speaking with Alba she confirmed the width will always be 304px. 
The only thing will need (in the feature) is to make it mobile friendly. Depending on the page container of course, usually a small screen is 320px so this should be fine.
Are you saying is better practice to have the parent set the width every-time? 
What's the benefit of that if we know that it must be 304px?  
Thank you!

wolandec: Hey, yes I'm saying that parent should set the width and element use all space available or needed for it. This is the best practice to have components adjustable by place of using if it's needed. And having fixed size of element is error prone approach, as we could have cases when it could break the layout.

src/components/ec-option-card/ec-option-card.vue
```diff
@@ -0,0 +1,196 @@
...
+const componentTag = computed(() => {
+  if (to?.value) {
+    return 'router-link';
+  } if (href?.value) {
+    return 'a';
+  }
+  return 'button';
+});
+
+const isOptionCardDisabled = computed(() => ({ disabled: isDisabled?.value }));
```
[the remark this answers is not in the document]
MaruaEl: I'm not a great expert in test coverage - it say all these lines are not covered: `72-74,81,83,87-89,93-95,99-101,105`
Maybe snapshot tests will help as I didn't put too many....  not sure - just a thought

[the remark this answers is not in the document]
wolandec: Let's start with removing of what I suggested :)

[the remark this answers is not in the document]
MaruaEl: Oh, I did already but is still not 100%?

src/components/ec-option-card/ec-option-card.spec.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    expect(wrapper.findByDataTest('ec-option-card').attributes('disabled')).toBeDefined();
+  });
+
+  it('renders the Option Card disabled (anchor) if the isDisabled prop is set to true', () => {
+    const wrapper = mountEcOptionCard({
+      isDisabled: true, title: 'Test title', href: '/test-href',
+    });
+
+    expect(wrapper.findByDataTest('ec-option-card').attributes('disabled')).toBe('true');
```
wolandec, line 101: do router link or href have disabled attributes or prop?

src/components/ec-option-card/ec-option-card.story.ts
```diff
@@ -0,0 +1,114 @@
+import { IconName } from '../ec-icon/iconNames';
+import EcOptionCard from './ec-option-card.vue';
+import type { OptionCardProps } from './types';
+
+export default {
+  title: 'Option Card',
+  component: EcOptionCard,
+};
```
wolandec, line 8: could you please add here argTypes similar as done in ec-input-field for the default story. This way we'll have correct controls in the storybook for option-card component.

src/components/ec-option-card/ec-option-card.vue
```diff
@@ -0,0 +1,187 @@
...
+    emit(OptionCardEvent.CLICK);
+  }
+}
+/* c8 ignore end */
+</script>
+
+<style>
+:root {
+  --ec-option-card-max-height: min-content;
+  --ec-option-card-min-height: 56px;
```
wolandec, line 118:
```suggestion
```
This min-height should be defined by min-content and paddings. For example if you have one line, only title with line height 24px + paddint-top 16 + padding-bottom 16, you'll get 56px of height.

src/components/ec-option-card/ec-option-card.vue
```diff
@@ -0,0 +1,187 @@
...
+  if (componentTag.value === 'button' && !isDisabled?.value) {
+    emit(OptionCardEvent.CLICK);
+  }
+}
+/* c8 ignore end */
+</script>
+
+<style>
+:root {
+  --ec-option-card-max-height: min-content;
```
wolandec, line 117: this could be removed too

src/components/ec-option-card/ec-option-card.vue
```diff
@@ -0,0 +1,187 @@
...
+}
+
+.ec-option-card {
+  @apply tw-flex tw-flex-col;
+  @apply tw-p-16;
+  @apply tw-border tw-border-solid tw-rounded tw-border-gray-6;
+  @apply tw-text-left tw-body-text tw-text-gray-3;
+
+  max-height: var(--ec-option-card-max-height);
+  min-height: var(--ec-option-card-min-height);
```
wolandec, line 128:
```suggestion
```

jeger-at (approved):

PR #1444, opened 2023-03-01
[ONL-7270] Fix popover height calculation in ec-dropdown-search

wolandec:


src/components/ec-dropdown-search/ec-dropdown-search.vue
```diff
@@ -451,6 +447,10 @@ function updateScroll() {
   }
 }
 
+watch(() => itemElements.value, () => {
+  setOverflowHeight();
```
mcibique, line 451: I think we should check that the dropdown is open before calling the function.

stavros-tomas (approved):

jeger-at (approved):

PR #1458, opened 2023-04-06
ONL-7401: Added data-test for different states of ec-option-card

jeger-at:
New data tests are set based on disabled and type flag

### DIFF
--- a/src/components/ec-option-card/ec-option-card.vue
+++ b/src/components/ec-option-card/ec-option-card.vue
@@ -1,6 +1,6 @@
 <template>
   <div
-    :data-test="getOptionCardDataTest"
+    :data-test="dataTest"
     class="ec-option-card"
     :class="getOptionCardClasses"
   >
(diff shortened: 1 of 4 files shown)


src/components/ec-option-card/ec-option-card.vue
```diff
@@ -51,6 +51,17 @@ const {
   iconName,
 } = toRefs(props);
 
+const getOptionCardDataTest = computed(() => {
```
wolandec, line 54:
```suggestion
const dataTest = computed(() => {
```

src/components/ec-option-card/ec-option-card.vue
```diff
@@ -51,6 +51,17 @@ const {
...
+  let optionCardDataTest = 'ec-option-card';
+  if (props.isDisabled) {
+    optionCardDataTest += ' ec-option-card--disabled';
+  }
+  if (props.type) {
+    optionCardDataTest += ` ec-option-card--${props.type}`;
+  }
+  return optionCardDataTest;
```
wolandec, line 62:
```suggestion
  return [
    'ec-option-card', 
    props.isDisabled && 'ec-option-card'--disabled, 
    props.type && `ec-option-card--${props.type}`
    ].filter(Boolean).join(' ');
```
Just as variant, if you like it.

wolandec (approved):

src/components/ec-option-card/ec-option-card.vue
jeger-at: that's much better.. Didn't like my implementation too much to be honest.. Guess I'm out of practice with development :D

PR #1486, opened 2023-04-28
[ONL-7467] - Show minutes in ec-timer

DisegnAle:


src/components/ec-timer/ec-timer.story.js
```diff
@@ -37,3 +37,21 @@ stopped.args = {
...
+stoppedWithMinutes.parameters = {
+  visualRegressionTests: { disable: true },
+};
+
+export const showMinutes = Template.bind({});
+showMinutes.args = {
+  showMinutes: true,
+  seconds: 80,
+  isRunning: true,
+};
```
wolandec, line 57: shouldn't we disable regression tests for this story?

src/components/ec-timer/ec-timer.story.js
```diff
@@ -37,3 +37,21 @@ stopped.args = {
...
+export const stoppedWithMinutes = Template.bind({});
+stoppedWithMinutes.args = {
+  seconds: 30,
+  showMinutes: true,
+  isRunning: false,
+};
+
+stoppedWithMinutes.parameters = {
+  visualRegressionTests: { disable: true },
+};
```
wolandec, line 50: can we have `visualRegressionTests` for this story as we have for the `basic`?

src/utils/countdown.js
```diff
@@ -6,23 +6,40 @@ export default class Countdown {
...
-    if (this.secondsLeft <= 0) {
+    this.totalSeconds = this.seconds - this.timeDifference;
+    this.minutesLeft = null;
+    this.secondsLeft = this.totalSeconds;
+
+    if (showMinutes) {
+      this.reduceMinutesLeft(this.totalSeconds);
+    }
+
+    this.emit('time-updated', {
```
wolandec, line 36: this logic is a presentation logic and it shouldn't be here. As earlier we just countdown seconds and emit an event about that. A user of this Countdown class will represent seconds in the way they want with of without minutes.

mcibique: Agree, it's also a breaking change - we should avoid those unless is really necessary. We use this util outside of ec-timer in EBO, we have to be cautious when changing the API of the component.

wolandec: https://fxsolutions.atlassian.net/browse/ONL-7483 also created a task and a proposal to get rid of Countdown class and replace it with composable with no mitt dependency.

mcibique: I think we already have one: https://fxsolutions.atlassian.net/browse/ONL-6889 proposing useEventBus.

wolandec: It's a proposal, we could discuss that, but I think we can remove eventBus behaviour and pass callback functions instead. It's more TS safety and simpler way. Something like this

```
function useCountdown(
{ 
 initialValue, 
 onUpdate 
} : {
   initialValue: number, 
   onUpdate: (currentValue: number) => void
 }
) {

instead emit something, useCountdown will be just calling onUpdate(59) where 59 is a current value.

}
```

src/components/ec-timer/ec-timer.story.js
DisegnAle: why should we?

wolandec: I guess we disable tests for dynamically changed stories, and here we have timer running.

src/components/ec-timer/ec-timer.story.js
DisegnAle: @mcibique what is the logic to show/disable the visual regression tests for a specific story?

src/components/ec-timer/ec-timer.vue
```diff
@@ -78,18 +93,42 @@ const diameter = computed(() => radius.value * 2 + strokeWidth.value);
...
 const circumference = computed(() => 2 * Math.PI * radius.value);
 const steps = computed(() => circumference.value / props.seconds);
-const offset = computed(() => circumference.value + steps.value * secondsLeft.value);
+const offset = computed(() => circumference.value + steps.value * totalSecondsLeft.value);
 
 const radius = ref(24);
 const strokeWidth = ref(4);
-const secondsLeft = ref(props.seconds);
+const minutesLeft = ref(getMinutes(props.seconds));
+const secondsLeft = ref(getSeconds(props.seconds));
```
wolandec, line 101: these variables are computed. We operate only with totalSecondsLeft.

```suggestion
const minutesLeft = computed(()=>  props.showMinutes 
  ?  Math.floor(totalSecondsLeft.value / 60);
  : null);
const secondsLeft = computed(()=>  props.showMinutes
   ? totalSecondsLeft.value - Math.round(minutesLeft.value * 60);
  : totalSecondsLeft);
```

src/components/ec-timer/ec-timer.vue
```diff
@@ -78,18 +93,38 @@ const diameter = computed(() => radius.value * 2 + strokeWidth.value);
...
+  }
+  if (props.showMinutes) {
+    return totalSecondsLeft.value - Math.round(minutesLeft.value * 60);
+  }
+  return totalSecondsLeft.value;
+});
+const totalSecondsLeft = ref(getTotalSecondsLeft(props.seconds));
 let countdown = null;
 
+function getTotalSecondsLeft(seconds) {
```
wolandec, line 119: why we need this function?

src/components/ec-timer/ec-timer.vue
```diff
@@ -78,18 +93,38 @@ const diameter = computed(() => radius.value * 2 + strokeWidth.value);
...
+const secondsLeft = computed(() => {
+  if (!props.isRunning) {
+    return props.seconds;
+  }
+  if (props.showMinutes) {
+    return totalSecondsLeft.value - Math.round(minutesLeft.value * 60);
+  }
+  return totalSecondsLeft.value;
+});
+const totalSecondsLeft = ref(getTotalSecondsLeft(props.seconds));
```
wolandec, line 116:
```suggestion
const totalSecondsLeft = ref(props.seconds);
```

src/components/ec-timer/ec-timer.vue
DisegnAle: cause we need to keep count of the totalSeconds that differ from seconds, when minutes are present to correctly give the offset a value

src/components/ec-timer/ec-timer.vue
DisegnAle: it breaks the logic when isRunning is false

src/components/ec-timer/ec-timer.vue
```diff
@@ -74,22 +89,40 @@ const props = defineProps({
...
+  if (props.isRunning) {
+    return circumference.value + steps.value * totalSecondsLeft.value;
+  }
+  return circumference.value;
+});
+
+const minutesLeft = computed(() => Math.floor(totalSecondsLeft.value / 60));
+
+const secondsLeft = computed(() => {
+  if (!props.isRunning) {
```
wolandec, line 109: what if timer is running at first, then isRunning is set to false, then we'll show wrong seconds and will show seconds from props, not seconds in minutes. We need to add a test for this case and let's rename this variable to minuteSecondsLeft

src/components/ec-timer/ec-timer.story.js
mcibique: Visual regression tests are very expensive to run in terms or generating pngs and comparing pngs, so we are trying to have them as small as possible and as less as possible.

If we can display all possible component states in one story (usually called `all`) then use that story for tests and ignore the rest, because it's duplicating what has been already tested (e.g. ec-input-field). If we don't have the `all` story, then we go with all other stories if they add any testing value (e.g. ec-timer).

I think the `ec-timer` doesn't have `all` story so we went with others but because `basic` and `stopped` stories renders the same png (animations are disabled while these tests are running so there's nothing more to test), we decided to save some time by ignoring the `stopped` story. If `stoppedWithMinutes` and `showMinutes` renders the same and we're sure that one of them covers all scenarios then we can keep only one of them.

src/components/ec-timer/ec-timer.vue
```diff
@@ -74,22 +89,40 @@ const props = defineProps({
 });
 const emit = defineEmits(['time-expired']);
 
-const diameter = computed(() => radius.value * 2 + strokeWidth.value);
+const radius = 24;
+const strokeWidth = 4;
+const diameter = computed(() => radius * 2 + strokeWidth);
```
mcibique, line 94: This doesn't have to be computed prop if there's nothing reactive used in it.

src/components/ec-timer/ec-timer.vue
```diff
@@ -74,22 +89,40 @@ const props = defineProps({
...
+});
+
+const minutesLeft = computed(() => Math.floor(totalSecondsLeft.value / 60));
+
+const secondsLeft = computed(() => {
+  if (!props.isRunning) {
+    return props.seconds;
+  }
+  if (props.showMinutes) {
+    return totalSecondsLeft.value - Math.round(minutesLeft.value * 60);
```
mcibique, line 113:
```suggestion
    return totalSecondsLeft.value % 60;
```

We could wrap it in `Math.round()` for additional safety but if it's guaranteed that totalSecondsLeft is always int we should safe.

src/components/ec-timer/ec-timer.vue
```diff
@@ -74,22 +89,40 @@ const props = defineProps({
 });
 const emit = defineEmits(['time-expired']);
 
-const diameter = computed(() => radius.value * 2 + strokeWidth.value);
+const radius = 24;
+const strokeWidth = 4;
+const diameter = computed(() => radius * 2 + strokeWidth);
 const viewbox = computed(() => `0 0 ${diameter.value} ${diameter.value}`);
-const circumference = computed(() => 2 * Math.PI * radius.value);
+const circumference = computed(() => 2 * Math.PI * radius);
```
mcibique, line 96: This doesn't have to be computed prop if there's nothing reactive used in it.

src/components/ec-timer/ec-timer.story.js
DisegnAle: ok, thank you @mcibique

src/components/ec-timer/ec-timer.vue
DisegnAle: if i remove the reactivity the blu circle stops to reduce itself

src/components/ec-timer/ec-timer.vue
DisegnAle: the same computed is used also to show only seconds (line 54). i would not rename it

wolandec: You have to use totalSecondsLeft on line 54

src/components/ec-timer/ec-timer.vue
mcibique: It's a multiplication of 3 constants so I can't have any effect. Did you clean up all `.value`s from anywhere? I tried this on my local and it worked for me (there's lots of things that can be removed):
```js
const radius = 24;
const strokeWidth = 4;
const diameter = radius * 2 + strokeWidth;
const viewbox = `0 0 ${diameter} ${diameter}`;
const circumference = 2 * Math.PI * radius;
const steps = computed(() => circumference / props.seconds);

const offset = computed(() => {
  if (props.isRunning) {
    return circumference + steps.value * totalSecondsLeft.value;
  }
  return circumference;
});

```

wolandec: Had you a TS error about that claimed about missing `value` property in `number` type?

src/components/ec-timer/ec-timer.vue
DisegnAle: now it should be ok

src/components/ec-timer/ec-timer.vue
```diff
@@ -74,22 +89,34 @@ const props = defineProps({
...
+  if (totalSecondsLeft.value === 0) {
+    return circumference;
+  }
+
+  return circumference + steps * totalSecondsLeft.value;
+});
+
+const minutesLeft = computed(() => Math.floor(totalSecondsLeft.value / 60));
+
+const minutesSecondsLeft = computed(() => totalSecondsLeft.value % 60);
```
wolandec, line 109:
```suggestion
const minuteSecondsLeft = computed(() => totalSecondsLeft.value % 60);
```

src/components/ec-timer/ec-timer.vue
DisegnAle: you mean in ebo?

DisegnAle: no, no errors

wolandec: No, I forgot that this component is not in TS yet :). Otherwise if you hadn't remove `.value` you'd have an error from TS that `circumReference` has no `value` prop, as it is type of `number` after removing `computed`.

DisegnAle: yes i meant i had not errors because the file is still in js

wolandec (approved):

PR #1489, opened 2023-05-03
[ONL-7403] Migrate: ec-main-container to TS

raulwwq0:
**Jira Ticket:** [ONL-7403](https://fxsolutions.atlassian.net/browse/ONL-7403)

Migrate ec-main-container to TS

[ONL-7403]: https://fxsolutions.atlassian.net/browse/ONL-7403?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/ec-main-container/ec-main-container.vue
```diff
@@ -43,23 +43,22 @@
...
-  title: {
-    type: String,
-    default: null,
-  },
-  titleIntro: {
-    type: String,
-    default: null,
-  },
+interface MainContainerProps {
+    title?: string | null,
```
wolandec, line 50: Do we need here null as title is already optional? The same for titleIntro

raulwwq0: yes if we want to keep the default value as `null`, otherwise is not necessary

wolandec: This is the question: do we need null here, what is the difference between null and undefined for those properties?

wolandec: Let's remove null here.

raulwwq0: In this case it seems there are no differences, so yep we can remove null default values for those properties

wolandec (approved):

PR #1496, opened 2023-05-05
[ONL-7455] Fix broken JSX stories

raulwwq0:
**Jira ticket:** [ONL-7455](https://fxsolutions.atlassian.net/browse/ONL-7455)

Install the  [`@vitejs/plugin-vue-jsx` package](https://github.com/vitejs/vite-plugin-vue/tree/main/packages/plugin-vue-jsx) to add JSX support to Vue and also change Vite config in `.storybook/main.js` to use this new plugin instead of the Storybook default one (React JSX).

[ONL-7455]: https://fxsolutions.atlassian.net/browse/ONL-7455?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -1,5 +1,6 @@
-// eslint-disable-next-line import/no-extraneous-dependencies
+/* eslint-disable import/no-extraneous-dependencies */
 const vueJsx = require('@vitejs/plugin-vue-jsx');
+const { mergeConfig } = require('vite');
 
 module.exports = {
   framework: '@storybook/vue3',
@@ -29,24 +30,30 @@ module.exports = {
       name: '@storybook/addon-postcss',
       options: {
         postcssLoaderOptions: {
-          // eslint-disable-next-line global-require,import/no-extraneous-dependencies
+          // eslint-disable-next-line global-require
           implementation: require('postcss'),
         },
       },
     },
   ],
   viteFinal: (config) => {
-    const pluginsWithoutReact = config.plugins.filter(plugin => !Array.isArray(plugin));
-    const vueJsxPlugin = vueJsx();
+    const reactPluginNameRegex = /^vite:react-/;
+    const reactPluginsArrayIndex = config.plugins.findIndex(plugin => Array.isArray(plugin) && plugin.every(p => p.name.match(reactPluginNameRegex)));
 
-    const plugins = [
-      ...pluginsWithoutReact,
-      vueJsxPlugin,
-    ];
+    if (reactPluginsArrayIndex === -1) {
+      throw new Error('Unable to patch plugins - cannot find vite:react plugins. Is this fix still necessary?');
+    }
 
-    return {
+    // remove react plugins and replace them with vue plugins
+    // see https://github.com/storybookjs/storybook/issues/21681
+    // remove after migrating to SB 7
+    const pluginsWithoutReact = config.plugins.filter((_, index) => index !== reactPluginsArrayIndex);
+    return mergeConfig({
       ...config,
-      plugins,
-    };
+      plugins: [
+        ...pluginsWithoutReact,
+        vueJsx(),
+      ],
+    });
   },
 };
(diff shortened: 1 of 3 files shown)


.storybook/main.js
```diff
@@ -32,4 +35,18 @@ module.exports = {
       },
     },
   ],
+  viteFinal: (config) => {
+    const pluginsWithoutReact = config.plugins.filter(plugin => !Array.isArray(plugin));
```
wolandec, line 39: Why we use here this isArray condition? Is there any documentation that all react plugins are placed in one array?

raulwwq0: I didn't find any official docs about it, but if you do a console.log of the plugins it return something like this:

```js
[
    {
          // vue-plugin
    },
    [
         {
             // react-plugin
         },
         // ...
    ],
    {
          // vue-plugin
    },
    // ...
]
```

That's why with isArray I can remove all unnecessary React plugins.

wolandec: Okay, let's add comment here why isArray is using to determine React plugins and use mergeConfig function in return statement.

mcibique: I believe this PR applies the workaround from this issue: https://github.com/storybookjs/storybook/issues/21681. I also find the solution easy to break if any other array is added or the structure of the plugins changes, we may break this fix and we will not notice that it no longer works or breaks new stuff. I'd rewrite this a bit like this:
```
    const reactPluginNameRegex = /^vite:react-/;
    const reactPluginsArrayIndex = config.plugins.findIndex(plugin => Array.isArray(plugin) && plugin.every(p => p.name.match(reactPluginNameRegex)));

    if (reactPluginsArrayIndex === -1) {
      throw new Error('Unable to patch plugins - cannot find vite:react plugins. Is this fix still necessary?');
    }

    // remove react plugins and replace them with vue plugins
    // see https://github.com/storybookjs/storybook/issues/21681
    // remove after migrating to SB 7
    config.plugins.splice(reactPluginsArrayIndex, 1, vueJsx());
    return config;
```

^^ This find the plugins by name, adds an exception when unable to find them. The issue also states that the problem has been removed with SB v7, I'd add a comment to remove it when migrating. WDYT?

mcibique:
There's also `vite.config.js` in the root folder where we don't have the JSX plugin. Do we need to patch that one too? Is that config actually used when SB has their own `viteFinal` section in their configuration?

.storybook/main.js
wolandec: Yep, that was my thought too. As I can understand only React plugins are in array, but anyway your solution is more sustainable. Also, I guess we avoid this problem and remove this fix once we are on 7 version.

Actually, I tried to migrate chameleon to storybook 7, jsx works with no hack, but I needed to make some adjustments. So we could stick with this fix for now, and have a task to do for later time to migrate to storybook 7.

raulwwq0: yes, this approach is better than mine because I didn't take into account the possibility of another array being added in the plugins :sweat_smile:, Should I implement this instead?

mcibique: We can quickly spawn an empty project with SB 7 and check the stories there if they work with JSX, and how the plugins array looks like there. Changelog for SB 7 says they rewritten the vite support and framework structure, introduced this new preset (vue3-vite: https://github.com/storybookjs/storybook/blob/next/code/frameworks/vue3-vite/src/preset.ts), so it's very likely it has been fixed.

wolandec: Yes, jsx fixed in 7 version.

raulwwq0:
Storybook seems to override the Vue JSX plugin if we place it in `vite.config.js` plugins, so it has to be placed in the `viteFinal` section after the React workaround.

.storybook/main.js
wolandec: I had a look another time and I can see that I used the same hack to filter react plugins for jsx :).

wolandec (approved):

mcibique (approved):

PR #1498, opened 2023-05-05
[ONL-7409] Migrate ec-btn to TS

wolandec:


src/components/ec-btn/ec-btn.vue
```diff
@@ -50,150 +50,115 @@
...
-    'ec-btn--full-width': isFullWidth.value,
-    'ec-btn--icon-only': icon.value && !hasDefaultSlot(),
-    [`ec-btn--${category.value}`]: category.value,
-    'ec-btn--outline': isOutline.value,
-    'ec-btn--primary-reverse': isReverse.value && category.value === 'primary',
-    'ec-btn--secondary-reverse': isReverse.value && category.value === 'secondary',
-    'ec-btn--success-reverse': isReverse.value && category.value === 'success',
-    'ec-btn--error-reverse': isReverse.value && category.value === 'error',
-    'ec-btn--warning-reverse': isReverse.value && category.value === 'warning',
+    'ec-btn--sm': props.size === 'sm',
```
stavros-tomas, line 150: Shall we use the enums here? `ButtonSize.Sm`, and few lines below `ButtonCategory.Primary` etc?

wolandec: Thank you! Sure we must!

src/components/ec-btn/ec-btn.spec.ts
```diff
@@ -1,26 +1,38 @@
...
+        props,
+        ...mountOpts,
+      },
+    ) as CVueWrapper;
   }
 
   it('should render a <button> element by default', () => {
-    const wrapper = mountBtn();
+    const wrapper = mountBtn({
+      category: ButtonCategory.Primary,
```
mcibique, line 24: We shouldn't change the tests while doing migrations. Why is this needed?

src/components/ec-btn/types.ts
```diff
@@ -0,0 +1,32 @@
+import type { RouteLocationRaw } from 'vue-router';
+
+import type { IconName } from '../ec-icon/iconNames';
+
+export enum ButtonSize {
+  Md = 'md',
```
mcibique, line 6: Enum names can use human readable string, rename it to Medium and Small. Keep the values in short form.

src/components/ec-btn/ec-btn.spec.ts
wolandec: Because category is a required prop.

mcibique: It wasn't required before, we should fix that then: https://github.com/Ebury/chameleon/blob/master/src/components/ec-btn/ec-btn.vue#L108-L113

wolandec: My bad, I'll fix it. Thanks!

stavros-tomas (approved):

derrick-g (approved):

PR #1508, opened 2023-05-11
[ONL-7117] Force kebab case filenames

wolandec:


.eslintrc.js
```diff
@@ -75,6 +75,10 @@ module.exports = {
         tsx: 'never',
       },
     ],
+    'filenames/match-regex': [
+      2,
+      '^[.a-z0-9][a-z0-9-]+(?:\\.[^.]+)?(?:\\.stub\\..*|\\.spec\\..*)?$',
```
mcibique, line 80: Sorry, I'm not sure I understand the regex. What are we checking for stub and spec? Wouldn't be `'^[.a-z0-9-]+$'` enough to check the kebab case?

wolandec: Yep, you are right, that's enough.

stavros-tomas (approved):

jeger-at (approved):

PR #1529, opened 2023-05-16
[ONL-7472] update ec-tooltips to behave correctly in the shadow root

corinaper:
+ ts migration


DisegnAle (approved):

src/directives/ec-tooltip/ec-tooltip.story.ts
```diff
@@ -11,56 +12,59 @@ export default {
...
+// @ts-ignore
 basic.args = {
   content: 'Test tooltip',
   shown: true,
-  placement: 'top',
+  placement: TooltipPlacement.TOP,
   delay: 100,
   popperClass: '',
-  trigger: 'hover',
+  triggers: TooltipTrigger.HOVER,
```
wolandec, line 58: Why it is plural now?

corinaper: it just matches the object key name used in the ec-tooltip, otherwise we'll have trigger inside the storybook, but we'll need to use triggers for the directive

derrick-g (approved):

stavros-tomas (approved):

jeger-at (approved):

PR #1548, opened 2023-05-18
[ONL-7553]-adjust inline svg to also allow use of a target element (inlineSvgSprites) instead of the body.

corinaper:


src/icons/browser.js
derrick-g, line 1: Can we convert this file to TS?

raulwwq0 (approved):

jeger-at (approved):

PR #1559, opened 2023-05-25
ONL-7576 Adds searchFields Prop to dropdown search

derrick-g:
This will allow us to specify the keys in an object we want to search against, it will default to the `text` key in the case nothing is found or if the `searchFields` are not defined. This allows for extensive potential to allow the user to search for what they can see in the dropdown.


src/components/ec-dropdown-search/ec-dropdown-search.vue
```diff
@@ -343,6 +347,19 @@ const filteredItems = computed(() => {
...
+        if (item[searchField]) {
+          const searchValue = removeDiacritics(item[searchField].trim().toLowerCase());
+          if (searchValue.includes(sanitisedText)) {
+            hasSearchValue = true;
+            break;
+          }
+        }
+      }
+      if (hasSearchValue) return hasSearchValue;
+    }
```
mcibique, line 362: The performance of this is getting worse and worse. We need to removeDiacritics from every field every time the user presses a key. We need some kind of cache here and calculate the searchable value only once per item.

We could have a separate computed prop which updates only when `items` changed, something like this:

```
const indexedItems = computed(() => {
	return new WeakMap(items.value.map(item => {
    return [item, makeIndex(item, props.searchFields ?? ['text'])]; // defaults to text field of the item if not passed
  }));
});
  
function makeIndex(item, searchFields) {
  return searchFields.map(searchField => {
    return removeDiacritics(item[searchField] ?? '').toLowerCase();
  }).join('\u00A0'); // all values have their values sanitised and concatenated into a single string separated by a value that user can't type on the keyboard.
}
```
Then we can use the cache to retrieve sanitised text of each item and search there:
```
const filteredItems = computed(() => {
  const sanitisedText = removeDiacritics(filterText.value.toLowerCase());
  if (!sanitisedText) {
    return props.items;
  }

  return props.items.filter((item) => {
    const indexedText = indexedItems.value.get(i);
    return indexedText.includes(sanitisedText);
  });
});
```

And this covers both scenarios when searchFields are set or not.

wolandec:
I'd consider more generic approach. Let's introduce matchFunction prop

```
 matchFunction: (searchText: string, item: unknown) => boolean
```
This function will be called inside component for every item if it is provided as a prop and user input anything in a search input.

Using new generic types for props are introduced in [Vue 3.3](https://vuejs.org/api/sfc-script-setup.html#generics)  we'll type item parameter of this function with generic T of item type.

```
<script setup lang="ts" generic="T">
defineProps<{
 items: T[]
 matchFunction: (searchText: string, item: T) => boolean
 ...the rest of the props
}>()
</script>
```

src/components/ec-dropdown-search/ec-dropdown-search.vue
```diff
@@ -336,15 +340,21 @@ const filterText = ref('');
 const isSearchInputFocused = ref(false);
 const searchInput = ref(null);
 
+const indexedItems = computed(() => new WeakMap(props.items.map(item => [item, makeIndex(item, props.searchFields ?? ['text'])])));
+
+function makeIndex(item, searchFields) {
```
wolandec, line 345:
```suggestion
function makeIndexText(item, searchFields) {
```

src/components/ec-dropdown/ec-dropdown.vue
```diff
@@ -75,6 +76,10 @@ const props = defineProps({
     type: Array,
     default: () => ([]),
   },
+  searchFields: {
+    type: Array,
+    default: null,
```
wolandec, line 81:
```suggestion
```

PR #1579, opened 2023-06-23
[ONL-7529] Update Vue dependencies

raulwwq0:


src/components/ec-popover/__snapshots__/ec-popover.spec.js.snap
```diff
@@ -1,81 +1,97 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`EcPopover component should add the z-index level class if the level was given 1`] = `
-<fv-dropdown-stub
-  ariaid="ec-popover-1"
-  arrowoverflow="true"
-  container="body"
-  data-test="ec-popover"
-  popperclass="ec-popover ec-popover--level-30"
```
wolandec, line 9: this `ec-popover--level-30` property should be set, but we can't see it in the new snapshot

src/components/ec-phone-number-input/ec-phone-number-input.vue
```diff
@@ -158,6 +158,8 @@ import { computed, ref } from 'vue';
 
 import type { Maybe, ZIndexLevel } from '../../../global';
 import useConfig from '../../composables/use-ec-config';
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
```
derrick-g, line 162: Do we need these ignores?

derrick-g (approved):

src/components/ec-phone-number-input/ec-phone-number-input.vue
raulwwq0: Yes because since directives are not use in the `<script>` we are getting a TS error that said we must import them as types. If we do that directives doesn't work, so in order to continue using them we have to ignore TS in those imports.

src/components/ec-popover/__snapshots__/ec-popover.spec.js.snap
raulwwq0: Fixed!

wolandec (approved):

src/components/ec-dropdown-search/ec-dropdown-search.vue
```diff
@@ -458,8 +458,9 @@ function updateScroll() {
   }
 }
 
-watch(() => itemElements.value, () => {
+watch(() => props.items, async () => {
```
mcibique, line 461: Sorry, but why is this fix needed?

src/directives/ec-amount/ec-amount.spec.js
```diff
@@ -48,17 +47,17 @@ describe('EcAmount', () => {
...
+    expect(spy.mock.calls).toContainEqual(['keydown', keydownHandler]);
   });
 
   it('should throw an error if the directive is not attached to an input', () => {
-    withMockedConsole((errorSpy, warnSpy) => {
-      expect(() => mountTemplate('<div v-ec-amount="{}"></div>')).toThrow(new TypeError('v-ec-amount requires 1 input'));
-      expect(warnSpy).toHaveBeenCalledTimes(1);
-      expect(warnSpy.mock.calls[0][0]).toContain('Unhandled error during execution of directive hook');
-    });
+    expect(() => mountTemplate('<div v-ec-amount="{}"></div>')).toThrow(new TypeError('v-ec-amount requires 1 input'));
```
mcibique, line 60: Out of curiosity, is this mentioned somewhere in the changelog why this has changed? is this a change in VTU or in Vue?

src/components/ec-dropdown-search/ec-dropdown-search.vue
raulwwq0: After the updates that watch was not covered by the tests, so after writing some tests for that Anton and I discovered that the watch was not working as it should. We tested manually and found that `itemsElements.value` was not changing and therefore the watch was not being triggered. Watching `props.items.value` instead (and also waiting for `nextTick` before setting the new height) solve the issue.

src/directives/ec-amount/ec-amount.spec.js
raulwwq0: I couldn´t find anything about that, but I would say is a change in Vue itself because it only happens with those unhandled errors, other errors such as [this one](https://github.com/Ebury/chameleon/pull/1579/commits/d1cdf3e03c59771fe4808f506df1aadf733e2674#diff-990b8ff5535146e45bbe39346c6918d91dd1d03380f7fbf22744bf55f4e784b5L160) are being printed in the console, so VTU is not really involved on that

jeger-at (approved):

PR #1597, opened 2023-07-06
[ONL-7579] Refactor defineOptions and defineEmits

raulwwq0:
There is an issue with Storybook because they are still using Vue 3.2.47, so the new `defineEmits` syntax is not being recognized by the controls and therefore we can't modify emits there because the events section is not being displayed.


mcibique (approved):

wolandec (approved):

wolandec:
We need to track storybook and update it when they bump Vue version to 3.3 to have events automatically generated in stories 
image

PR #1627, opened 2023-08-05
[ONL 7731] feat: new tag component

stavros-tomas:
New tag component


src/components/ec-tag/ec-tag.vue
```diff
@@ -0,0 +1,61 @@
...
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+import vEcTooltip from '../../directives/ec-tooltip';
+import EcIcon from '../ec-icon';
+import type { IconName, IconType } from '../ec-icon/types';
+
+interface TagProps {
+  text: string,
+  iconName?: IconName,
+  iconType?: IconType,
```
wolandec, line 35: Lets replace these two with one prop `iconProps: IconProps`?

stavros-tomas: Not sure about it, we typically have it separate. Some examples [input-field](https://github.com/Ebury/chameleon/blob/master/src/components/ec-input-field/ec-input-field.vue#L122), [navigation-link](https://github.com/Ebury/chameleon/blob/master/src/components/ec-navigation-link/ec-navigation-link.vue#L78) etc. Would you be ok to keep them as is?

wolandec: Yes, it could be refactored in the future.

Now any time we change icon props we should go and change the same props here as well. For example we change the name of `IconType` prop to `IconKind` in `ec-icon`, we need to change it in template and in this prop declaration too. But what is the sense of these movements if we could just bind `iconProps` to `ec-icon` in this `ec-tag` component?

Another case if we want to force user of this component to use only two props from IconProps - then this approach with distinct props should be used. If it is true - then lets stick with initial solution using separate props.

stavros-tomas: In many cases we do not care about type, we use only name. Size is also hardcoded this should not change. That's why I think we have the approach we have, if all were required all the time, then yes another perspective :)

src/components/ec-tag/ec-tag.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+          mocks: {
+            vEcTooltip: EcTooltipDirectiveMock,
+          },
+        },
+      },
+    ) as CVueWrapper;
+  }
+
+  describe(':props', () => {
+    it('should throw if no text props is given', () => {
```
wolandec, line 28: we can skip this test as we have TS which check required props for us

tests/integration/snapshots/visual-regression.spec.js/tag--basic__chrome.snap.png
wolandec, line 1: lets add story for a truncated text

wolandec (approved):

PR #1640, opened 2023-08-11
[ONL-7806] EC Inline Input  typescript and changes

derrick-g:
Adds typescript support for ex-inline-input
- Makes the labels Sentence case
- Allows the edit/copy icon to be aligned either right aligned (default) or spaced next to the text
- Adds title attribute for text to be visible if too long when hover


src/components/ec-inline-input-field/components/copy/copy.vue
```diff
@@ -12,101 +13,77 @@
...
-    type: String,
-    required: true,
-  },
-  isSensitive: {
-    type: Boolean,
-    default: false,
-  },
-});
+const isCopied = ref(false);
+const showTooltip = ref(false);
```
wolandec, line 49:
```suggestion
const isTooltipShown = ref(false);
```

src/components/ec-inline-input-field/ec-inline-input-field.spec.ts
```diff
@@ -69,26 +97,34 @@ describe('EcInlineInputField', () => {
           expect(wrapper.emitted('edit')).toBeUndefined();
           await wrapper.findByDataTest('ec-inline-input-field-value-text__action').trigger('click');
 
-          expect(wrapper.emitted('edit').length).toBe(1);
+          expect(wrapper.emitted('edit')?.length).toBe(1);
```
wolandec, line 100: we should use InlineInputEvent enum here. There are couple of places in vue file as well

wolandec (approved):

wolandec:
Befor merging this PR, we also have to create a PR in EBO to amend changes are required for new ec-inline-input-filed written in TS

src/components/ec-inline-input-field/components/value-text/value-text.vue
```diff
@@ -14,37 +18,41 @@
...
-  value: {
-    default: '',
-    type: String,
-  },
-  isSensitive: {
-    type: Boolean,
-    default: false,
-  },
+interface InlineInputValueTextProps {
+  value?: InlineInputProps['value'],
```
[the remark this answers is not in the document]
derrick-g: Yeah I was quite happy about this, especially as they are so interlinked.

derrick-g:
> Before merging this PR, we also have to create a PR in EBO to amend changes are required for new ec-inline-input-filed written in TS

Can do and minimise the time between PRs in case of another chamleon update in the interim.

PR #1660, opened 2023-08-30
[ONL-7881] Navigation arrows component

wolandec:


src/components/ec-navigation-arrows/ec-navigation-arrows.vue
```diff
@@ -0,0 +1,79 @@
...
+
+<style>
+.ec-navigation-arrows {
+  @apply tw-text-gray-4 tw-h-24;
+
+  &__button {
+    @apply tw-h-24 tw-cursor-pointer;
+    @apply tw-border-0 tw-p-0 tw-bg-transparent tw-fill-current;
+
+    &:hover {
```
mcibique, line 66: We need to apply this for `:focus` too for a11y.

src/components/ec-navigation-arrows/ec-navigation-arrows.vue
```diff
@@ -0,0 +1,79 @@
...
+
+import { computed } from 'vue';
+
+import EcIcon from '../ec-icon/ec-icon.vue';
+import { IconName } from '../ec-icon/types';
+import { NavigationArrowsEvent } from './types';
+
+const emit = defineEmits<{
+  'right-arrow-click': []
+  'left-arrow-click': [],
```
mcibique, line 36: If the component is called navigation arrows then the events could also be called `previous` and `next`.

src/components/ec-navigation-arrows/ec-navigation-arrows.vue
wolandec: The problem with this :focus is that if you focus using keyboard, then hover cursor and move cursor away - focus is still applied. Also if you click using mouse on button, focus is set and even if cursor is not over the button it renders as 
hovered.
image

Not sure this is the behaviour we want to have. We could introduce js code to handle it, but is it worth to support this focus color change as we already have outline on selected button?
image

mcibique: The user must be able to tell which element is focused right now. We must have ability to navigate through clickable components with just the keyboard and it must be clear where the focus currently is. I don't mind that the element is still highlighted after the mouseout. The ugly black outline of the Chrome focus should not be used. Basically, check the ec-btn, this can be rendered as `<button>` and `<a>`, they have focus and hover as well. We also have `Elements` section in the storybook for checking bare element styles, check the anchors and how they behave when focused and hovered.

mcibique (approved):

PR #1662, opened 2023-08-31
[ONL-7885] Corrects missing reactivity on ec-multiple-values filter search

derrick-g:


stavros-tomas (approved):

src/components/ec-multiple-values-selection/ec-multiple-values-selection.spec.js
```diff
@@ -199,6 +199,18 @@ describe('EcMultipleValuesSelection', () => {
...
         ['ABCD efgh'],
       ]);
+      expect(wrapper.findByDataTest('ec-multiple-values-selection__search-input').element.value).toBe('ABCD efgh');
+    });
+
+    it('show set the search query model correctly', async () => {
+      const wrapper = mountEcMultipleValuesSelection({ items, isSearchable: true });
+
+      // In order to test the computed model that is a getter/setter, we directly pass a value to the model to ensure correct reactivity
+      wrapper.vm.searchModel = 'testing word';
```
mcibique, line 209: We shouldn't be testing `wrapper.vm`. This accesses component's internals and it will break in the next refactoring.  Isn't adding the check in the previous test enough to test this behaviour?
```
expect(wrapper.findByDataTest('ec-multiple-values-selection__search-input').element.value).toBe('ABCD efgh');
```
The test above simulates typing to the element. was the element.value empty there before you fixed that?

derrick-g: Thats what I assumed as well, however when testing, its still showed the correct value to be what we set for both the previous issue and with the fix. Only when directly setting the computed model could I see the reactivity event occurring correctly and failing when the fix was not applied.

mcibique: Alright, it's strange. Maybe there's something in VTU that triggers the computed property in every tick. Keep it then.

mcibique (approved):

PR #1682, opened 2023-09-14
[ONL-7879] Left icon for EcInputField

wolandec:


src/components/ec-input-field/ec-input-field.vue
```diff
@@ -44,9 +44,23 @@
...
+        data-test="ec-input-field__left-icon"
+        :name="leftIcon"
+        :size="leftIconSize"
+        :type="leftIconType"
+      />
+    </div>
     <div
       v-if="isLoading || icon"
-      class="ec-input-field__icon-wrapper"
+      class="ec-input-field__icon-wrapper tw-right-0"
```
mcibique, line 63: We don't use TW classes directly in the HTML of reusable components unless it's a deviation from the provided design. We should create separate classes like `ec-input-field__left-icon-wrapper` or it could be modifier `ec-input-field__icon-wrapper--right`. Same for `ec-input-field__icon tw-fill-gray-5` and `ec-input-field__icon-wrapper tw-left-0`.

tests/integration/specs/visual-regression/visual-regression.spec.js
mcibique, line 1: We need to revert this file.

mcibique (approved):

corinaper (approved):

PR #1687, opened 2023-09-19
[ONL-7877] Add support to show slot content instead of a row in ec-table

raulwwq0:
Add a slot to display data in `ec-table` and `ec-smart-table`, mainly to use cards instead of the normal table format.

**Migration of `ec-table` to TS will be done in a future ticket**


src/components/ec-table/ec-table.vue
```diff
@@ -14,6 +14,7 @@
         class="ec-table"
       >
         <ec-table-head
+          v-if="!canShowCustomSlot"
```
wolandec, line 17: not sure we should link displaying of the header and slot in the table body this way.

Let's add a prop `isTableHeaderHidden` it also could have three values and by default header is hidden when slot is shown.

src/components/ec-smart-table/ec-smart-table.vue
```diff
@@ -199,6 +208,9 @@ const isEmpty = computed(() => (props.data ?? []).length === 0);
 
 // slots
 const slots = useSlots();
+const isInCustomSlotThreshold = useMediaQuery('(max-width: 768px)');
```
wolandec, line 211: This functionality is already implemented in `ec-table` here we should just let user of smart table possibility to force using the `row` slot.

src/components/ec-smart-table/ec-smart-table.vue
```diff
@@ -160,6 +165,10 @@ const props = defineProps({
   additionalPayload: Object,
   isFetching: Boolean,
   error: [Error, Object, String],
+  isCustomSlotShown: {
```
wolandec, line 168: let name it isCustomRowShown, and in ec-table as well.

src/components/ec-table/ec-table.vue
```diff
@@ -27,9 +28,18 @@
             :class="{ 'ec-table__row--is-clickable': !!attrs.onRowClick }"
             @click="attrs.onRowClick && attrs.onRowClick({ data: row, rowIndex })"
           >
+            <td
+              v-if="canShowCustomRow"
+              colspan="100%"
```
wolandec, line 33:
```suggestion
              colspan="numberOfColumns"
```

tests/integration/snapshots/visual-regression.spec.js/table--all__chrome.snap.png
wolandec, line 1: we also need to update smart-table story

raulwwq0: smart-table screenshot didn't change because the screenshot width is bigger than 768px, so custom rows are not being shown. Should I create a different smart-table story to force the use of custom rows as well?

wolandec: Yes, let's add it, please

raulwwq0: okay!

wolandec (approved):

wolandec (approved):

PR #1688, opened 2023-09-20
[WIP] [ONL-7886] ec-multiple-values-filter changes

stavros-tomas:
In this PR:
* fixed scrolling on `ec-multiple-values-selection` component
* re-adjusted the icon sizes to 20px after consulting designers
* added extra prop `hasRoundedIcons` to make square icons rounded if needed


tests/integration/snapshots/visual-regression.spec.js/filters-multiple-values-filter--basic__firefox.snap.png
stavros-tomas, line 1: On Firefox scrollbar does not appear unless you hover that's why is not appearing in the screenshot below.

src/components/ec-multiple-values-selection/ec-multiple-values-selection.vue
```diff
@@ -220,6 +225,11 @@ function isItemChecked(item) {
 <style>
 @import '../../styles/tools/scrollbars';
 
+:root,
```
wolandec, line 228: what is the goal of the defining this custom variable in :root element? We don't use it anywhere else except this component.

stavros-tomas: I can totally remove, I mostly tried to followed a pattern. I see a lot this lately without obvious reason but if you also find it weird will hard code instead.

wolandec (approved):

jeger-at (approved):

mcibique (approved):

PR #1690, opened 2023-09-21
[ONL-7918] ec-table-sort to TS

raulwwq0:
In this ticket I include the following changes:
- Migrate all files from `ec-table-sort` component to TS
- Also migrate the `sort-direction` enum to TS as it was required for the component migration


.eslintrc.js
```diff
@@ -121,7 +121,7 @@ module.exports = {
       ],
       rules: {
         'import/first': 'off',
-        'chameleon/vue-props-interface': 'error',
+        // 'chameleon/vue-props-interface': 'error',
```
raulwwq0, line 124: Now that we can use imported types for props this rule seems to not be necessary anymore. However, since we are using a hacky solution to achieve that I was wondering if we should get rid of the rule completely or it would be better to keep it commented for now?

wolandec: I wouldn't remove this in the current PR as we have another check in it(typing defineProps with generic) and we should check if import is working fine everywhere if we remove interfaces from vue files.

src/components/ec-table-sort/ec-table-sort.vue
```diff
@@ -23,32 +23,28 @@
...
-const props = defineProps({
-  direction: {
-    type: String,
-    default: null,
-    validator(value) {
-      return value === null || value === '' || [SortDirection.ASC, SortDirection.DESC].includes(value);
-    },
-  },
-});
+const props = defineProps<TableSortProps>();
```
wolandec, line 38: we had a default value for `direction`, why it's been removed?

raulwwq0: My bad, it should not have been removed, I'm adding it

src/components/ec-table-sort/ec-table-sort.vue
```diff
@@ -61,8 +57,9 @@ const directionTitle = computed(() => {
   return 'Not sorted';
 });
 
-function onSort() {
-  emit('sort', props.direction);
+function onSort(): void {
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+  emit('sort', props.direction!);
```
wolandec, line 62:
```suggestion
  emit('sort', props.direction);
```
If you use withDefaults on line 38

wolandec, line 62:
```suggestion
  emit('sort', props.direction);
```
If you use withDefaults on line 38

src/components/ec-table-sort/ec-table-sort.vue
wolandec, line 1: Hm, it's strange but we have no this component in the main.js, I think we need to export it there as well as it's types

raulwwq0: I think it's because it's an internal use only component, I mean none of the other parts of the table components (`ec-table-head`, `ec-table-footer` or event the ones from `ec-smart-table`) are being exported there. Anyways, I can export it if necessary (and the rest of the components when I migrate them)

wolandec: I think we should expose them as they are ordinary components which could be used elsewhere not only in the smart table component. @mcibique ? Or if they are meant to be used only in `ec-table-head`, then they should be the part of `ec-table-head` component and be placed under `ec-table-head/components` folder. But I think it's fine to just expose those in `main.js`.

.eslintrc.js
raulwwq0: I see, then I'm going to revert that and add a `eslint-disable` comment for that rule in this component

wolandec: I'd prefer to use the same way we use elsewhere, and in the separate ticket move all at once on importing props approach.

raulwwq0: mm... okay! Then should I remove the fix and add it again in that future ticket?

wolandec: yes

src/enums/sort-direction.ts
wolandec, line 1: we could convert it to index.ts. Next time we migrate enum from enums folder we'll place it here alongside SortDirection so all enums will be in one file in the end.

wolandec (approved):

wolandec (approved):

src/components/ec-table-sort/ec-table-sort.vue
mcibique: I don't have a strong opinion, `ec-table` and `ec-table-*` components were first components we created and back then we didn't have a rule that internal components should be in parent `ec-table/components`. Every `ec-table-*` component should be used in `ec-table` not in `ec-table-head`.

I'm not sure these components are going to be that useful alone so I'd exported them only when needed in the future. For example, metroline-item is nested in metroline/components and is also exported because you can't create metroline structure without the metroline-item. That is useful, I don't think we are going to have a place where we would need to compose the table from scratch using these ec-table-* blocks.

wolandec (approved):

PR #1691, opened 2023-09-22
ONL-7883-Introduce flex width for main container button slot

corinaper:
image

ERI: https://drive.google.com/file/d/1uw7jE5K3VWHUSlP_CcfvyKIxPTRrRAy_/view?usp=sharing
Guidelines: https://www.figma.com/file/R8pt5Ic85URRNSyTagcKBd/%F0%9F%91%A5-Beneficiaries-redesign?node-id=3097%3A201712&mode=dev && https://www.figma.com/file/qLv2in1OHCrSOEoNivIQA1/%F0%9F%93%98-Guidelines?node-id=1707%3A30165&mode=dev


src/components/ec-main-container/ec-main-container.vue
```diff
@@ -54,6 +60,11 @@ interface MainContainerProps {
 defineProps<MainContainerProps>();
 
 const slots = useSlots();
+const { width } = useWindowSize();
```
wolandec, line 63: https://vueuse.org/core/useMediaQuery/ we use this hook for such cases

src/components/ec-main-container/ec-main-container.vue
```diff
@@ -55,6 +61,8 @@ defineProps<MainContainerProps>();
 
 const slots = useSlots();
 
+const isBelowBreakPointSize = useMediaQuery('(max-width: 344px)');
```
wolandec, line 64: as far as I understand the mobile threshold is 320, why 344 is used here?

corinaper: This is when the button should be wrapped based on new guidelines: https://www.figma.com/file/R8pt5Ic85URRNSyTagcKBd/%F0%9F%91%A5-Beneficiaries-redesign?node-id=3097%3A201712&mode=dev

wolandec: it's better to title the breakpoint with more meaningfull name, e.g. `isMobileView`

wolandec: That is not a style guide I guess, for example for 460px we have different representation. As far as I remember we selected 768 and 320 as thresholds, the last one is for mobile view. Let's at least clarify it?

wolandec:
We need to update story screenshots

src/components/ec-main-container/ec-main-container.vue
```diff
@@ -3,42 +3,49 @@
     class="ec-main-container"
     data-test="ec-main-container"
   >
+
     <div
-      v-if="title || hasSlot('cta')"
-      class="tw-grid"
+      v-if="hasSlot('breadcrumbs')"
```
wolandec, line 8: if there is no `breadcrumbs` slot then we won't show `cta` slot, I guess it's not wanted behaviour.

wolandec: I see we have another v-if below

wolandec (approved):

PR #1696, opened 2023-09-26
[ONL-7919] ec-table-head to TS

raulwwq0:


src/components/ec-table-head/ec-table-head.vue
```diff
@@ -49,47 +49,52 @@
...
+const props = withDefaults(defineProps<TableHeadProps>(), {
+  columns: () => [],
+  sorts: () => [],
 });
 
-function getSortDirection(column) {
+function getSortDirection(column: TableHeadColumn): SortDirection | undefined {
   const existingSort = props.sorts.find(sort => sort.column === column.name);
   return existingSort && existingSort.direction;
 }
```
wolandec, line 84:
```suggestion
function getSortDirection(column: TableHeadColumn) {
  return props.sorts.find(sort => sort.column === column.name)?.direction;
}
```

src/components/ec-table-head/ec-table-head.vue
```diff
@@ -49,47 +49,52 @@
...
 
-function onSort(column) {
+function onSort(column: TableHeadColumn): void {
   emit('sort', column);
 }
```
wolandec, line 88:
```suggestion
function onSort(column: TableHeadColumn) {
  emit('sort', column);
}
```
TS infers such kind of types from the function code

src/components/ec-table-head/ec-table-head.vue
```diff
@@ -49,47 +49,52 @@
...
 
-function getWidthStyle(column) {
+function getWidthStyle(column: TableHeadColumn): StyleValue {
   if (column && (column.maxWidth || column.minWidth)) {
     return { maxWidth: column.maxWidth, minWidth: column.minWidth };
   }
-  return null;
+  return null as unknown as StyleValue;
```
wolandec, line 94:
```suggestion
  return null;
```
null doesn't match type StyleValue

src/components/ec-table-head/ec-table-head.vue
```diff
@@ -49,47 +49,52 @@
...
   return existingSort && existingSort.direction;
 }
 
-function onSort(column) {
+function onSort(column: TableHeadColumn): void {
   emit('sort', column);
 }
 
-function getWidthStyle(column) {
+function getWidthStyle(column: TableHeadColumn): StyleValue {
```
wolandec, line 90:
```suggestion
function getWidthStyle(column: TableHeadColumn) {
```

src/components/ec-table-head/ec-table-head.vue
```diff
@@ -49,47 +49,52 @@
...
   if (column && (column.maxWidth || column.minWidth)) {
     return { maxWidth: column.maxWidth, minWidth: column.minWidth };
   }
-  return null;
+  return null as unknown as StyleValue;
 }
 
-function getStickyColumnClass(colIndex, columns) {
+function getStickyColumnClass(colIndex: number, columns: TableHeadColumn[]): string | null {
```
wolandec, line 97:
```suggestion
function getStickyColumnClass(colIndex: number) {
```

raulwwq0: We can't remove `columns` because is needed to get `columns.length`

wolandec: We can use props.columns

src/components/ec-table-head/ec-table-head.vue
wolandec: Where was the problem in template or in this function?

src/components/ec-table-head/ec-table-head.vue
raulwwq0: oh... makes sense... I'm changing it!

src/components/ec-table-head/ec-table-head.vue
raulwwq0: In the template

wolandec: I think we could return undefined then, instead of null.

wolandec (approved):

wolandec (approved):

PR #1705, opened 2023-10-03
[ONL-7970] Flag to hide specific table filters and clear filters

raulwwq0:


src/components/ec-table-filter/ec-table-filter.vue
```diff
@@ -58,11 +64,28 @@ const props = defineProps({
...
     default: 'Clear filters',
   },
+  areFiltersHidden: {
+    type: Boolean,
+    default: () => undefined,
+  },
+  hiddenFiltersNames: {
+    type: Array,
+    default: () => [],
+  },
```
wolandec, line 74: `hiddenFilterNames` enough to define what filter to hide(if filter is in the array - it is hidden).
areFiltersHidden is not necessary, but it could be replaced by prop `isClearButtonHidden`

So you have full control what to show and what to hide.

src/components/ec-table-filter/ec-table-filter.vue
```diff
@@ -58,11 +64,28 @@ const props = defineProps({
...
 });
 
 const emit = defineEmits(['update:modelValue', 'change']);
+const areFiltersHiddenThreshold = useMediaQuery('(max-width: 768px)');
 
+const alwaysShownFilters = props.filters ? props.filters.filter(filter => (!props.hiddenFiltersNames.includes(filter.name))) : [];
 const hasFilters = computed(() => !!Object.keys(props.modelValue).length);
+const canHideFilters = computed(() => (props.areFiltersHidden || (props.areFiltersHidden === undefined && areFiltersHiddenThreshold.value)));
```
wolandec, line 82: I don't think we should have this logic here, let's base hidden state for every filter instead of introducing the hiddenFilterNames prop.

So we have filters prop and can pass idHidden state for every filter.

src/components/ec-table-filter/ec-table-filter.vue
derrick-g: Can we place the 'isHidden' in the actual `filters` object array, then we control the filter and can toggle its visibility from the parent component. The filters object looks like: 
```
    label: ...,
    name: ....,
    component: ....,
    items: [....],
    isHidden: boolean
```

Ditto on the button prop @wolandec suggested,

src/components/ec-table-filter/ec-table-filter.vue
```diff
@@ -58,11 +64,28 @@ const props = defineProps({
...
+    default: () => undefined,
+  },
+  hiddenFiltersNames: {
+    type: Array,
+    default: () => [],
+  },
 });
 
 const emit = defineEmits(['update:modelValue', 'change']);
+const areFiltersHiddenThreshold = useMediaQuery('(max-width: 768px)');
```
derrick-g, line 78: I think the toggling of visibility should be handled by the parent component when used.

derrick-g, line 78: I think the toggling of visibility should be handled by the parent component when used.

raulwwq0: By doing that I think that in this case we can also reuse the [`useMediaQuery` from `ec-table`](https://github.com/Ebury/chameleon/blob/master/src/components/ec-table/ec-table.vue#L83) as well since it's the same threshold, wdyt?

src/components/ec-table-filter/ec-table-filter.vue
raulwwq0: I'm agree with both changes, I'm going to change them

src/components/ec-table-filter/ec-table-filter.vue
wolandec: as soon as we introduce isHidden prop to every filter, then we let user of this component let to decide when and what filters to hide.

src/components/ec-table-filter/ec-table-filter.vue
```diff
@@ -58,11 +63,22 @@ const props = defineProps({
...
+    default: false,
+  },
 });
 
 const emit = defineEmits(['update:modelValue', 'change']);
 
+const shownFilters = computed(() => (props.filters ? props.filters.filter(filter => !filter.isHidden) : []));
 const hasFilters = computed(() => !!Object.keys(props.modelValue).length);
+// eslint-disable-next-line no-underscore-dangle
+const haveOnlyTextFilter = computed(() => (shownFilters.value[0].component.__name === 'ec-text-filter') && props.isClearButtonHidden);
```
wolandec, line 77: let's move this to filter prop as well, so we could have any filter full width

wolandec: prop for filter could have name isFullWidth

raulwwq0: okay!

src/components/ec-smart-table/ec-smart-table.vue
```diff
@@ -169,6 +170,10 @@ const props = defineProps({
     type: Boolean,
     default: () => undefined,
   },
+  isClearButtonHidden: {
```
wolandec, line 173:
```suggestion
  isClearFiltersButtonHidden: {
```

src/components/ec-table-filter/ec-table-filter.vue
```diff
@@ -58,12 +63,21 @@ const props = defineProps({
...
+    default: false,
+  },
 });
 
 const emit = defineEmits(['update:modelValue', 'change']);
 
+const shownFilters = computed(() => (props.filters ? props.filters.filter(filter => !filter.isHidden) : []));
 const hasFilters = computed(() => !!Object.keys(props.modelValue).length);
 
+watch(() => shownFilters.value, () => {
```
wolandec, line 77: why we clear filters if change hide or show some?

raulwwq0: We agreed to clear them when a filter is hidden to avoid having a filter set that cannot be clear by the user. However, now that you mention it maybe it will be a better idea to not clear all of them but clearing only the ones that are going to be hidden...

wolandec: I'd delegate this logic to component user to have flexibility in place where we use filters.

raulwwq0: mm... tbh I'm not to sure about that, I mean how are we going to clean the filters from other components?

wolandec: We use `ec-smart-table` and pass filters(including `isHidden` flag for each of them). When we decide to hide a filter at the same time we could clear the filter value changing `filterModel`.

raulwwq0: oh okay! Then I'm going to remove this

derrick-g (approved):

src/components/ec-table-filter/ec-table-filter.spec.js
```diff
@@ -87,6 +87,39 @@ describe('EcTableFilter', () => {
     expect(wrapper.element).toMatchSnapshot();
   });
 
+  it('should hide filters with "tw-hidden" class when "isHidden" is set', async () => {
```
wolandec, line 90: can we have test for isClearButtonHidden?

src/components/ec-smart-table/ec-smart-table.vue
```diff
@@ -169,6 +170,10 @@ const props = defineProps({
     type: Boolean,
     default: () => undefined,
   },
+  isClearFiltersButtonHidden: {
```
wolandec, line 173: we need to add a test for this prop

src/components/ec-table-filter/ec-table-filter.spec.js
```diff
@@ -87,6 +87,57 @@ describe('EcTableFilter', () => {
...
+      {
+        data() {
+          return {
+            value: { feeType: modelValue.feeType },
+            filters,
+          };
+        },
+      },
+    );
+    expect(wrapper.element).toMatchSnapshot();
```
wolandec, line 138: let's narrow the snapshot or literally check whether clear filters button exists?

raulwwq0: I'm going to check if the button exists, should I do the same for the `ec-smart-table` test?

wolandec: Yeah, that snapshot is not too big, but we could do the same there

wolandec (approved):

wolandec (approved):

PR #1714, opened 2023-10-13
[ONL-8007] Fix ec-smart-table filter model reactivity

raulwwq0:
It was not possible to update filters using the model from a parent component because it was not reacting to `filter` prop changes

### DIFF
--- a/src/components/ec-smart-table/ec-smart-table.spec.js
+++ b/src/components/ec-smart-table/ec-smart-table.spec.js
@@ -582,6 +582,20 @@ describe('EcSmartTable', () => {
       }]);
     });
 
+    it('should change filters when filter prop is changed', async () => {
+      const wrapper = mountEcSmartTable({
+        columns,
+        filters,
+        filter: prefilter,
+      });
+      const tableFiltersElement = wrapper.findByDataTest('ec-smart-table__filter').element;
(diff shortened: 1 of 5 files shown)


src/components/ec-smart-table/ec-smart-table.spec.js
```diff
@@ -547,6 +547,29 @@ describe('EcSmartTable', () => {
       }]);
     });
 
+    it('should trigger fetch when filter prop is changed', async () => {
```
wolandec, line 550: Normally it shouldn't as in this case we can't say whether this event was triggered by changes from props or by changes which were done by user. But

raulwwq0: I think in both cases it's necessary to fetch benes because filters were changed and also this test is just to check if the model is being updated (checking the watch we added) to avoid having the same issue in the future

wolandec: To check whether model is updated we should check filters snapshot before the change and after. This event triggering test could stay as well.

raulwwq0: okay! I'm going to add a snapshot test as well

PR #1719, opened 2023-10-17
[ONL-7972] Make table pagination responsive

raulwwq0:


src/components/ec-table-pagination/ec-table-pagination.vue
```diff
@@ -128,6 +135,14 @@ const props = defineProps({
     type: String,
     default: 'Items per page',
   },
+  isPageSizeHidden: {
```
wolandec, line 138: In ec-table.vue we [have](https://github.com/Ebury/chameleon/blob/e5633ff288ce61459b5a075d5cb35711efc32591/src/components/ec-table/ec-table.vue#L84) this default behaviour, I think we should use the same approach here for the compact pagination view.

wolandec: That 768 could be included in the [config](https://github.com/Ebury/chameleon/blob/fc43b627328347d9d66a1988fc9818e303ea7f64/src/config.js#L1-L0) as 
```
screenSizes: {
  md: 768
}
```
@mcibique if you don't mind

raulwwq0: I thought about doing is as the [isClearFiltersButtonHidden prop](https://github.com/Ebury/chameleon/blob/e5633ff288ce61459b5a075d5cb35711efc32591/src/components/ec-table-filter/ec-table-filter.vue#L66), but it makes more sense to follow that default behavior 

And regarding the config change I think we can do the same for ebo as we have 3 different sizes for `useMediaQuery` there: 1280, 768 and 576 px

wolandec: Yes, I think it makes sense to have all thresholds in screenSizes config property.

src/components/ec-table-pagination/ec-table-pagination.vue
```diff
@@ -149,6 +149,7 @@ const isDesktop = useMediaQuery('(min-width: 768px)');
 
 const canShowPageSize = computed(() => (props.isPageSizeHidden === false || (props.isPageSizeHidden === undefined && isDesktop.value)));
 const canShowCustomInfo = computed(() => (props.isCustomInfoHidden === false || (props.isCustomInfoHidden === undefined && isDesktop.value)));
+const canShowTextEllipsis = computed(() => (!isDesktop.value && (!canShowCustomInfo.value || !canShowPageSize.value)));
```
wolandec, line 152: what if this text will be always ellipsis?

raulwwq0: Tables where we don't have the responsive changes applied yet will look like this:
image

So I think it would be better to not use ellipsis in these cases

src/components/ec-table-pagination/ec-table-pagination.spec.js
```diff
@@ -26,7 +36,13 @@ describe('EcTablePagination', () => {
   });
 
   it('should display page information by calculating total number of pages', () => {
-    const wrapper = mountEcTablePagination({ total: 20, numberOfItems: 5, page: 2 });
+    const wrapper = mountEcTablePagination({
+      total: 20,
+      numberOfItems: 5,
+      page: 2,
+      isPageSizeHidden: false,
+      isCustomInfoHidden: false,
```
wolandec, line 44: can we have tests for true values?

src/components/ec-table-pagination/ec-table-pagination.vue
wolandec: In this case we won't have flexibility, for example for some cases. We want to show pagination with no hidden parts for md threshold, but want to show customSlot for table rows(some long strings in the table data, so we want to show custom slot earlier than md threshold).

Sticking with tailwind and css we are losing this flexibility.

mcibique: In general, hiding/displaying something based on the media query should be always done via CSS. the page size and total should be always rendered and hidden when reaching a breakpoint.

src/components/ec-smart-table/ec-smart-table.vue
```diff
@@ -173,6 +175,14 @@ const props = defineProps({
     type: Boolean,
     default: false,
   },
+  isPageSizeHidden: {
+    type: Boolean,
+    default: () => undefined,
+  },
+  isPaginationCustomInfoHidden: {
```
mcibique, line 182: `isTotalHidden`

src/components/ec-table-pagination/ec-table-pagination.vue
wolandec: We'll face the case eventually, when we'll have to introduce this props to maintain view from the parent component.

src/components/ec-smart-table/ec-smart-table.vue
```diff
@@ -173,6 +175,14 @@ const props = defineProps({
     type: Boolean,
     default: false,
   },
+  isPageSizeHidden: {
+    type: Boolean,
+    default: () => undefined,
```
mcibique, line 180: Why we need 3 states for these booleans everywhere? Use true/false to render the page size part and then media query in CSS to hide it/show it when on small screens. If it was rendered, it will hide/show, if it's not rendered, then it will be ignored.

src/components/ec-table-pagination/ec-table-pagination.vue
mcibique: You can't configure tailwind from EBO anyway, so passing breakpoints via config will result in some components going after value from TW and some of them using different from a config.

If you want to ignore responsivnes declared in CSS, create one single prop called is-responsive, add a modifier to a component and style the component based on that modifier. See ec-alert using approach like this. ec-alert has also a mixin that allows to redefine the breakpoint if needed.

mcibique: Using a custom class: https://github.com/Ebury/chameleon/blob/master/src/components/ec-alert/ec-alert.story.ts#L66 with https://github.com/Ebury/chameleon/blob/master/src/components/ec-alert/ec-alert.story.css.

wolandec: For smart-table for example this approach could be too restricted or we'll have to define new thresholds for all mixins for subcomponents which leads to know us as users of the smart-table component about it's structure and what it is using under the hood, which what we want to avoid.

src/components/ec-smart-table/ec-smart-table.vue
wolandec: Three states was to have possibility force component to show/hide(whatever component is doing regarding the media query) and undefined for the default behaviour.

wolandec (approved):

src/components/ec-table-pagination/ec-table-pagination.spec.js
```diff
@@ -12,6 +12,16 @@ describe('EcTablePagination', () => {
     });
   }
 
+  beforeEach(() => {
+    window.matchMedia = jest.fn().mockImplementation(query => ({
```
mcibique, line 16: Is this still needed? If yes, we need to restore it back to the origin in `afterEach` so it doesn't spoil other tests running after this one.

src/components/ec-smart-table/ec-smart-table.spec.js
```diff
@@ -54,6 +54,16 @@ describe('EcSmartTable', () => {
     }, mountOpts);
   }
 
+  beforeEach(() => {
+    window.matchMedia = jest.fn().mockImplementation(query => ({
```
mcibique, line 58: Same here, is this still needed?

src/components/ec-table-pagination/ec-table-pagination.spec.js
raulwwq0: It is not needed at all, thanks for noticing!

mcibique (approved):

mcibique (approved):

PR #1728, opened 2023-10-27
[ONL-8028] Migrates the checkbox to TS

derrick-g:


mcibique (approved):

src/components/ec-checkbox/ec-checkbox.vue
```diff
@@ -80,49 +80,46 @@
...
   watch,
 } from 'vue';
 
 import { getUid } from '../../utils/uid';
 import EcIcon from '../ec-icon';
+import { IconName } from '../ec-icon/types';
+import { CheckboxEvent, type CheckboxEvents } from './types';
+
+interface CheckboxProps {
+  modelValue?: boolean,
```
corinaper, line 104: shouldn't modelValue be required?

derrick-g: For DX we want the user of the component to have optional control of the component, similar to the ec-input-field. For example we may want to show a disabled checked checkbox with no ability for the user to interact with, so forcing a modelValue would be unnecessary binding.

corinaper (approved):

stavros-tomas (approved):

stavros-tomas (approved):

PR #1737, opened 2023-11-02
[ONL-8065] Replace jest with vitest

mcibique:
- Replaced jest with vitest
- Replaced all jest mocks with vitest mocks
- Updated eslint rules to match the plugin for jest
- Removed `.toMatchDiffSnapshot` because there's no plugin like that for vitest
- Updated all snapshots because vitest serialiser uses different naming
- Removed CVueWrapper. We're now extending VueWrapper directly. This allows us to use all generics that VueWrapper provides, e.g. `.findByDataTest<HTMLAnchorElement>('my-anchor')` instead of `findByDataTest('my-anchor').element as HTMLAnchorElement`.
- Vitest coverage has a better detection than previous one so it detected tons of places in SFC templates which were not covered by tests. This lead to adding tests for these and sometimes fixing issues (e.g. data-test attrs in some components were not merged but replaced). I also rewritten tests for ec-tooltip to be able to cover all cases.
- Replaced `(aMethod as jest.Mock)` with`vi.mocked(aMethod)` for better type safety that is provided by vitest.
- Removed as many `any` as I could in the test
- Fixed typos, missing describe blocks


src/components/ec-btn-dropdown/ec-btn-dropdown.spec.js
```diff
@@ -115,7 +136,7 @@ describe('EcBtnDropdown', () => {
         },
         {
           attrs: {},
-          href: '/payments/?add-new-payment',
+          href: '#/payments/',
```
wolandec, line 139: Just curiosity, why `href` is changed?

tests/utils/global.d.ts
```diff
@@ -1,8 +1,14 @@
-import { VueWrapper } from '@vue/test-utils';
+/* eslint-disable @typescript-eslint/no-unused-vars */
```
wolandec, line 1:
```suggestion
```
T on line 11 could be removed as well?

wolandec (approved):

derrick-g (approved):

src/components/ec-btn-dropdown/ec-btn-dropdown.spec.js
mcibique: I was getting errors from JSDOM that navigation is not implemented only hash navigation is supported. We have tests where we click on anchors with this href - to remove the errors, I changed the hrefs to hashes.

tests/utils/global.d.ts
mcibique: Yeap, removed.

wolandec (approved):

stavros-tomas (approved):

PR #1745, opened 2023-11-16
[ONL-8058] Infinite scrolling in smart table

raulwwq0:


src/components/ec-smart-table/ec-smart-table.vue
```diff
@@ -260,4 +280,38 @@ function getEcTableSlots() {
   delete tableSlots.pages;
   return tableSlots;
 }
+
+// infinite scroll
+
+const intersectionTarget = ref(null);
+
+const canLoadMore = computed(() => props.data && props.data.length < props.totalRecords);
```
wolandec, line 288: are we going to pass all data to make this computed false?

For example we have 100 total records, and 10 per page, to stop intersection we need to pass all 100 records to `props.data`, and in this case we'll display all 100 records in the table.

What if user have 1000, 5000 totalRecords, we'll paste them all in our table and DOM?

src/components/ec-smart-table/ec-smart-table.vue
```diff
@@ -119,20 +119,36 @@
             v-bind="slotData"
           /></template>
         </ec-table>
+        <div
+          ref="intersectionTarget"
```
wolandec, line 123: I would name it more semantic, `tableEndDetector`

src/components/ec-smart-table/ec-smart-table.vue
raulwwq0: Sadly yes, I know that in terms of performance is going to be awful when we have thousand of records and the user wants to scroll all the way down. 

Even if we want to remove from DOM the ones that were already scrolled (let's say that in the screen you can only fit 10 records, in theory we can remove the ones outside the viewport from DOM) we still need to store them in memory so I don't think this would be helpful at all.

It's possible to implement a kind of 2-way infinite scrolling (when scrolling down but also when scrolling up) to avoid that, but:
- I think is over-engineering for our purpose, at least for now
- In terms of UX is even worse than the normal infinite scrolling

wolandec: Even though I can accept data storing for all fetched records, which is also debatable. I think we should prevent to render lots of records setting maximum of rendered rows. On some env we have thousands of records for this table, as I understand it's normal situation.

src/components/ec-smart-table/ec-smart-table.spec.js
```diff
@@ -677,4 +677,61 @@ describe('EcSmartTable', () => {
...
+      const wrapper = mountEcSmartTableWithData(data, {
+        isInfiniteScrollEnabled: true,
+      });
+      const tableEndDetector = wrapper.findByDataTest('ec-smart-table__table-end-detector');
+      expect(tableEndDetector.element).toMatchSnapshot();
+    });
+
+    it('should not render EcLoading loading SVG if "isInfiniteScrollEnabled" prop is true', () => {
+      const wrapper = mountEcSmartTableWithData(data, {
+        isInfiniteScrollEnabled: true,
```
wolandec, line 692: loading is rendered when `isFetching` prop is true. So we should set it to true and check whether loading is rendered.

wolandec: Have you tried to fail this test by setting `isInfiniteScrollEnabled: false`?

raulwwq0: eh... no, I didn't, thanks for noticing!

src/components/ec-smart-table/ec-smart-table.vue
```diff
@@ -260,4 +280,38 @@ function getEcTableSlots() {
...
+    ([{ isIntersecting }]) => {
+      if (isIntersecting && !props.isFetching) {
+        page.value += 1;
+      }
+    },
+    /* c8 ignore stop */
+  );
+
+  watch(() => canLoadMore.value, () => {
+    if (!canLoadMore.value) {
```
wolandec, line 303: what if it is changed from false to true? Should we resume IntersectionObserver again?

raulwwq0: Not in the current implementation because when `canLoadMore.value` is false (basically when we already fetch all data) we just completely stop the IntersectionObserver as it is no longer needed

If in future tasks we don't want that behavior we can pause and resume the observer by using [`pause` and `resume`](https://github.com/vueuse/vueuse/blob/main/packages/core/useIntersectionObserver/index.ts#L112) methods provided by the `useIntersectionObserver` composable instead

jeger-at (approved):

wolandec (approved):

wolandec (approved):

PR #1765, opened 2023-12-05
[ONL-8097] Fix search filter width

raulwwq0:
The original idea was to move the search bar together with the "Clear filters" button below the rest of the filters when both of the previous mentioned element width together reach a minimum width value. However, after some testing I found this was not a proper solution due to it was really complex and not very reliable (you can see how it was going in the [wip commit](https://github.com/Ebury/chameleon/commit/8ca87839033fb49ee4d369532b78d19e5b8c8e7e), but keep in mind that it's not the whole implementation).

So, instead of following that and after talking about the issue with Alba we decide moving the search bar filter and the button below the filters when the screen width is below 850px. We decide this value because it allows displaying the "Clear filters" button next to the search bar when it's bigger that 850px without breaking in any of the current translations (even German...) and there is still some space in case we have longer translations in the future.

Also, I removed some Tailwind classes that were used directly in the template and replaced them with custom CSS that wrap those classes.


src/components/ec-table-filter/ec-table-filter.vue
```diff
@@ -95,8 +101,9 @@ function clearFilters() {
 
 <style>
 .ec-table-filter {
-  @apply tw-block;
+  @apply tw-flex;
```
wolandec, line 104: I'd always use grid

src/components/ec-table-filter/ec-table-filter.story.js
```diff
@@ -211,6 +211,33 @@ const AllTemplate = storyArgs => ({
...
+      component: markRaw(EcDateRangeFilter),
+    }, {
+      label: 'Price',
+      name: 'price',
+      component: markRaw(EcCurrencyFilter),
+      comparisonSymbolItems,
+      currencyItems,
+    }, {
+      name: 'text',
+      class: 'lg:tw-col-auto tw-col-start-1 tw-col-end-[-2]',
```
wolandec, line 238: do you think we still need this class for filter? You can describe the grid in the layout class.

raulwwq0: in this case it's needed because we want that filter (search bar) to span through almost all columns when width is below `lg` threshold and, if I'm not wrong (because maybe I miss something), we need to apply those classes to the filter itself

wolandec: Filter could have full width of what container provides. Container for each filter is a grid cell, which we define using grid layout templates.

wolandec: No, I'd doubt we can use area names with Tailwind. I mean something like this.

image

raulwwq0: The issue and the reason behind using those classes is that when using more than one row we can't set a different cell width, otherwise it will happen something like this:

image

So that's why I use those classes to span the filter across the other columns:

image

I also thought about using subgrids but it was not possible in our case due to how we render the filters (and also because I don't knnow if we can use it with tailwind...)

wolandec: I see, okay, let's keep filter classes then.

src/components/ec-table-filter/ec-table-filter.vue
```diff
@@ -67,12 +72,20 @@ const props = defineProps({
...
+  },
 });
 
 const emit = defineEmits(['update:modelValue', 'change']);
 
 const hasFilters = computed(() => !!Object.keys(props.modelValue).length);
 
+const hasFullWidthFilter = computed(() => (props.filters ? Object.values(props.filters).some(filter => filter.isFullWidth) : false));
+
+console.log(hasFullWidthFilter.value);
```
wolandec, line 87:
```suggestion
```

src/components/ec-table-filter/ec-table-filter.story.js
```diff
@@ -280,6 +308,20 @@ const AllTemplate = storyArgs => ({
         />
       </div>
     </div>
+    <h2 class="tw-m-24">Use custom CSS Grid layout</h2>
+    <div class="tw-flex tw-px-20">
+      <div class="tw-my-auto tw-mx-20 tw-w-full ec-card">
+        <ec-table-filter
+          v-model="model"
+          :filters="searchFilterFillingRemainingSpace"
+          layout-class="!tw-grid tw-grid-rows-2 !tw-grid-flow-row tw-grid-cols-[repeat(4,_auto)_1fr_auto] lg:tw-grid-rows-1"
```
mcibique, line 317: This violates basic principles that parent should not dictate how a child should be displayed. I'm still lost why is this needed?

raulwwq0: The idea is to let the user of the `ec-table-filter` component to style the way filters are being displayed. In this case is used to move the search filter bar below the other filters when needed, but if we want to use the same component but with a different layout the idea was using this prop instead of modifying the component at all

mcibique, line 317: BTW, we should never use `important`. Why is that needed?

raulwwq0: This is only needed here because TW classes have lower specificity than the CSS classes we used inside the component, so styles are not being applied (unless we wrapped those in a CSS class, but I don't know if we can do that in Storybook stories tbh)

mcibique: We should pass a boolean `stretch` to the filter and ec-table-filter should determine the layout based on that. If it detects that all filters in the `filters` prop have stretch set to false, use one layout, otherwise use different layout.

wolandec: In this case I suggested this approach to let parent component to set grid layout. Yes it's not the ideal solution, but to have flexibility depends on a particular use case we can use it.

We share a restricted part of knowledge to the parent component. Just grid layout information.

I wanted to avoid situation when tomorrow we'll be asked to show all filters one under another on mobile and `clear` button on the same line with the last one. We'll whether decline this design or go and implement again some stuff in the table-filter component implementing this particular case.

mcibique: We should use BEM and only have classes with specificity of 0100 so they never clash and can be easily overridden. That's the whole purpose of using BEM.

mcibique: I really don't want go this way. I have never seen a library of reusable components where you can pass classes to components to override their behaviour or layout or styles. It's responsibility to ec-table-filter to lay down components inside, having a class to override that means we have to provide these classes everywhere we use ec-table-filter which means they are going to break every time we refactor something inside ec-table-filter. It's going to be maintenance nightmare. Follow the YAGNI and don't try to optimise/support for cases we don't have.

tests/integration/snapshots/visual-regression.spec.js/smart-table--all__firefox.snap.png
wolandec, line 1: There is a problem with alignment
image

raulwwq0: Yeap, I added a `margin-bottom` to the filters, but that download button in the table header itself not on filters. I'm fixing it

src/components/ec-table-filter/ec-table-filter.vue
```diff
@@ -67,12 +72,16 @@ const props = defineProps({
...
   },
+  hasStretchFilter: Boolean,
 });
 
 const emit = defineEmits(['update:modelValue', 'change']);
 
 const hasFilters = computed(() => !!Object.keys(props.modelValue).length);
 
+const hasFullWidthFilter = computed(() => (props.filters ? Object.values(props.filters).some(filter => filter.isFullWidth) : false));
+const stretchLayoutColumns = computed(() => (props.filters ? `repeat(${Object.values(props.filters).length - 1}, auto) 1fr auto` : ''));
```
wolandec, line 83: the problem I can see here is purge css process, tailwind used it before, now it doesn't, however it has a step which cleans up the bundle and remove all unused classes, we need to check whether it could affect this case.

https://tailwindcss.com/docs/content-configuration#dynamic-class-names

raulwwq0: How `v-bind` in CSS works is by creating an empty CSS var in the component style and update its value when needed. 
For example, lets say that `stretchLayoutColumns` creates a CSS var called `--XXXXX`, so:
1. add a style attribute with the empty var -> `<div class="..." style="--XXXXX: ;">...</div>`
2. when we use `v-bind` with `stretchLayoutColumns` within the component it will replace it for the CSS var -> 
```
&--has-stretch-filter {
   grid-template-columns: v-bind(stretchLayoutColumns);
}
/* ^^ This is transformed into:*/
&--has-stretch-filter {
   grid-template-columns: var(--XXXXX);
}
```
3. When `stretchLayoutColumns` value changes Vue updates the CSS var value from the style attribute

Even if the CSS var value is empty is being used, so purge CSS can't remove it. I just build Chameleon on my local to test it and here is how it looks in browser devtools:
image

wolandec: I see, yes it works, as it is not tailwind class. Sure. Sorry, for some reason I was in the context of tailwind classes considering this :).

wolandec (approved):

src/components/ec-table-filter/ec-table-filter.vue
```diff
@@ -95,17 +104,62 @@ function clearFilters() {
...
+    @apply tw-auto-cols-auto;
+  }
+
+  &--has-stretch-filter {
+    @apply tw-grid-flow-row;
+
+    /* Since the number of filter can vary we need to set the columns layout according to the filters we have
+    We can't use Tailwind for this because the generated class (i.e. tw-grid-cols-[repeat(4,_auto)_1fr_auto] if we have 4 filters)
+    won't be generated because TW can't react to those changes. However we can use Vue v-bind function to achieve that.
+    See https://vuejs.org/api/sfc-css-features.html#v-bind-in-css */
```
wolandec, line 127: I think this could be removed as well

wolandec (approved):

PR #1818, opened 2024-01-16
[ONL-6889] Replace mitt with useEcCountdown.

mcibique:
Removed last code using `mitt` and created composable with createEventHook from vueuse instead.

Fixed ec-timer not responding when the `seconds` prop has changed.


src/components/ec-timer/ec-timer.vue
```diff
@@ -94,46 +92,32 @@ const strokeWidth = 4;
 const diameter = radius * 2 + strokeWidth;
 const viewbox = `0 0 ${diameter} ${diameter}`;
 const circumference = 2 * Math.PI * radius;
-const steps = circumference / props.seconds;
+const steps = computed(() => circumference / props.seconds);
```
wolandec, line 95: we need test for the case when we change props.seconds

src/composables/use-ec-countdown/use-ec-countdown.ts
```diff
@@ -0,0 +1,71 @@
...
+} from 'vue';
+
+export default function useEcCountdown(secondsToGo: MaybeRef<number>) {
+  const secondsLeft = ref<number>(unref(secondsToGo));
+
+  watch(() => unref(secondsToGo), () => {
+    restart(unref(secondsToGo));
+  });
+
+  const isRunning = ref<boolean>(false);
```
wolandec, line 13:
```suggestion
  const isRunning = ref(false);
```

src/composables/use-ec-countdown/use-ec-countdown.ts
```diff
@@ -0,0 +1,71 @@
+import { createEventHook } from '@vueuse/core';
+import {
+  type MaybeRef, ref, unref, watch,
+} from 'vue';
+
+export default function useEcCountdown(secondsToGo: MaybeRef<number>) {
+  const secondsLeft = ref<number>(unref(secondsToGo));
```
wolandec, line 7:
```suggestion
  const secondsLeft = ref(unref(secondsToGo));
```
type of secondsLeft will be infer from unref(secondsToGo) return type

wolandec (approved):

raulwwq0 (approved):

raulwwq0 (approved):

PR #1820, opened 2024-01-18
[ONL-7281] feat:replace classic date input with datepicker

stavros-tomas:
Ec datepicker for date-range component


src/components/ec-date-range-filter/ec-date-range-filter.vue
```diff
@@ -147,11 +142,30 @@ const numberOfSelectedFilters = computed(() => {
...
+  });
 }
+
 function onBlur() {
   emit('blur', { from: fromValueDate.value, to: toValueDate.value });
 }
+
+const isAutoHideEnabled = ref(true);
+const allPopoverOptions = computed(() => ({
+  autoHide: isAutoHideEnabled.value, // autoHide of the ec-filter-popover should be disabled while flatpickr is open, otherwise selecting value in the flatpickr will close this popover too.
```
wolandec, line 157: if we want to manage this option regardless `props.popoverOptions.autoHide` value, we need to put this after ..props.popoverOptions

tests/integration/snapshots/visual-regression.spec.js/smart-table--all__chrome.snap.png
wolandec, line 1: why these screenshots are canged?

stavros-tomas: The typical minor pixels. changes you might get because of typography etc. Either way they were failing I decided to update them

wolandec: I can see significant changes between two screenshots, I think that loading state could be a mistake.

stavros-tomas: reverted the snapshots.

src/components/ec-date-range-filter/ec-date-range-filter.vue
```diff
@@ -147,11 +142,30 @@ const numberOfSelectedFilters = computed(() => {
...
+const isAutoHideEnabled = ref(true);
+const allPopoverOptions = computed(() => ({
+  ...props.popoverOptions,
+  autoHide: isAutoHideEnabled.value, // autoHide of the ec-filter-popover should be disabled while flatpickr is open, otherwise selecting value in the flatpickr will close this popover too.
+  hideTriggers: ['close'],
+}));
+
+/* c8 ignore start */
+function onClose() {
+  requestAnimationFrame(() => { // Floating vue requests next animation frame to set values and decide where to autohide or not. We need to do the same and wait for the next animation frame before re-enabling autohide.
```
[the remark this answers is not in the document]
stavros-tomas: Regarding why,  I left another comment few lines above that explains "why" to enable/disable autohide. 

Now as to "when"...
The difference with other cases that we have popover inside popover on chameleon like on [currency filter](https://github.com/Ebury/chameleon/blob/03653312d6ffa4ebc6d78750b4b888aa34d1f7ae/src/components/ec-currency-filter/ec-currency-filter.vue#L55)
is that `flatpicker` does not emit the `after-close` event like `floating-vue` does. So we had two choices either use an abstract timeout or use the same mechanism (requestAnimationFrame) as  [floating vue does](https://github.com/Akryum/floating-vue/blob/62014c34d6dcb25ee9e2189c12552c5b059e7834/packages/floating-vue/src/components/Popper.ts#L1072-L1076)

wolandec (approved):

wolandec (approved):

wolandec (approved):

tests/integration/snapshots/visual-regression.spec.js/smart-table--basic__chrome.snap.png
raulwwq0, line 1: The "Clear filters" button is now being shown even if there is no filter set, it seems setting [`dueDate` filter](https://github.com/Ebury/chameleon/blob/ONL-7281-ec-date-range-input-refactor/src/components/ec-smart-table/ec-smart-table.story.js#L207) (even if both properties are null) counts as setting the filter, so we should check the condition that shows that button.

raulwwq0 (approved):

raulwwq0 (approved):

wolandec (approved):

PR #1831, opened 2024-01-24
[ONL-8197] Migrate composables

mcibique:
Refactored all composables in chameleon and enums they are using. Remaining enums where either deleting or refactored too.

I noticed that TS checks are not executed in CI so I added it.


src/composables/use-ec-sorting/use-ec-sorting.ts
```diff
@@ -0,0 +1,50 @@
+import { ref, unref } from 'vue';
+
+import { SortDirection, SortDirectionCycle } from '../../enums';
+import type { SortingOptions } from './types';
+
+export default function useEcSorting({
+  initialSorts = [],
+  isMultiSort = false,
+  sortCycle: sortCycleInOptions,
+}: Partial<SortingOptions> = {}) {
```
wolandec, line 10: I think we can add `?` to type definition in types file and remove Partial from here. As now if I look at that type I see that all fields are required, but they are definitely not.

src/composables/use-ec-sorting/use-ec-sorting.ts
```diff
@@ -0,0 +1,50 @@
...
+  const defaultSortCycle = unref(sortCycleInOptions) ?? SortDirectionCycle.LOWEST_FIRST;
```
wolandec, line 11: why do we need unref here as by types we don't have reactive members in options

src/composables/use-ec-sorting/use-ec-sorting.ts
```diff
@@ -0,0 +1,50 @@
...
+  // eslint-disable-next-line no-param-reassign
+  isMultiSort = unref(isMultiSort);
```
wolandec, line 13: here the same, why we need `unref`?

src/composables/use-ec-sorting/use-ec-sorting.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  const sorts = ref(unref(initialSorts));
```
wolandec, line 15: and here, I think we don't need unref in all of them, or we need to change type of `SortingOptions`

wolandec (approved):

raulwwq0 (approved):

wolandec (approved):

PR #1848, opened 2024-02-06
[ONL-8215] Removed duplicated types from vue files.

mcibique:
Since Vue 3.3 we don't have to mirror prop types in the vue files any more.


eslint/docs/rules/vue-props-interface.md
```diff
@@ -1,83 +0,0 @@
...
-  label?: string,
-  description?: string,
-  isDisabled?: boolean,
-  isTextInline?: boolean,
-  name?: string,
-  errorMessage?: string,
-  hasError?: boolean
-}
-```
-This following code is incorrect because `<script>` lang is `ts`, and in this case `defineProps` must be typed by interface.
```
wolandec, line 54: How we are going to force this?

mcibique: I didn't notice this, I'll have a look on this rule: https://eslint.vuejs.org/rules/define-props-declaration.html.

mcibique: Wait a minute, we already have that rule enabled: https://github.com/Ebury/chameleon/blob/ONL-8215-removed-duplicated-types/.eslintrc.js#L66. And it works, I can't use `defineProps({})` on `lang="ts"`, same for `defineEmits()`.

wolandec (approved):

stavros-tomas (approved):

wolandec (approved):

PR #1850, opened 2024-02-07
[ONL-8216] Refactored layout components.

mcibique:


src/components/ec-navigation-link/types.ts
```diff
@@ -0,0 +1,15 @@
...
+
+import type { IconName } from '../ec-icon/icon-names';
+
+export interface NavigationLinkProps {
+  text: string,
+  iconName: IconName,
+  iconSize?: number,
+  url: string,
+  isRouterLink?: boolean,
+  isActive?: boolean | Ref<boolean>,
```
wolandec, line 11: there is a type `MaybeRef` in vue for this case. It's also worth considering `MaybeRefOrGetter`

src/components/ec-navigation/types.ts
```diff
@@ -0,0 +1,11 @@
+export interface NavigationBranding {
+  logo: string,
+  name: string,
+}
+
+export interface NavigationProps {
+  isCollapsed: boolean,
+  isCollapsable: boolean,
+  branding: NavigationBranding,
```
wolandec, line 9: is this prop required?

wolandec (approved):

src/components/ec-navigation/types.ts
mcibique: I made it required. JS version didn't require it but the implementation requires it. It's not a big change, EBO passes the branding prop all the time.

PR #1853, opened 2024-02-09
[ONL-8252] Migrated floating components to TS.

mcibique:
Migrated a selection of floating components: 
- Modal
- Panel
- Fullscreen overlay
- Datepicker
- Popover
- Toaster

I have removed all `mountXYZAsTemplate` from spec files because it's impossible to type them when half of the component definition is in the test and half in the mountAsTemplate method. using defineComponent directly in the test when everything is bundled as 1 single object means that the inferred wrapper has everything typed, included `wrapper.vm` and regardless whether we use composition or options API.


global.d.ts
```diff
@@ -5,3 +5,13 @@ declare module 'vue' {
 }
 
 export type Maybe<T> = T | null;
+
+// source: https://github.com/microsoft/TypeScript/issues/43505
+export type NumericRange<
```
wolandec, line 10: https://github.com/sindresorhus/type-fest/blob/main/source/int-range.d.ts I think we can use this type utils library, or something similar. Such a library could be useful in other cases.

wolandec (approved):

derrick-g (approved):

PR #1861, opened 2024-02-14
[ONL-8270 Migrated dropdowns and other inputs to TS.

mcibique:
Migrated:
- EcDropdownSearch
- EcDropdown
- EcBtnDropdown
- EcTextarea
- EcAmountInput
- EcPhoneNumberInput

Fixed:
- ec-amount directive used by ec-amount-input not preserving options whole time. Added tests for this.
- ec-datepicker occasionally crashing in integration tests because we try to reposition destroyed datepicker in C&P.


src/components/ec-amount-input/ec-amount-input.vue
```diff
@@ -85,8 +77,8 @@ const inputModel = computed({
 });
 
 watch([() => props.currency, () => props.locale], () => {
-  if (formattedValue.value) {
-    const formatted = new Intl.NumberFormat(props.locale, { type: 'decimal', maximumFractionDigits: precision.value }).format(unformattedValue.value);
+  if (formattedValue.value && typeof unformattedValue.value === 'number') {
```
wolandec, line 80:
```suggestion
  if (formattedValue.value && typeof unformattedValue.value !== null) {
```

src/components/ec-amount-input/ec-amount-input.vue
```diff
@@ -120,7 +112,13 @@ watch(() => props.modelValue, (newValue) => {
       return;
     }
 
-    formattedValue.value = newValue;
+    if (newValue === null || typeof newValue !== 'string') {
```
wolandec, line 115: it means that if we pass number in `props.modelValue` we won't set `formattedValue` to `newValue`? Previous logic was different.

mcibique: Yes, TS flagged that. The problem with this component is that the modelValue is string when isMasked is true and is number | null if isMasked is false. formattedValue should hold the nicely formatted version of the modelValue and unformattedValue should hold the raw number or null. The old logic could assign the number set via v-model to formattedValue, that's wrong and I had to fix that to make TS happy.

wolandec: There is definitely a lack of discriminated union support for props type in Vue.

wolandec (approved):

stavros-tomas (approved):

src/components/ec-dropdown-search/types.ts
```diff
@@ -0,0 +1,33 @@
+import type { StyleValue } from 'vue';
+
+import type { TooltipOptions } from '../../directives/ec-tooltip/types';
+import type { ZIndexLevel } from '../../enums';
+import type { PopoverProps } from '../ec-popover/types';
+
+export interface DropdownItem<TValue = string> {
```
wolandec, line 7:
```suggestion
export interface DropdownSearchItem<TValue = string> {
```

mcibique: I was thinking about it too but this is shared between ec-dropdown and ec-dropdown-search, so because the ec-dropdown is used more often, it's going to be better to name it DropdownItem. I can also try to make an alias and declare `DropdownItem<TValue> = DropdownSearchItem<TValue>`.

wolandec: I would use prefixes consistently. So here `DropdownSearch` and in normal dropdown just `Dropdown`. It also reflects the dependency `dropdown` on `dropdown-search` if we use `DropdownSearchItem` in `dropdown` component.

stavros-tomas (approved):

src/components/ec-btn-dropdown/ec-btn-dropdown.spec.ts
```diff
@@ -62,17 +66,17 @@ describe('EcBtnDropdown', () => {
...
-    expect(wrapper.emitted('click').length).toBe(1);
+    expect(wrapper.emitted('click')?.length).toBe(1);
   });
 
   it('should emit the change event when the user clicks on an item of the dropdown list', async () => {
     const wrapper = mountBtnDropdown({ items });
 
     await wrapper.findByDataTest('ec-btn-dropdown__dropdown-btn').trigger('click');
-    await wrapper.findAllByDataTest('ec-dropdown-search__item').at(0).trigger('click');
+    await wrapper.findAllByDataTest<HTMLLIElement>('ec-dropdown-search__item').at(0)?.trigger('click');
```
wolandec, line 76:
```suggestion
    await wrapper.findAllByDataTest<HTMLElement>('ec-dropdown-search__item').at(0)?.trigger('click');
```
Misprint?

wolandec (approved):

wolandec (approved):

wolandec (approved):

PR #1871, opened 2024-02-22
[ONL-8283] Migrate remaining components

mcibique:
Those are last remaining chameleon components that were not migrated to TS yet. After this, we only have HOCs in plain JS and we're not going to migrate them because they are obsolete.

I have also removed all `mountXXXAsTemplate` because those can't be typed and updated our docs.


src/components/ec-inline-actions/ec-inline-actions.vue
```diff
@@ -43,7 +43,7 @@
                     v-if="item.icon"
                     :name="item.icon"
                     :size="24"
-                    :type="item.iconType ? item.iconType : null"
+                    :type="item.iconType ? item.iconType : undefined"
```
raulwwq0, line 46:
```suggestion
                    :type="item.iconType ?? undefined"
```

raulwwq0 (approved):

src/components/ec-inline-actions/ec-inline-actions.vue
mcibique: Actually, iconType is already marked as optional and ec-icon's type can accept undefined already, so it should be just `:type="item.iconType`. Fixed!

raulwwq0 (approved):

derrick-g (approved):

PR #1910, opened 2024-04-04
[CHUX-295] Force ec-modal to trap focus

wolandec:
Considering that modal should always trap the focus(as it is modal nature), even if ec-modal component somehow loses it, we add this continious activation of focusTrap to ensure that modal always traps focus back.

### DIFF
--- a/src/components/ec-modal/ec-modal.vue
+++ b/src/components/ec-modal/ec-modal.vue
@@ -188,7 +188,9 @@ function getFocusTrapOptions(): UseFocusTrapOptions {
   return options;
 }
 
-const { deactivate } = useFocusTrap(focusTrapTarget, getFocusTrapOptions());
+const { deactivate, activate } = useFocusTrap(focusTrapTarget, getFocusTrapOptions());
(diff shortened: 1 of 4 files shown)


src/components/ec-modal/ec-modal.vue
```diff
@@ -188,7 +188,9 @@ function getFocusTrapOptions(): UseFocusTrapOptions {
   return options;
 }
 
-const { deactivate } = useFocusTrap(focusTrapTarget, getFocusTrapOptions());
+const { deactivate, activate } = useFocusTrap(focusTrapTarget, getFocusTrapOptions());
+const focusTrapInterval = setInterval(() => activate(), 300);
```
derrick-g, line 192: So this keeps activating focustrap every 300ms, is this the best way?

wolandec: Yes it keeps activating, I haven't found another way to trap the focus on modal and don't lose it if another component want's to trap it.

derrick-g (approved):

stavros-tomas (approved):

PR #1947, opened 2024-05-07
[CHUX-460] Add new prop to ec-container

raulwwq0:
Add new prop to apply styles for a mobile layout


src/components/ec-container/types.ts
```diff
@@ -1,3 +1,4 @@
 export interface ContainerProps {
   isCollapsable?: boolean,
+  showMobileLayout?: boolean,
```
wolandec, line 3: we name such props `isResponsive` e.g. smart-table

wolandec: It's `true` by default and we need to set this prop based on flag value in EBO

raulwwq0: yeap, however we will need to set `isResponsive` to false for now or it will break EBO UI, should we do a release now to amend it or wait and add it in [this ticket](https://fxsolutions.atlassian.net/browse/CHUX-501)?

wolandec: Ok, let it be false by default.

wolandec (approved):

PR #1948, opened 2024-05-08
[CHUX-242] Remove ellipsis on table filters

stavros-tomas:
Removing the max width limitations, if we do not like it in the future we can revert back to ellipsis or think of a better solution like inline the filters.


raulwwq0 (approved):

src/components/ec-filter-popover/ec-filter-popover.vue
```diff
@@ -111,9 +111,6 @@ function onOpen(isOpen: boolean) {
 
   &__label {
     @apply tw-text-gray-3;
-    @apply tw-truncate;
```
raulwwq0, line 114: Just one thing, I don't think we should allow the text from the filters to wrap, it should be in the same line

image

stavros-tomas: I don't think we should restrict it. It's unnatural to don't allow the element to wrap, fold etc. 
If you don't allow them to do it and you also remove the ellipsis,  then will be even worst they will overflow.

If we see that they wrap then we need to do something in the UI like remove a filter or so.

raulwwq0: Fine by me, but I would ask Alba anyways just in case...

stavros-tomas: I asked Alba if she still wants to remove the ellipsis (which i think is the correct way to do this)

stavros-tomas: We will go with this solution, if we don't like we can revert back to ellipsis or do some more major redesign of the header.

raulwwq0 (approved):

PR #1954, opened 2024-05-15
[CHUX-462] ec-dropdown - expand ui to accomodate new design

corinaper:
ERI: https://drive.google.com/file/d/10wuX4Lplm7LDReVtw96Dbzhw1HbkdQwZ/view?usp=sharing


src/components/ec-input-field/ec-input-field.vue
```diff
@@ -135,6 +135,8 @@ const props = withDefaults(defineProps<InputFieldProps>(), {
   isLoading: false,
   isSensitive: false,
   isWarning: false,
+  isInLightMode: false,
+  isDropdownOpen: false,
```
wolandec, line 139: Input field shouldn't care about dropdown, we shouldn't introduce props from the context of another component.

corinaper: should i rename it to is-input-bg-white?

wolandec: What is this prop for? Why we didn't have it before? Also could you please add light mode in storybook for dropdown and input and update screenshots.

corinaper: I added it in storybook for dropdown, will add it to input too

Based on the design we have to change the input bg color when the dropdown is visible, that's why i added the prop. We didn't have this change before, that's why we didn't have a prop

raulwwq0 (approved):

src/components/ec-input-field/ec-input-field.vue
```diff
@@ -192,6 +190,22 @@ const inputClasses = computed(() => {
   return classes;
 });
 
+function getInputTooltipOptions(): TooltipOptions {
```
wolandec, line 193: I think this should be computed

corinaper: it looses reactivity if I send a computed to vTootlitp. Not sure of the reason, but the function seems to work

wolandec: https://github.com/Ebury/chameleon/blob/2de00b219d1ad3d015917418d2847f47a1e9ba35/src/components/ec-currency-input/ec-currency-input.vue#L213 we do can use computed with tooltip

src/components/ec-input-field/types.ts
```diff
@@ -46,4 +47,6 @@ export interface InputFieldProps {
   isSensitive?: boolean,
   isWarning?: boolean,
   autocomplete?: string,
+  bgGrayColorLevel?: 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8,
```
wolandec, line 50:
```suggestion
  bgColorLevel?: 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8,
```

src/components/ec-input-field/ec-input-field.vue
```diff
@@ -156,6 +162,9 @@ const inputModel = computed<InputFieldProps['modelValue']>({
 const inputClasses = computed(() => {
   const classes = ['ec-input-field__input'];
 
+  if (props.bgGrayColorLevel) {
+    classes.push(`tw-bg-gray-${props.bgGrayColorLevel}`);
```
wolandec, line 166: this won't be recognised by CSS class cleaning tool as purgeCSS for example. We need to define classes for all levels in style tag and use them in this push. Otherwise we can loose this classes if they are not used in other cases in style tag or template, which is unlikely in this case as bg-gray is used widely, but....

src/components/ec-input-field/ec-input-field.vue
```diff
@@ -127,6 +130,7 @@ const props = withDefaults(defineProps<InputFieldProps>(), {
   type: InputFieldType.TEXT,
   label: '',
   labelTooltip: '',
+  inputTooltip: '',
```
wolandec, line 133: why do we need this prop? can't we use modelValue?

src/components/ec-input-field/ec-input-field.vue
```diff
@@ -135,6 +139,8 @@ const props = withDefaults(defineProps<InputFieldProps>(), {
   isLoading: false,
   isSensitive: false,
   isWarning: false,
+  bgGrayColorLevel: 8,
+  showPointerCursor: false,
```
wolandec, line 143: why do we need this prop?

src/components/ec-input-field/ec-input-field.vue
corinaper: we can, but I added it in case we don't want to show a tooltip. Do you think it's better then to have a show-input-tooltip prop instead?

src/components/ec-input-field/ec-input-field.vue
corinaper: we don't want to show the pointer cursor for a normal input, but we do show it if it's a dropdown input

wolandec: I think we should always show it if the content doesn't fit.

src/components/ec-input-field/ec-input-field.vue
wolandec: I think we should always show it if the content doesn't fit.

corinaper: I thought it would be a bit weird to have a tooltip for an input when the user is just typing and can simply scroll

wolandec: Ok

wolandec: BTW if tooltip is passed, then we should show it in any case, so `getInputTooltipOptions` is not needed.

raulwwq0 (approved):

stavros-tomas (approved):

PR #1962, opened 2024-05-23
[CHUX-466] ec-navigation sidebar

raulwwq0:


src/components/ec-navigation/ec-navigation.vue
```diff
@@ -21,6 +25,9 @@
     <div
       v-if="showBrandingLogo && branding.logo"
       class="ec-navigation__branding"
+      :class="{
+        'ec-navigation__branding--light-mode': isInLightMode,
+      }"
```
corinaper, line 30: maybe instead of checking the prop again and assigning a new class, we could just target in styles the class .ec-navigation__branding that proceeds ec-navigation--light-mode, so:
```
.ec-navigation {
  &--light-mode {
    @apply tw-bg-gray-7;
    @apply tw-text-gray-3;
    @apply tw-py-32 tw-px-24;

    & .ec-navigation__branding{
      @apply tw-m-0 tw-mb-16 tw-p-0;
      @apply tw-text-left;
    }
  }

```
We won't have any scenarios in which the branding logo won't need this style if the light mode prop is true, right?

raulwwq0: Good point, you're right, I'm going to change it

corinaper (approved):

corinaper (approved):

PR #1968, opened 2024-05-31
[CHUX-524] Align the full screen content container

raulwwq0:
EBO with the changes and max-width from `ebo-container` removed:
image


corinaper (approved):

corinaper (approved):

stavros-tomas (approved):

src/styles/settings/layout.css
```diff
@@ -1,4 +1,5 @@
 :root,
 :host {
   --ec-body-max-width: 1260px;
+  --ec-main-container-max-width: 1200px;
```
stavros-tomas, line 4: I am not sure if we can and if it make sense to add this value here,
If you have time have a deeper look into it is purely a thought.

 https://github.com/Ebury/chameleon/blob/master/tailwind.config.js#L745
 
 
 Other than that feel free to merge, lgtm :muscle:

raulwwq0: Good point! Actually it seems none of those is the proper solution:
- Following ITCSS principles, adding it in settings/layout.css can be done only if the CSS var is going to be used in other places (such as colors), so the var should not be there
- Adding it in the Tailwind config is not valid as well since this value is not going to be reused anywhere else

The way we used to proceed with this kind of vars that are going to be used only in one place is just place it in the `<style>` tag of the component where is going to be used. here you have some examples of this: [ec-currency-input](https://github.com/Ebury/chameleon/blob/master/src/components/ec-currency-input/ec-currency-input.vue#L223-L226), [ec-dropdown-search](https://github.com/Ebury/chameleon/blob/master/src/components/ec-dropdown-search/ec-dropdown-search.vue#L569-L577) or [ec-filter-popover](https://github.com/Ebury/chameleon/blob/master/src/components/ec-filter-popover/ec-filter-popover.vue#L81-L85)

PR #1980, opened 2024-06-13
[CHUX-661] Add prop to ec-letter-icon to control hover styles

raulwwq0:
This is needed if we want to keep using good CSS practices in EBO, as it can be done by modifying the styles of the Chameleon component on EBO, but this approach is against best practices.

### DIFF
--- a/src/components/ec-letter-icon/ec-letter-icon.vue
+++ b/src/components/ec-letter-icon/ec-letter-icon.vue
@@ -6,6 +6,7 @@
     class="ec-letter-icon"
     :class="{
       'ec-letter-icon--clickable': isClickable,
+      'ec-letter-icon--hover-styles': applyHoverStyles,
       [`ec-letter-icon--${size}`]: true,
     }"
   >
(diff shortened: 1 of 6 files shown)


src/components/ec-letter-icon/ec-letter-icon.vue
```diff
@@ -26,6 +27,7 @@ import { type LetterIconProps, LetterIconSize } from './types';
 const props = withDefaults(defineProps<LetterIconProps>(), {
   size: LetterIconSize.SMALL,
   isClickable: false,
+  applyHoverStyles: false,
```
stavros-tomas, line 30: nit picking here, but since is effectively a boolean should be `is` or `has`

So something like `isHoverable` or `hasHoverEnabled` or similar

raulwwq0: Sure, let's go with `isParentHovered` since the purpose of the prop is to trigger the hover styles when the parent is hovered

stavros-tomas (approved):

stavros-tomas (approved):

PR #2002, opened 2024-07-03
[CHUX-887] sticky mobile header

stavros-tomas:
Make mobile header sticky


corinaper (approved):

src/components/ec-navigation/ec-navigation.vue
```diff
@@ -1,17 +1,21 @@
 <template>
-  <ec-mobile-header
+  <div
     v-if="isResponsive && !isMobileMenuOpen"
-    @open-mobile-menu="isMobileMenuOpen = true"
+    class="ec-navigation__mobile-header__container"
```
corinaper, line 4: I'm not sure about this class. Maybe it would be better ec-navigation__mobile-header-container
It looks like this goes against the BEM style: https://github.com/Ebury/eburyonline/blob/master/docs/FRONTEND.md
`2. You should **never** use nested elements like `menu__tabs__item`.`

stavros-tomas: correct will amend in the next pr buttons

corinaper (approved):

PR #2005, opened 2024-07-08
[CHUX-690] Make metroline footer cta mobile friendly

stavros-tomas:
In this PR
- Removing the max-width limitation of button components.
- Making the metroline CTA footer expand the full width on mobile devices. As a consequence this will also allow the cta items to expands as well to the full width. 

also
- fixing some BEM stuff

also
- setting up a z-index to ec-panel because mobile header is overlaying it

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@ebury/chameleon-components",
-  "version": "2.8.15",
+  "version": "2.8.16",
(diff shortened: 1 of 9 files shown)


corinaper (approved):

corinaper:
We might need to update the visual regression test

corinaper (approved):

PR #2063, opened 2025-03-06
[CHOPS-1868] Add Chile / CLP country flag

AfrazHussain:
This PR add the country flag for Chile / CLP. I have generated updated snapshots for visual regression tests.


corinaper:
Could you please check the svg, based on the snapshot, the icon isn't displaying correctly

corinaper:
> @corinaper Thanks for the review. The SVG was exported from Figma. I did not minify/optimize it using SVGO (if you want I can do that but the others aren't optimised). Could you please point to the snapshot where the icon isn't displaying correctly so I can look into this? Thanks.

I haven't looked into the cause of the issue, but if you check all the generated snapshots, the icon is not displaying correctly

corinaper:
> @corinaper Thanks for the review. The SVG was exported from Figma. I did not minify/optimize it using SVGO (if you want I can do that but the others aren't optimised). Could you please point to the snapshot where the icon isn't displaying correctly so I can look into this? Thanks.

sorry, my mistake. I couldn't find the new icon and assumed the no-currency icon is the new one added and that is not displaying correctly

corinaper (approved):

AfrazHussain:
Thanks a lot @corinaper :100:

PR #2065, opened 2025-03-28
[NBL-480]: Update table components to support multi selection

teogrosu:
Updating the smart-table, table and heading components to allow adding an optional column that contains checkboxes to select things in the table. We should also be able to pass an `isSelectableCheck` function so that we can filter which rows can be selectable. 
** changes to onRowClick behaviour - the `onRowClick` fn should only be called when clicking anywhere in the row besides the checkbox cell - we usually want to open up a details panel when clicking on the row and this makes sure that selecting a row won't also open up that panel.

Using the new props in the activity log table (new Neyes functionality) - without passing `isSelectableCheck`:
image

Using `isSelectableCheck` so only the changes in the table that have a status of 'pending your approval' can be selected:
image

Check out this PR https://github.com/Ebury/eburyonline/pull/2440 for the full implementation.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@ebury/chameleon-components",
-  "version": "2.8.21",
+  "version": "2.8.22",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@ebury/chameleon-components",
-      "version": "2.8.21",
+      "version": "2.8.22",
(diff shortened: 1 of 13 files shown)


pavlos-pavlopoulos-ebury (approved):
Looks good!

pavlos-pavlopoulos-ebury (approved):

AfrazHussain (approved):
lgtm.

